PR #76, opened 2023-02-16
feat: fall back to `getLogs` if filters are not supported

jxom:
Some RPC providers do not support `eth_newFilter`. If this is the case, we will fall back and try `eth_getLogs` at the cost of potentially higher compute units.


0xOlias:
Do you guys know examples of RPC providers that don't support `eth_newFilter`?

jxom:
Cloudflare

PR #113, opened 2023-03-02
wip: local signer (ethers-rs x wasm)

jxom:


package.json
```diff
@@ -7,6 +7,7 @@
     "bench": "vitest bench --no-threads",
     "bench:ci": "CI=true vitest bench --no-threads",
     "build": "tsup",
+    "build:wasm": "cd wasm && wasm-pack build -t nodejs",
```
roninjin10, line 10: I think you can just pass wasm in as a path argument instead of cding
```suggestion
    "build:wasm": "wasm-pack build -t nodejs wasm",
```

PR #145, opened 2023-03-07
docs: fix home page

chenxsan:
Hi, just noticed a weird bug where toc has no padding:

CleanShot 2023-03-07 at 19 10 12@2x

And it's caused by the global style defined in home page:

```
<style>
  .VPDoc .container {
    max-width: 1120px !important;
  }

  .VPDoc .content {
    max-width: 100% !important;
    padding: 0px !important;
    width: 100% !important;
  }
</style>
```

In this fix, I've removed that global styles and wrapped the whole content inside a `div` with tailwindcss classes.

And here's the toc after the fix:

CleanShot 2023-03-07 at 19 13 25@2x

### DIFF
--- a/site/.vitepress/theme/components/HomeSponsors.vue
+++ b/site/.vitepress/theme/components/HomeSponsors.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import { VPButton, VPIconHeart, VPSponsors } from 'vitepress/theme';
 import { sponsors } from '../sponsors'
 </script>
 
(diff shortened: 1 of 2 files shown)


jxom:
Screenshot 2023-03-08 at 7 41 27 am

looks like we lost the sponsors 👁️⌒👁️

chenxsan:
> Screenshot 2023-03-08 at 7 41 27 am
> 
> looks like we lost the sponsors 👁️⌒👁️

My fault, just took it for granted that I can remove those without double check:

CleanShot 2023-03-08 at 08 02 55@2x

I've reverted the related changes.

PR #146, opened 2023-03-07
chore: ignore `@algolia/client-search` peer dependency

0xYami:
Hello there, peer dependency `@algolia/client-search` could be ignored as per VitePress [doc](https://vitepress.vuejs.org/guide/getting-started#step-2-install-vitepress)


tmm:
Nice one!

PR #170, opened 2023-03-09
feat: infer `multicallAddress` from client chain for `multicall`

jxom:
Infers `multicallAddress` from client chain for `multicall` like we do in `getEnsName` and `getEnsAddress`.

### DIFF
--- a/.changeset/spotty-llamas-think.md
+++ b/.changeset/spotty-llamas-think.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Infer multicall address from client chain
+Added inference for multicall address from client chain.
--- a/.changeset/spotty-llamas-think.md
+++ b/.changeset/spotty-llamas-think.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Infer multicall address from client chain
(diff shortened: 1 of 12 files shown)


.changeset/spotty-llamas-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"viem": patch
+---
+
+Infer multicall address from client chain
```
tmm, line 5:
```suggestion
Added inference for multicall address from client chain.
```

src/utils/chain.ts
```diff
@@ -1,8 +1,43 @@
 import { Chain, Formatters } from '../types'
+import { ChainDoesNotSupportContract } from '../errors'
+import { Contract } from '../types/chain'
```
tmm, line 3: Should `Contract` be exported from `../types/index.ts`? If there's a naming conflict, we could call it `ChainContract`.

src/utils/chain.ts
```diff
@@ -1,8 +1,43 @@
...
 
 export function defineChain<
   TFormatters extends Formatters = Formatters,
   TChain extends Chain<TFormatters> = Chain<TFormatters>,
 >(chain: TChain) {
   return chain
 }
+
+export function getChainContractAddress({
```
tmm, line 12: nice refactor!

tmm (approved):

PR #178, opened 2023-03-10
fix: topics type for logs in rpc responses

0xOlias:
AFAICT, the `topics` property of logs in RPC responses can only be an array of hex strings. Elements of this array will never be null, and will never be an array of strings, so I think the type suggested here is more precise. See [this discussion](https://github.com/wagmi-dev/viem/discussions/163) and the [JSON-RPC spec](https://ethereum.github.io/execution-apis/api-documentation/) for more details.

To summarize:
1. The `topics` argument passed to `eth_getLogs` and `eth_newFilter` has the type `(Hex | Hex[] | null)[]`
2. The `topics` property of logs returned by `eth_getLogs`, `eth_getFilterLogs`, and `eth_getFilterChanges` has the type `Hex[]` (and has length <= 4)

### DIFF
--- a/.changeset/dull-windows-tease.md
+++ b/.changeset/dull-windows-tease.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed type of `topics` field on `Log`.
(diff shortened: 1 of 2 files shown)


jxom (commented):
Can we add a changeset?

PR #181, opened 2023-03-10
chore: bump abitype

tmm:
- Bumps [ABIType](https://abitype.dev)
- Removes custom config
- Refactors to `AbiFunction` and `AbiConstructor`


package.json
```diff
@@ -133,6 +132,7 @@
     "@vitest/coverage-c8": "^0.29.2",
     "@vitest/ui": "^0.29.2",
     "@wagmi/cli": "^0.1.6",
+    "abitype": "^0.7.1",
```
tmm, line 135: I'm good locking it, but can add `~` back if you want.

src/types/abitype.ts
```diff
@@ -1,14 +1,3 @@
-import type { Hex } from './misc'
-
-declare module 'abitype' {
-  export interface Config {
-    BytesType: {
-      inputs: Hex
-      outputs: Hex
-    }
-  }
-}
```
tmm, line 10: This is the new default behavior in `abitype@0.7`

src/types/abitype.ts
```diff
@@ -1,14 +1,3 @@
...
-
 export type {
   Abi,
   AbiError,
```
tmm, line 3: Can we remove this file? Seems easier to import these from abitype directly rather than an internal export.

jxom: Go for it

jxom (approved):

package.json
jxom: Doesn't abitype need to be a dependency?

PR #204, opened 2023-03-13
feat: test types

tmm:
Adds type tests to a few key places.


tmm (commented):
Probably don't need to add these tests everywhere as long as we test the main utility types and a couple other places.

jxom (approved):

PR #223, opened 2023-03-16
feat: assert that client chain matches wallet's active chain

jxom:
We currently check if the **`chain` arg** on `sendTransaction` matches the wallet's active chain, but we don't check if the **wallet client `chain`** matches the wallet's active chain. Would be good to add it in.

### DIFF
--- a/site/docs/contract/writeContract.md
+++ b/site/docs/contract/writeContract.md
@@ -270,6 +270,49 @@ await walletClient.writeContract({
 })
 ```
 
+### assertChain (optional)
+
+- **Type:** `boolean`
+- **Default:** `true`
+
+Throws an error if `chain` does not match the current wallet chain.
+
+Defaults to `true`, but you can turn this off if your dapp is primarily multi-chain.
+
+```ts
+import { optimism } from 'viem/chains' // [!code focus]
+
+await walletClient.writeContract({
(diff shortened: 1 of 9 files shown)


site/docs/contract/writeContract.md
```diff
@@ -270,6 +270,49 @@ await walletClient.writeContract({
 })
 ```
 
+### assertChain (optional)
```
jxom, line 273: Adding missing docs on `writeContract`.

src/actions/wallet/sendTransaction.test.ts
```diff
@@ -120,6 +127,32 @@ test.skip('sends transaction w/ no value', async () => {
...
+test('client chain mismatch', async () => {
+  const walletClient = createWalletClient({
+    chain: optimism,
+    transport: http(localHttpUrl),
+  })
+  await expect(() =>
+    sendTransaction(walletClient, {
+      account: getAccount(sourceAccount.address),
+      to: targetAccount.address,
+      value: parseEther('1'),
```
roninjin10, line 139: I think it would be good to pass in some kind of `autoSwitch: true` here that would automatically switch chains first if connected to wrong chain.   Made a discussion [here](https://github.com/wagmi-dev/viem/discussions/224)

PR #227, opened 2023-03-16
feat: require chain for `sendTransaction`, `writeContract` & `deployContract`

jxom:
Right now, a Chain is optional to execute wallet Actions such as `sendTransaction`, `writeContract`, and `deployContract`. However, not passing a Chain could lead to unintended UX consequences (a user could be connected to a different chain).

This PR sets up a better default, whereby we require a Chain for these Actions by default to avoid such UX issues. 

A Chain can either be passed to the Client for the Actions to inherit, or a Chain can be directly passed to the Action itself.


src/actions/wallet/sendTransaction.test.ts
```diff
@@ -153,6 +153,30 @@ test('client chain mismatch', async () => {
...
+  ).rejects.toThrowErrorMatchingInlineSnapshot(`
+    "No chain was provided to the request.
+    Please provide a chain with the \`chain\` argument on the Action, or by supplying a \`chain\` to WalletClient.
+
+    Request Arguments:
+      from:   0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
+      to:     0x70997970c51812dc3a010c7d01b50e0d17dc79c8
+      value:  1 ETH
+
+    Version: viem@1.0.2"
```
roninjin10, line 176: Do you mock out the version so it always stays 1.0.2?

src/actions/public/waitForTransactionReceipt.ts
```diff
@@ -14,30 +14,34 @@ import type { GetTransactionReceiptReturnType } from './getTransactionReceipt'
...
+> = GetTransactionReceiptReturnType<TChain>
+
+export type WaitForTransactionReceiptParameters<
+  TChain extends Chain | undefined = Chain,
+> = {
+  /** The number of confirmations (blocks that have passed) to wait before resolving. */
+  confirmations?: number
+  /** The hash of the transaction. */
+  hash: Hash
+  onReplaced?: (response: ReplacementReturnType<TChain>) => void
```
roninjin10, line 35:
```suggestion 
  /** Optional callback to emit if the transaction has been replaced. */
  onReplaced?: (response: ReplacementReturnType<TChain>) => void
```

roninjin10 (approved):
I think this is generally a good futureproof idea to try to completely avoid the concept of being connected to a single chain

src/actions/wallet/sendTransaction.test.ts
tmm: Yes https://github.com/wagmi-dev/viem/blob/main/src/_test/setup.ts#L13

tmm (approved):

PR #234, opened 2023-03-19
feat: private key & hd accounts

jxom:
- Added Local Account implementations: `privateKeyToAccount`, `mnemonicToAccount`, `hdKeyToAccount`
- Added Private Key generation util: `generatePrivateKey`
- Added BIP39 wordlist generation util: `generateMnemonic`
- Added Account signing utils: `signMessage`, `signTypedData`, `signTransaction`
- Added Account signature/pubkey utils: `signatureToHex`, `publicKeyToAddress`


site/docs/accounts/hd.md
```diff
@@ -0,0 +1,139 @@
...
+const hdKey = HDKey.fromMasterSeed(...) // [!code focus:3]
+const hdKey = HDKey.fromExtendedKey(...)
+const hdKey = HDKey.fromJSON({ xpriv: ... })
+
+const account = hdKeyToAccount(hdKey) // [!code focus]
+
+const client = createWalletClient({
+  account,
+  chain: mainnet,
+  transport: custom(window.ethereum)
```
tmm, line 57:
```suggestion
  transport: http(),
```

src/accounts/wordlists.ts
```diff
@@ -0,0 +1,9 @@
+export { wordlist as czech } from '@scure/bip39/wordlists/czech'
+export { wordlist as english } from '@scure/bip39/wordlists/english'
+export { wordlist as french } from '@scure/bip39/wordlists/french'
+export { wordlist as italian } from '@scure/bip39/wordlists/italian'
+export { wordlist as japanese } from '@scure/bip39/wordlists/japanese'
+export { wordlist as korean } from '@scure/bip39/wordlists/korean'
+export { wordlist as simplifiedChinese } from '@scure/bip39/wordlists/simplified-chinese'
+export { wordlist as spanish } from '@scure/bip39/wordlists/spanish'
+export { wordlist as traditionalChinese } from '@scure/bip39/wordlists/traditional-chinese'
```
tmm, line 9: Should we add `Wordlist` to the end of these (e.g. `englishWordlist`)? No strong preference.

package.json
```diff
@@ -6,7 +6,7 @@
     "anvil": "source .env && anvil --fork-url $VITE_ANVIL_FORK_URL --fork-block-number $VITE_ANVIL_BLOCK_NUMBER --block-time $VITE_ANVIL_BLOCK_TIME",
     "bench": "vitest bench --no-threads",
     "bench:ci": "CI=true vitest bench --no-threads",
-    "build": "tsup",
+    "build": "tsup && size-limit",
```
tmm, line 9: Is the object of adding `size-limit` here to fail the build if greater than a certain size?

site/docs/clients/wallet.md
```diff
@@ -20,7 +20,7 @@ The `createWalletClient` function sets up a Wallet Client with a given [Transpor
 
 The Wallet Client supports signing over:
 - a [JSON-RPC Account](#json-rpc-accounts) (ie. Browser Extension Wallets, WalletConnect, etc). 
-- a [Local Account](#local-accounts-experimental) (ie. local private key/mnemonic wallets).
+- a [Local Account](#local-accounts-private-key-mnemonic-etc) (ie. private key/mnemonic wallets).
```
tmm, line 23: I'd make this plural as well as `JSON-RPC Account` above (won't let me add a suggestion to it).

```suggestion
- [Local Accounts](#local-accounts-private-key-mnemonic-etc) (ie. private key/mnemonic wallets).
```

src/types/utils.ts
```diff
@@ -116,6 +116,15 @@ type TrimRight<T, Chars extends string = ' '> = T extends `${infer R}${Chars}`
   ? TrimRight<R>
   : T
 
+/**
+ * @description Creates a type with required keys K from T.
+ *
+ * @example
+ * RequiredBy<{ a?: string, b?: number, c: number }, 'a' | 'c'>
+ * => { a: string, b?: number, c: number }
```
tmm, line 124: Recently trying to use `// ^?` for `@example`s

```suggestion
 * type Result = RequiredBy<{ a?: string, b?: number, c: number }, 'a' | 'c'>
 * //   ^? { a: string, b?: number, c: number }
```

src/utils/accounts.ts
```diff
@@ -0,0 +1 @@
+export { parseAccount } from '../accounts/utils/parseAccount'
```
tmm, line 1: Should we remove this file an import `parseAccount` from `src/accounts` instead?

tmm (approved):

src/utils/accounts.ts
jxom: Would love to, unfortunately, the bundler gets confused with tree-shaking when we using index entrypoints, so stuff like @noble/secp256k1 and the bip32 libs make their way into the main bundle (an extra 8kb or so). Need to fix this at some point.

package.json
jxom: Just to keep an eye on bundle size really. Can set a limit also if we want.

src/utils/accounts.ts
tmm: Got it. Want to add a note to the file?

PR #237, opened 2023-03-21
feat: automated `fallback` ranking

jxom:
Adds an automated ranking algorithm to rank Transports provided to the `fallback` Transport.

By default, each of the Transports passed to the `fallback` Transport are automatically ranked based on their **latency** & **stability** via a weighted moving score algorithm. 

Every 10 seconds (`interval`), the `fallback` Transport will ping each transport in the list. For the past 10 pings (`sampleCount`), they will be ranked based on if they responded (stability) and how fast they responded (latency). The algorithm applies a weight of `0.7` to the stability score, and a weight of `0.3` to the latency score to derive the final score which it is ranked on.

The Transport that has the best latency & stability score over the sample period is prioritized first.

### DIFF
--- a/src/clients/transports/fallback.ts
+++ b/src/clients/transports/fallback.ts
@@ -1,12 +1,44 @@
+import { Chain } from '../../types'
 import { isDeterministicError } from '../../utils/buildRequest'
+import { wait } from '../../utils/wait'
 import type { Transport, TransportConfig } from './createTransport'
 import { createTransport } from './createTransport'
+
+type RankOptions = {
+  /**
+   * The polling interval (in ms) at which the ranker should ping the RPC URL.
+   * @default client.pollingInterval
+   */
+  interval?: number
+  /**
+   * The weight to apply to the latency score.
+   * @default 0.3
+   */
+  latencyWeight?: number
+  /**
+   * The number of previous samples to perform ranking on.
+   * @default 10
+   */
+  sampleCount?: number
+  /**
+   * The weight to apply to the stability score.
+   * @default 0.7
+   */
+  stabilityWeight?: number
+  /**
+   * Timeout when sampling transports.
+   * @default 1_000
+   */
+  timeout?: number
+}
 
 export type FallbackTransportConfig = {
   /** The key of the Fallback transport. */
   key?: TransportConfig['key']
   /** The name of the Fallback transport. */
   name?: TransportConfig['name']
+  /** Toggle to enable ranking, or rank options. */
+  rank?: boolean | RankOptions
   /** The max number of times to retry. */
   retryCount?: TransportConfig['retryCount']
   /** The base delay (in ms) between retries. */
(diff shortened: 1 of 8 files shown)


src/clients/transports/fallback.ts
```diff
@@ -1,12 +1,44 @@
...
+  /**
+   * The number of previous samples to perform ranking on.
+   * @default 10
+   */
+  sampleCount?: number
+  /**
+   * The weight to apply to the stability score.
+   * @default 0.7
+   */
+  stabilityWeight?: number
```
tmm, line 27: what happens if `latencyWeight` and `stabilityWeight` sum to more than `1`? should we add a check for this?

could also imagine maybe we add additional factors in the future. should we move latencyWeight` and `stabilityWeight` to a `weights` property?

```diff
const client = createPublicClient({
  transport: fallback(
    [...],
    { 
      rank: {
        interval: 60_000,
-       latencyWeight: 0.4,
        sampleCount: 5,
-       stabilityWeight: 0.6,
        timeout: 500,
+       weights: {
+         latency: 0.4,
+         stability: 0.6,
+       },
      },
    },
  ),
})
```

src/clients/transports/fallback.ts
```diff
@@ -60,4 +100,104 @@ export function fallback(
...
+  chain?: Chain
+  interval: number
+  latencyWeight?: number
+  onTransports: (transports: Transport[]) => void
+  sampleCount?: number
+  stabilityWeight?: number
+  timeout?: number
+  transports: Transport[]
+}) {
+  let samples: Sample[] = []
```
tmm, line 144: If `SampleData` and `Sample` are only used internally, could move them inside the function so we don't accidentally use them anywhere else.

```suggestion
export function rankTransports({
  chain,
  interval = 4_000,
  latencyWeight = 0.3,
  onTransports,
  sampleCount = 10,
  stabilityWeight = 0.7,
  timeout = 1_000,
  transports,
}: {
  chain?: Chain
  interval: number
  latencyWeight?: number
  onTransports: (transports: Transport[]) => void
  sampleCount?: number
  stabilityWeight?: number
  timeout?: number
  transports: Transport[]
}) {
  type SampleData = [latency: number, success: number]
  type Sample = SampleData[]
  let samples: Sample[] = []
```

src/clients/transports/fallback.ts
```diff
@@ -60,4 +100,104 @@ export function fallback(
...
+        timeout: rankOptions.timeout,
+        transports,
+      })
+    }
+
+    return transport
+  }
+}
+
+type SampleData = [latency: number, success: number]
```
tmm, line 122: love labelled tuples, but might be easier to follow if this was an object (especially if we add more data properties in the future).

src/clients/transports/fallback.ts
```diff
@@ -60,4 +100,104 @@ export function fallback(
...
+  latencyWeight?: number
+  onTransports: (transports: Transport[]) => void
+  sampleCount?: number
+  stabilityWeight?: number
+  timeout?: number
+  transports: Transport[]
+}) {
+  let samples: Sample[] = []
+
+  const rankTransports_ = async () => {
```
tmm, line 146: super clean to follow. some comments might be nice at each step for the future. Something like:

```
// 1. Sample from each transport
...

// 2. Store samples and calculate max latency
...

// 3. Compare samples and sort transports
...

// 4. Pause for interval and rerun rank
...
```

tmm (approved):
awesome work!

src/clients/transports/fallback.ts
jxom: You could technically also set `latencyWeight` to `40` and `stabilityWeight` to `60`, and it will be the same result. Maybe we say in the docs that the weights are proportional.

PR #250, opened 2023-03-27
[WIP] chore: use `NodeNext` module resolution and use `tsc` for builds

### LINKED ISSUE #244: bug: type inferrence not working fully with NodeNext moduleResolution

### Package Version

0.1.22

### Current Behavior

I haven't looked into this further yet but wanted to drop this issue here before I call it a night ...:

When using `NodeNext` moduleResolution, building with `tsc` & declaration / declarationMap enabled errors for just about every inferred type that one may try to re-export from `viem`:

E.g. for something like

```typescript
export const client = createClient({ ... });
```

You'd get:

```
The inferred type of 'client' cannot be named without a reference to '../node_modules/viem/dist/createClient-a28317a9.js'. This is likely not portable. A type annotation is necessary.
```

### Expected Behavior

Viem also works with NodeNext module resolution

### Steps To Reproduce

I can set up a reproduction tomorrow and dive deeper & fix this issue, but here are the steps anyways:

- Set up a pure esm project
- Re-export sth. from `viem`, e.g. a client object (`createClient(...)`)
- Use standard `tsc` for compilation. Do not use `esbuild` or `swc` or any other derived tools like `tsup` as they generally do not process & honor `declaration` and `declarationMap` build output to the same extent  (or at all) as `tsc`
- Enable `declaration` and `declarationMap` output in tsconfig.json
- Set `moduleResolution` to `NodeNext`
- Run `tsc` with build output (this is important as this issue doesn't appear with `--noEmit` or if `declaration` or `declarationMap` output is disabled

You'll now see the aforementioned error ...

Will share tomorrow (or attempt to submit a fix)

### Anything else?

I love the work you are doing to modernize the JS/TS ecosystem in this space. It's long overdue! :-)

Not quite related, but ... I'd really like for `viem` to also generate (& publish) `declarationMap` output for better IDE navigation capabilities (same for abitype & wagmi). I understand why `tsup` has gained this much adoption and I've (and am) using it a lot for side projects too but it can't replace `tsc` entirely imho, especially for libraries like these!

fubhy:
Fixes #244

NOTE: For now, this is just meant to help illustrate what I'd suggest for the build setup and type config. Feel free to reject this if you'd rather stick with `tsup`. There are some other tsconfig options (see `tsconfig.base.json`) that I'd suggest to enable too if you decide to roll with this (some of them currently fail if enabled, would require a bit of refactoring).

ALSO: This is still a work in progress of course. I've also removed the `import from package.json` to load the version to keep and force/restrict the build root to `src`. This means we'd now need a small script or `prebuild` command to copy & replace the version in `src/errors/version.ts`. 

Happy to discuss and continue this further if you want to explore this more.


jxom:
Thanks for the PR! Love the super pragmatic approach, and agree that developing in an environment similar to the consumer's environment would be nice. Haven't had much time yet to dig closer into this, but from a high-level, I have a few questions (I am not too familiar with `tsc` as a bundler):

- Could we utilise TypeScript 5's `moduleResolution: "bundler"` with `tsup` to help resolve #244?
- How will bundling with `tsc` affect tree-shaking/is tree-shaking well supported?
- How will bundling with `tsc` affect bundle size, does it have some smarts around optimization here?
- How does `tsc` stack up with `tsup` in terms of bundle time? Is it still bearable?
- Is it at all possible to use it without adding `.js` extensions to everything?

Agree that declaration maps will definitely be a nice addition, but I don't think `tsup` supports them yet 😅. One thing we could do if we still want to roll with `tsup` is run a step after the bundling process to run `tsc --emitDeclarationOnly` or something.

fubhy:
Btw. for full transparency regarding my stance on publishing source maps & declaration maps... It is controversial and it's matter of taste. I'm personally leaning pretty strongly towards publishing source code & source maps & declaration maps together with the package.

You've got two options:

1. Publish source code, source maps and declaration maps all together
2. Do not publish source code, source maps or declaration maps at all

There's not really a middleground because publishing source maps without source code basically leaves you with references to files (in your sourcemaps) that aren't included in your package. That has the potential to trip up tooling that tries to parse your source maps (e.g. for stack traces, etc.) and can potentially break things.

So here are the benefits & downsides of publishing all of this with the package (my recommendation):

# Benefits

- Added utility in IDEs, e.g:
- Ability to not only jump to type declarations but also to the underlying source code
- Added utility for error handling & tracing services, e.g:
- Ability to create awesome error stack traces also for library-scoped errors
- Conveniently debug library related problems by browsing the included source code instead of having to clone the corresponding repository (at the right commit hash corresponding to the used package version)

# Downsides

- Increased package size

For a plain typescript library, the increase in package size isn't significant enough, imho, to really argue against doing this. If your src directory had any large binary (or other) files in it (e.g. images, etc.), I'd simply exclude those from being published.

Related discussion: https://github.com/googleapis/google-cloud-node/issues/2867

fubhy:
Btw. is there a discord server that you guys hang out in?

jxom:
Sorry, excuse my language of using the term “bundling”! I definitely meant “compiling” (I was writing that response half awake 😜).

Also definitely want to move away from the index/barrel imports from within viem itself as you said — we just recently discovered this has some nuances with tree-shaking (for example, importing a simple util pulls in 10kB gzipped or something when it should be much lower depending on the util). 

Open to adapting to the `.js` extension on imports as well!

Definitely down to publish declaration maps as well, no brainer in terms of DX of digging into viem source code as a consumer.

fubhy:
Superceded by #265 . Will come back to `NodeNext` in a separate, dedicated PR to reduce the noise from all the import changes.

PR #256, opened 2023-03-29
wip – feat: getEnsTextRecord

kesar:
In order to match ethers getAvatar() implementation

discussion: https://github.com/wagmi-dev/viem/discussions/255

### DIFF
--- a/site/docs/ens/actions/getEnsAvatar.md
+++ b/site/docs/ens/actions/getEnsAvatar.md
@@ -1,0 +1,111 @@
+---
+head:
+  - - meta
+    - property: og:title
+      content: getEnsAvatar
+  - - meta
+    - name: description
+      content: Gets avatar for specified name.
+  - - meta
+    - property: og:description
+      content: Gets avatar for specified name.
+
+---
+
+# getEnsAvatar
+
+Gets avatar for specified ENS.
+
+Calls `text(bytes, string)` on ENS Universal Resolver Contract to find avatar text record.
+
+## Usage
+
+::: code-group
+
+```ts [example.ts]
+import { publicClient } from './client'
+ 
+const ensAvatar = await publicClient.getEnsAvatar({
+  name: 'wagmi-dev.eth',
+})
+// 'wagmi-dev.eth'
+```
+
+```ts [client.ts]
+import { createPublicClient, http } from 'viem'
+import { mainnet } from 'viem/chains'
+
+export const publicClient = createPublicClient({
+  chain: mainnet,
+  transport: http()
+})
+```
+
+:::
+
+::: warning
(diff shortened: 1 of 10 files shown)


jxom:
Thanks for the PR! Agree that having an abstraction to fetch an ENS avatar will definitely be required in viem. 

From the stance of viem's vision of "atomic" Actions, this PR is fantastic. However, I feel like 95% of the consumers who will use this function will also want to resolve the image URL from: base64 metadata, ERC `"eip155:1/"` strings, IPFS url, SVGs, etc etc... Ethers does [some work](https://github.com/ethers-io/ethers.js/blob/main/src.ts/providers/ens-resolver.ts#L350-L484) internally for this, however, it's not "complete". I feel like the [ens-avatar repository](https://github.com/ensdomains/ens-avatar) does a much better job at doing this. 

I'd still like to go ahead with this PR, and instead repurpose it to generically support text record resolution instead. What do you think of making this `getTextRecord` instead of `getAvatar`?

kesar:
> Thanks for the PR! Agree that having an abstraction to fetch an ENS avatar will definitely be required in viem.
> 
> From the stance of viem's vision of "atomic" Actions, this PR is fantastic. However, I feel like 95% of the consumers who will use this function will also want to resolve the image URL from: base64 metadata, ERC `"eip155:1/"` strings, IPFS url, SVGs, etc etc... Ethers does [some work](https://github.com/ethers-io/ethers.js/blob/main/src.ts/providers/ens-resolver.ts#L350-L484) internally for this, however, it's not "complete". I feel like the [ens-avatar repository](https://github.com/ensdomains/ens-avatar) does a much better job at doing this.
> 
> I'd still like to go ahead with this PR, and instead repurpose it to generically support text record resolution instead. What do you think of making this `getTextRecord` instead of `getAvatar`? We are planning to look into a more complete avatar solution very soon as we will need this for the wagmi v1 upgrade. If you also want to take that on, happy to chat about it.

sure, I had a similar feeling when I was writing the implementation. Gonna work on that and make it more generic 👍🏻 

btw big kudos for the dev experience to contribute, everything worked perfect following the contribution documentation

jxom:
Thanks! Also, this PR is using the ENS Public Resolver. Can we instead use the UniversalResolver? Here is [example usage](https://github.com/wagmi-dev/viem/blob/main/src/actions/ens/getEnsAddress.ts#L60-L96) – we will need to swap out `addr` with `text` instead obviously.

kesar:
fixed using UniversalResolver and made it more generic to support other textRecords.

nand2:
Awesome! Thanks for this, I was looking for this :)

jxom:
Going to close this in favor of #259. Unforunately, there are conflicting implementations here, and we have to chose one. Probably less complicated if we merge #259. Will use the docs in this PR though, so I'll make sure to attribute you as the author of that commit. Thanks so much once again!

PR #257, opened 2023-03-30
feat: migrate from `@noble/secp256k1` to `@noble/curves/secp256k1`

jxom:
We should be using the audited [`@noble/curves`](https://github.com/paulmillr/noble-curves) library instead.


src/_test/utils.ts
```diff
@@ -186,23 +183,4 @@ export async function setBlockNumber(blockNumber: bigint) {
     jsonRpcUrl: process.env.VITE_ANVIL_FORK_URL,
   })
 }
-
-export function signTransaction(serializedTransaction: Hex, privateKey: Hex) {
```
jxom, line 190: Can just use `sign` directly in tests now.

tmm (approved):

PR #258, opened 2023-03-30
feat: `getEnsResolver`

jxom:
Adds `getEnsResolver`.


src/index.ts
```diff
@@ -25,6 +25,12 @@ export type {
   GetBytecodeParameters,
   GetBytecodeReturnType,
   GetChainIdReturnType,
+  GetEnsAddressParameters,
```
tmm, line 28: Not really an issue: Is it weird that we export action types from here, but not the actions?

tmm (approved):

src/index.ts
jxom: not sure why we don't actually lol, can make a future PR for it

PR #259, opened 2023-03-30
feat: getEnsText, getEnsAddress multicoin support, getEnsAvatar

TateB:
changes:
- `getEnsText()` based on #256, now using universalresolver
- `getEnsAvatar()` implemented, with full resolution
- added multicoin support to `getEnsAddress()` with optional `coin` input
- user needs to get the multicoin request item separately (similar to normalisation) with `multicoinFromName()` or `multicoinFromCoinId()`
- added basic resolver abis to abi constants file


src/actions/ens/getEnsText.test.ts
```diff
@@ -0,0 +1,107 @@
+import { expect, test } from 'vitest'
+import { optimism } from '../../chains'
+import { createPublicClient, http } from '../../clients'
+
+import { localHttpUrl, publicClient } from '../../_test'
+import { getEnsText } from './getEnsText'
+
+test('gets text record for name', async () => {
```
jxom, line 8: fyi: probably need to manually set block number in anvil as we currently fork on an older block number than the block number the new universal resolver was deployed on.

https://github.com/wagmi-dev/viem/pull/259/files#diff-6d705e586e88c47f940b3a9dc8604c61719133e8caa553a7d91e6deace56b915L9-L17

jxom (approved):

PR #265, opened 2023-03-30
chore: switch to tsc & publish source & declaration maps & related chores 

fubhy:
This should be easier to review and find consensus for. Will follow-up with the other steps after this one.

# Done
- [x] switch to `tsc` & remove `tsup`
- [x] publish declaration maps
- [x] publish source maps
- [x] publish source code
- [x] enable or highlight stricter type checks (see tsconfig.base.json)
- [x] bump tsconfig language environment to Node 16+ equivalents (we already have an implicit Node v17/18+ reliance due to `fetch`)
- [x] clean up package.json (remove unused deps, delete redundant deps like `@types/node`)
- [x] re-enable build-less local development for playgrounds
- [x] replace package.json version lookup with `version.ts`
- [x] research if we can make bundlephobia report better / smarter results for the (now larger) dist dir
- [x] all sorts of related chores & clean ups
- [x] bump `moduleResolution` to NodeNext and adjust imports

# Follow-up
- Add CI step for building & verifying examples & playground (check they are functional, track bundle sizes, etc.)

# Note
- I removed the .json import of package.json to determine the version. I did that for two reasons: a) to remove a few bytes from the resulting build output and b) to get `tsc` to use `src` as the source root. This means that we now need a build script for putting the right version into `src/errors/version.ts`. That script runs after the `changeset version` (or `npm set-version` for canary releases) call and *before* building the dist output.
~~- Examples are slightly less self-contained and "1:1 copyable" from the repo now because I added `tsconfig` path aliases there so they can be used with the raw `viem/src` and support hot-module replacement without requiring a build~~
- Examples do not have path aliases (so they remain self-contained). That means, however, that they require a build of the main `viem` package (`pnpm build`) to run them.


jxom:
Looks good! 

Might be worth basing this PR against the `0.2.0` branch as we have some substantial changes for the upcoming 0.2 release.

Also noticed that the proxy packages have been removed. Is there a reason for that? This means that CJS consumers won't be able to import from entrypoints (e.g. `viem/accounts`) right?
We might also want to keep the [auto-gen of `exports`](https://github.com/wagmi-dev/viem/blob/main/scripts/tsup.ts#L87-L127) script too.

fubhy:
> Might be worth basing this PR against the `0.2.0` branch as we have some substantial changes for the upcoming 0.2 release.

Ah, I see. Can do!

> We might also want to keep the [auto-gen of `exports`](https://github.com/wagmi-dev/viem/blob/main/scripts/tsup.ts#L87-L127) script too.

I didn't notice that script, whoops. I can bring that back and adjust it to generate exports in the "new" format if you think it's worth doing. I've got a similar bash script here (https://github.com/enzymefinance/enzyme-sdk/blob/main/packages/abis/scripts/abis.sh) to generate exports for all abis seperately: https://github.com/enzymefinance/enzyme-sdk/blob/main/packages/abis/package.json#L22

But that's because there's roughly ~120 of those :-P

> Also, do you know if there are any caveats to removing proxy packages? Is package.json `exports` well supported across all environments these days? I'm a bit hesitent to remove them sweat_smile

Since we essentially decided to go with Node v16+, I'd not be worried about that. It's definitely support since Node v12+. But I can see why you are worried about that! I can run some tests with different node version consumers?

fubhy:
Okay so it looks like currently these definitions here (https://github.com/wagmi-dev/viem/blob/main/package.json#L33-L47) are translated into `exports` with that script.

How about instead of generating those (which is a one-time task and basically just a matter of copy & paste), we instead test that they are correct and working in consumers? We can do that both with a test of the package.json contents (checking that `typesVersions` is correct and that there are no typos or mismatches between cjs vs. esm vs. type exports), etc. and we can also test them via the examples that would import from them (and break if they are not properly declared).

fubhy:
I rebased this PR on 0.2.0. Note that `lint` is not happy atm. but that's not introduced by me here :-).

fubhy:
@jxom How'd you like to solve the `version` problem for the error handling? I'd suggest we adjust the workflow so that we first tag the version in package.json, then run the script to take the newly tagged version and drop it into `src/errors/version.ts`, then run build, and then publish. That should work both for changesets and the canary release tagging.

fubhy:
> Also, do you know if there are any caveats to removing proxy packages? Is package.json exports well supported across all environments these days? I'm a bit hesitent to remove them sweat_smile

I just tested with Node v16 and the bottleneck there actually isn't the lack of proxy packages (exports works fine), but the lack of `fetch` ... obviously :-). I think it's reasonable to expect people to use Node 17+ though.

fubhy:
I've been playing around with the examples packages and node consumers for the past hour or so using the new pure tsc output. The results you get now from things like `vite-bundle-visualizer` are _really beautiful_ and much more insightful. I can confirm that tree-shaking works beautifully with this at least with vite. I'll run some comparisons between the 0.2.0 branch and this PR tomorrow and post results here, but the `fetching-transactions` example, yields this result for instance:

# Before:
✓ 20 modules transformed.
dist/index.html                 0.36 kB
dist/assets/index-38501da2.js  83.78 kB │ gzip: 26.74 kB

# After:
✓ 253 modules transformed.
dist/index.html                 0.36 kB
dist/assets/index-bfb45e50.js  79.36 kB │ gzip: 21.99 kB

Note how we are bundling much more granularly (253 modules vs. 20 pre-bundled modules).

package.json
```diff
@@ -2,131 +2,147 @@
...
+    "dist/**/*.js.map",
+    "dist/**/*.d.ts",
+    "dist/**/*.d.ts.map",
+    "src/**/*.ts",
+    "!src/**/*.test.ts",
+    "!src/**/*.test-d.ts",
+    "!src/**/*.bench.ts",
+    "!src/_test/**/*"
   ],
+  "type": "module",
```
jxom, line 16: Is this meant to be here?

fubhy: This is fine. It simply makes this a ESM module by default but because we also support CJS through our `require` exports and the `main` entry point, it works in the environments that we support (Node v18+ effectively due to the implicit `fetch` dependency). Would also work in Node v12.7+ fwiw though if we didn't have the `fetch` dep.

You an verify that this works by creating a basic `cjs` playground (e.g. copy from `bun` playground) that only has this tsconfig and changing it's own `type` (package.json` to `commonjs`):

```
{
  "include": ["index.ts"],
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "skipLibCheck": true
  }
}
```

It should work fine and you should also see that it imports the `cjs` entrypoints from there (e.g. add a `console.log` to the `dist/cjs/index.js` to confirm).

playgrounds/browser/src/App.tsx
```diff
@@ -1,11 +1,11 @@
-import { useState } from 'react';
-import 'viem/window';
+import { useState } from 'react'
```
fubhy, line 1: No idea why `rome` suddenly starts caring about the `playground` files. I don't see where that would be ignored?

fubhy: Ah, the `pnpm format` command atm. only targets `src/`. Let me change that ...

examples/blocks/watching-blocks/vite.config.ts
jxom, line 1: Think it would be better if these examples could be ran completely standalone. We need it to behave like that as they should be also runnable in StackBlitz (see the README of this example).

fubhy: Ok. Let's remove the path aliases from those then!

package.json
```diff
@@ -2,131 +2,150 @@
   "name": "viem",
   "description": "TypeScript Interface for Ethereum",
   "version": "0.1.23",
-  "scripts": {
```
jxom, line 5: Can we keep these scripts higher up? Feel like it's better to have it up here so contributors can get off the ground running quickly.

package.json
```diff
@@ -2,131 +2,150 @@
...
+      ]
+    }
+  },
+  "scripts": {
+    "anvil": "source .env && anvil --fork-url $VITE_ANVIL_FORK_URL --fork-block-number $VITE_ANVIL_BLOCK_NUMBER --block-time $VITE_ANVIL_BLOCK_TIME",
+    "bench": "vitest bench --no-threads",
+    "bench:ci": "CI=true vitest bench --no-threads",
+    "build": "pnpm run clean && pnpm run build:cjs && pnpm run build:esm+types && size-limit",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm+types": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm --declaration --declarationMap --declarationDir ./dist/types",
```
jxom, line 127: Should this be esnext?

```suggestion
    "build:esm+types": "tsc --project tsconfig.build.json --module esnext --outDir ./dist/esm --declaration --declarationMap --declarationDir ./dist/types",
```

package.json
fubhy: For sure. Good point

package.json
fubhy: There would be no difference in our case:https://www.typescriptlang.org/tsconfig#module

Hence why I chose the "lowest" boundary

src/clients/decorators/public.ts
```diff
@@ -256,7 +260,15 @@ export type PublicActions<
...
-export const publicActions = <
+// NOTE: The explicit type annotation is necessary to work around a type inferrence issue with
+// `moduleResolution` set to `NodeNext`. It might be possible to remove it after bumping the
+// `abitype` package to `NodeNext` too.
+export const publicActions: <
+  TTransport extends Transport = Transport,
+  TChain extends Chain | undefined = Chain | undefined,
+>(
+  client: PublicClient<TTransport, TChain>,
+) => PublicActions<TTransport, TChain> = <
```
fubhy, line 271: This is the original type issue that I also experienced downstream in my library. Now that we are bumping `viem` to NodeNext, it also occurs here, which means that the issue actually originates in `abitype`. It's not the end of the world to set up an explicit type annotation here (and one can also do that downstream to fix it there too), but it can likely be avoided by adjusting `abitype` similarly. I'd be more than happy to do that after a bit of a testphase for that in `viem` as already discussed with you guys!

fubhy:
@jxom Should I rebase this on `main`?

jxom:
@fubhy – Yep! Sounds good! Sorry, I didn't explicitly close this – must have automatically closed after I merged the `0.2.0` branch.

fubhy:
Yep github does that for you when you rugpull the base branch :-P. Will redo the commit history because rebase is pretty much impossible right now lol

src/clients/decorators/public.ts
roninjin10: Just trying to understand what this is for my own learning.   @fubhy if you get a chance coudl you explain the "why" here?   Why does type resolution fail for NodeNext talking to Node?

PR #282, opened 2023-04-03
chore: switch to tsc & related chores

fubhy:
@jxom Re-opening with just the tsc & formatting related cleanups / chores. So everything except the `NodeNext` bump. That's always a huge pain to rebase so I'm waiting for the definitive go-ahead to re-apply those changes!

In fact... Let's do that in a separate follow-up PR altogether if you don't mind ;-)


examples/blocks/fetching-blocks/tsconfig.json
jxom, line 1: Do the examples need to have a `tsconfig.node.json` config? Ideally these examples should be as minimal as possible to get the user off the ground running without questioning what the difference if between two tsconfigs.

examples/contracts/multicall/index.ts
```diff
@@ -1,42 +1,42 @@
-import { createPublicClient, http } from "viem";
-import { mainnet } from "viem/chains";
-import { wagmiContract } from "./contract";
+import { createPublicClient, http } from 'viem'
+import { mainnet } from 'viem/chains'
+import { wagmiContract } from './contract.js'
```
jxom, line 3: Drop the .js extensions for examples?

```suggestion
import { wagmiContract } from './contract'
```

package.json
```diff
@@ -6,19 +6,20 @@
     "anvil": "dotenv -- sh -c 'anvil --fork-url $VITE_ANVIL_FORK_URL --fork-block-number $VITE_ANVIL_BLOCK_NUMBER --block-time $VITE_ANVIL_BLOCK_TIME'",
     "bench": "vitest bench --no-threads",
     "bench:ci": "CI=true vitest bench --no-threads",
-    "build": "tsup && size-limit",
+    "build": "pnpm run clean && pnpm run build:cjs && pnpm run build:esm+types && size-limit",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
+    "build:esm+types": "tsc --project tsconfig.build.json --module es2015 --outDir ./dist/esm --declaration --declarationMap --declarationDir ./dist/types",
```
jxom, line 11: Should we add the ability to run this in `--watch` mode for the examples?

fubhy: Good idea indeed. Should be fine also in terms of performance with incremental mode.

tsconfig.base.json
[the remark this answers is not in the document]
fubhy: Yeah that makes sense.

package.json
```diff
@@ -6,19 +6,20 @@
     "anvil": "dotenv -- sh -c 'anvil --fork-url $VITE_ANVIL_FORK_URL --fork-block-number $VITE_ANVIL_BLOCK_NUMBER --block-time $VITE_ANVIL_BLOCK_TIME'",
     "bench": "vitest bench --no-threads",
     "bench:ci": "CI=true vitest bench --no-threads",
-    "build": "tsup && size-limit",
+    "build": "pnpm run clean && pnpm run build:cjs && pnpm run build:esm+types && size-limit",
+    "build:cjs": "tsc --project tsconfig.build.json --module commonjs --outDir ./dist/cjs && echo > ./dist/cjs/package.json '{\"type\":\"commonjs\"}'",
```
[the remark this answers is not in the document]
fubhy: It's generated after the build command and put inside the created `./dist/cjs` dir

examples/contracts/multicall/index.ts
fubhy: Ok, fair point. Let's just do `src` for now (I think there's a good chance that `nodenext` will catch on and hopefully eventually also reach applications, it has really grown on me tbh, but until then... indeed let's jsut do ./src)

examples/blocks/fetching-blocks/tsconfig.json
fubhy: I added those because the default `pnpm create vite` (as in the vite docs) adds it automatically too. But up to you. It's not strictly necessary (although it's technically corect because that vite config file is not part of the app tsconfig.

fubhy:
@jxom I opened #309 which is closer to what you are asking for with your feedback here (just does `nodenext` for ./src for now)

examples/blocks/fetching-blocks/tsconfig.json
jxom: Ah fair enough! Makes sense.

src/errors/utils.ts
```diff
@@ -1,7 +1,7 @@
 // @ts-ignore
 import { Address } from 'abitype'
-import pkg from '../../package.json'
```
tmm, line 3: Why doesn't this work anymore? `--resolveJsonModule`  not part of `tsconfig.build.json`?

scripts/updateVersion.ts
```diff
@@ -0,0 +1,9 @@
+import { readJsonSync, writeFileSync } from 'fs-extra'
+import path from 'path'
+
+// Writes the current package.json version to `./src/errors/version.ts`.
+const versionFilePath = path.join(__dirname, '../src/errors/version.ts')
+const packageJsonPath = path.join(__dirname, '../package.json')
+const packageVersion = readJsonSync(packageJsonPath).version
+
+writeFileSync(versionFilePath, `export const version = '${packageVersion}'`)
```
tmm, line 9: Might be nice to add this file to `.gitignore` or add a comment to it that it's autogenerated.

tmm (commented):
Overall really solid. Few thoughts!

- Ran through [`publint`](https://github.com/bluwy/publint) (`pnpm build && pnpx publint .`) and looks good!
- We [generated `package.json#exports` before](https://github.com/wagmi-dev/viem/blob/main/scripts/tsup.ts#L87). Not something that changes very often, but might be nice to add something like this back in since updating manually can lead to errors. Also, generating `package.json#typesVersions` could be nice.
- Could be nice if we could drop the `tsconfig.node.json` files and combine them with `tsconfig.json`

jxom (approved):

src/errors/utils.ts
fubhy: It does work but do you really want the entire package.json in the bundle? Also, it would mean that we'd lose the nice `src` => `dist` paths, and would instead get `dist/src/...` (because the source root, as far as tsc is concerned, would then be the repository root.

scripts/updateVersion.ts
fubhy: We need this for the tests to run (or are you saying everyone should generate that locally e.g. in postinstall?). A comment would be good indeed

fubhy:
@jxom @tmm Thank you for the patience with the back & fort here (me re-doing it all the time, sorry about that).

I re-did it now with the bare minimum changes required for `tsc`. But please don't merge it yet because I want to verify that this works for absolutely all import scenarios (including webpack 4 and the likes).

src/errors/utils.ts
tmm: > entire package.json in the bundle

good point! think we can do without that.

fubhy:
> * pnpm build && pnpx publint .

Btw. I like the idea that @jxom had the other day about having a very specific folder in `src` that hosts the `entrypoints` and represents basically the only place in src that has barrel files. We could then generate (and test) the exports based on those. What do you think? That could follow after the next "clean up barrel files" step?

scripts/updateVersion.ts
tmm: either postinstall or as-is now is fine. comment is only thing that really matters!

fubhy:
Oh, and I recently found https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/15 

I want to check our exports with that too first :-)

src/actions/getContract.test.ts
```diff
@@ -140,7 +140,7 @@ test('watchEvent', async () => {
 
 test('write', async () => {
   await expect(
-    contract.write.mint({
+    contract.write.mint([0n], {
```
fubhy, line 143: This change was needed after the update to typescript v5. Review carefully please!

Not sure what to do here... Seems that something happened in typescript v5 that now causes it to fall through to the second overload?

src/actions/getContract.test-d.ts
```diff
@@ -369,7 +369,7 @@ test('with and without wallet client `account`', () => {
   expectTypeOf(contractWithoutAccount.write.approve).parameters.toEqualTypeOf<
     [
       args: readonly [`0x${string}`, bigint],
-      options: { account: Account | Address },
+      options?: { account: Account | Address },
```
fubhy, line 372: This change was needed after the update to typescript v5. Review carefully please!

src/actions/getContract.test-d.ts
```diff
@@ -392,7 +392,7 @@ test('with and without wallet client `chain`', () => {
   expectTypeOf(contractWithoutChain.write.approve).parameters.toEqualTypeOf<
     [
       args: readonly [`0x${string}`, bigint],
-      params: { chain: Chain | null | undefined },
+      params?: { chain: Chain | null | undefined },
```
fubhy, line 395: This change was needed after the update to typescript v5. Review carefully please!

fubhy:
I've tried all examples & playground and tested the package.json exports & esm & cjs build outputs with both https://arethetypeswrong.github.io/  (using the cli from https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/15) and `publint` (that tool seems to be limited btw.!, it didn't point out some issues that https://arethetypeswrong.github.io/ pointed out).

The only remaining "issue" is the one that also Mark Erikson is experiencing in `redux-thunk`: Types for ESM are masqueraded as CJS types. That doesn't appear to be a problem though. Especially because we don't use default exports.

So I think this is ready to go? Just please carefully review the `src` and `test` changes (left comments)

src/utils/ens/normalize.ts
```diff
@@ -1,4 +1,4 @@
-import uts46 from 'idna-uts46-hx'
+import { toUnicode } from 'idna-uts46-hx'
```
jxom, line 1: Does this work from a pure-ESM consumer PoV? I was facing issues consuming this in a pure ESM environment as it could not import the named export.

fubhy: As far as I can tell it works, yeah. It works in the examples, in the playground and also in the sdk I'm building (just tested).

jxom: Interesting. This wasn’t working in the wagmi migration before. I’ll test this build against that and see if anything’s changed.

fubhy: /me *rubs hands* ... so are you saying there's a type issue in `wagmi` related to the tsconfig & build pipeline and I should tackle that next? :-P

jxom: It wasn’t a type issue, not sure if it would be related to the build. Occurred as I was running the tests in wagmi with Vitest. 
It was the `”idna-uts46-hx” is a CommonJS module, which may not support all module.exports as named exports.` error off the top of my head. Will test when I’m at computer next!

jxom: Wonder if it’s worth switching over to https://www.npmjs.com/package/@adraffy/ens-normalize

fubhy: > It wasn’t a type issue, not sure if it would be related to the build. Occurred as I was running the tests in wagmi with Vitest. It was the `”idna-uts46-hx” is a CommonJS module, which may not support all module.exports as named exports.` error off the top of my head. Will test when I’m at computer next!

Gotcha. Will try those tests too.

fubhy: Not really possible to test that right now without the `NodeNext` bump (fixing directory imports, adding .js extension). Without that, it indeed breaks, but that's expected (we are publishing .js code without adding the .js file extension because `tsc` does not rewrite those on build like esbuild / rollup / tsup do).

fubhy: I published a snapshot that includes the nodenext changes: viem@tsc-and-nodenext ...
I can confirm that I'm also seeing that error in the `wagmi` tests now: 

```SyntaxError: Named export 'toUnicode' not found. The requested module 'idna-uts46-hx' is a CommonJS module, which may not support all module.exports as named exports.```

Will debug further.

Comment (fubhy) [reply src/utils/ens/normalize.ts]: Good catch. I reverted that import change.

**Comment by fubhy** [comment]
@jxom There are some leftover imports & exports that are missing the `import type` keyword (so incorrectly exporting / importing types as source). Need to go through thouse first before this is (hopefully) finally ready.

**Comment by fubhy** [comment]
@jxom Seems to now work with `viem@0.0.0-tsc-and-nodenext-20230405123821`

**Comment by tmm** [comment]
Might be good to merge this into a feature branch so #309 can get in right after it before any releases happen (`main` does canary releases)

File: src/actions/getContract.test.ts (hunk 6 above)
Comment (jxom) [reply src/actions/getContract.test.ts:143]: Hmm yeah, seems like this has regressed. Will need to make it work with overloads.

File: src/actions/getContract.test-d.ts (hunk 8 above)
Comment (jxom) [reply src/actions/getContract.test-d.ts:395]: Looks like this regressed too (we need a chain).

File: src/actions/getContract.test-d.ts (hunk 7 above)
Comment (jxom) [reply src/actions/getContract.test-d.ts:372]: Looks like this regressed too (we need an account).

File: src/actions/getContract.test-d.ts (hunk 8 above)
Comment (jxom) [reply src/actions/getContract.test-d.ts]: Fixed here: https://github.com/wagmi-dev/viem/pull/282/commits/06ccb36a2f76c2918c61e9b84640cf76dfb610e6

File: src/actions/getContract.test-d.ts (hunk 7 above)
Comment (jxom) [reply src/actions/getContract.test-d.ts]: Fixed here: https://github.com/wagmi-dev/viem/pull/282/commits/06ccb36a2f76c2918c61e9b84640cf76dfb610e6

File: src/actions/getContract.test.ts (hunk 6 above)
Comment (jxom) [reply src/actions/getContract.test.ts]: Fixed here: https://github.com/wagmi-dev/viem/pull/282/commits/06ccb36a2f76c2918c61e9b84640cf76dfb610e6

**Comment by tmm** [comment]
Is bumping TypeScript required for this change or can we leave it as-is?

File: package.json [hunk 10]
```diff
@@ -6,19 +6,20 @@
...
+    "build:esm+types": "tsc --project tsconfig.build.json --module es2015 --verbatimModuleSyntax --outDir ./dist/esm --declaration --declarationMap --declarationDir ./dist/types && echo > ./dist/esm/package.json '{\"type\":\"module\"}'",
```
roninjin10, line 11: nice TIL verbatimModuleSyntax

fubhy:
> Is bumping TypeScript required for this change or can we leave it as-is?

It's not required, no. It's a result of my testing with the different typescript versions and when it hinted that there are problems with typescript 5 I decided to point it out by including it here.

The problem is that typescript is like a peer dependencies. It doesn't help if we ignore that problem here because if someone uses typescript 5 downstream (which they will, because that's what you get when you install it today), they'll surface this in their implementation too. Hence why I think we should test with different typescript versions. If just to make us aware of which gotchas there are so we can either decide to fix them or document them (and ask ppl to upgrade).

That said, I'm happy to remove typescript 5 from this PR. But we should definitely fix it :-)

fubhy:
Ah, I see @jxom already fixed it. Wonderful <3

roninjin10:
I generally think it's a bad idea to ever use tsc to build javascript.   It's almost always better to use pretty much any tool that doesn't typecheck, babel, esbuild, swc, for all pipelines (build, tests) and have .d.ts generation and typechecking as it's own seperate pipeline.   Pretty much everything is faster better and more futureproof if you set stuff up like this to start.

That said viem isn't really a monorepo (excluding examples and harnesses) and in this case removing tsup is removing an entire tool at little cost

fubhy:
> I generally think it's a bad idea to ever use tsc to build javascript. It's almost always better to use pretty much any tool that doesn't typecheck, babel, esbuild, swc, for all pipelines (build, tests) and have .d.ts generation and typechecking as it's own seperate pipeline. Pretty much everything is faster better and more futureproof if you set stuff up like this to start.
> 
> That said viem isn't really a monorepo (excluding examples and harnesses) and in this case removing tsup is removing an entire tool at little cost

I guess I just don't see the point of using something other than `tsc` for building & publishing both dist and types for a library. What's the downside other than slightly slower builds (which, if we are not talking about a large monorepo, is negligible and also not a problem considering that during development, you are not doing builds (path aliases, etc.)). 

Or are you concerned about the quality & future-proofness of the output of `tsc`?

I'm personally not concerned about that and would also argue that with the current state of JS/TS, you'd generally have:

1. a Node / Bun / Deno consumer where bundling, even for the consuming application, is mostly not necessary or taken care of by the consumer
2. A web application in which case you can expect it to use some form of framework or bundler that is taking care of the necessary steps to produce the right compatibility for the expected end user browsers

So TL;DR my stance on this is the exact opposite: TypeScript libraries should generally not need a bundler *at all*

fubhy:
That said, I absolutely want `tsc` to become better at building full packages end to end. The dance we have to do atm. to produce cjs, esm and types and publish them properly with just the right package.json properties (and getting punished for the slightest mistakes), is *not great*.

roninjin10:
Also you mentioned it yourself but in general building packages tsc is simpler to use than something like tsup until it isn't.  The moment I want to do anything like publish both esm and cjs, or consume esm in a cjs project, ect. tsc stops being the easiest tool to use

fubhy:
I hear you. I agree that this can become a huge issue for large monorepos or when building larger application bundles at scale. I've been there too and agree with that. Especially the point about blocking CI pipelines needlessly.

What I've done to solve that in the past, which admittedly only works for unit & local integration tests, is to link the sources together (e.g. via vitest / jest module resolution).

fubhy:
> Also you mentioned it yourself but in general building packages tsc is simpler to use than something like tsup until it isn't. The moment I want to do anything like publish both esm and cjs, or consume esm in a cjs project, ect. tsc stops being the easiest tool to use

True. You can see that in this PR and how much fine tuning it took to get it set up properly. I wish `tsc` provided a complementary packaging tool & package linting tool. Luckily, https://arethetypeswrong.github.io/ exists now and is maintained by `ts` core devs. Maybe that'll eventually lead to better utils (both for linting packages and producing them in the first place). Fingers crossed.

tmm:
> That said, I'm happy to remove typescript 5 from this PR. But we should definitely fix it :-)

Let's keep it. We need to bump TypeScript for abitype and wagmi at some point anyway so viem can lead the way. (Also, TS doesn't follow semver and 5 doesn't have any [breaking changes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#breaking-changes-and-deprecations) we care about.)

src/actions/getContract.ts
```diff
@@ -587,11 +587,10 @@ type GetWriteFunction<
...
   // For making `options` parameter required if `TAccount` or `TChain` is undefined
-  IsOptionsRequired = [IsUndefined<TAccount>, IsUndefined<TChain>] extends
-    | [true, false]
-    | [false, true]
-    ? true
-    : false,
+  IsOptionsRequired = Contains<
+    [IsUndefined<TAccount>, IsUndefined<TChain>],
+    true
+  >,
```
tmm, line 593: Weird that this stopped working in new version of TypeScript. Probably better to use `Or` here instead.

```suggestion
  IsOptionsRequired = Or<
    IsUndefined<TAccount>, IsUndefined<TChain>,
  >,
```

Code for `Or`:

```ts
/**
 * @description Evaluates boolean "or" condition
 * @param T - Type to check
 * @param U - Type to check
 * @example
 * type Result = Or<true, false>
 * //   ^? type Result = true
 */
type Or<T, U> = T extends true ? true : U extends true ? true : false
```

[TypeScript Playground (with tests)](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgUwB5mQYxgFQJ4YDyAZnAL5zFQQhwDkAbsDMgM4x0BQnA9AFR9OcPnAACAEzaYowMDGAQAdnACiDAIYAbAK7qWrOACMIETcnXKARNEtxMS8cwWKhI0WHVR1tHHAC0cPgYcDAQdgAWWADWrmIeXrQAqv6BBMghYZiRmDHCYmjeYGaxMGlwAEps2prwALxwhFAAPDBQ2sgANJRarMgAfLE8PHAjAHoA-CFllazVdSFtyK48nKXBjU04XYl9cPW+aCyK4gat7XCTZ+kAXHDJh8jHp4sXC+e3xD1LnGgY2EHIEhNDZXLpXPp9AAUAEoAHShFQAR10mgBQPBMJ+6CwuDSQJBiy6n00vQhMPhECRKLRxBaiyh0Kxf1xRFpBPaRK+ZLhCORWhpdPaDKZOIF7M6cEUyAYyCg3IpVP5eNpGMZv1FyuBzWJvU5JP6DIVfNRmp1-UxQA)

jxom: Think `Contains` might be more future-proof (beyond two generics)?

jxom: Could do an array to `Or` actually!

tmm: Works! Good points. Fine with either.

PR #289, opened 2023-04-04
chore: pin dependencies, make external exports ESM friendly

jxom:
The `@scure/bip39/wordlists/*` & `idna-uts46-hx` exports seemed to be breaking for ESM consumers as they are CJS modules (named imports don't work well here). This PR converts the named imports to default imports, and extracts the named function accordingly.


package.json
```diff
@@ -128,16 +128,15 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "dependencies": {
-    "@ensdomains/address-encoder": "^0.2.21",
```
jxom, line 131: Didn't end up using this.

tmm (approved):

PR #290, opened 2023-04-04
fix: resolve 0x-prefixed ENS names

holic:
0x-prefixed names (i.e. `0xyoshi.eth`) fail because somewhere in the stack it's assumed that strings starting with `0x` are hex


src/actions/ens/getEnsAddress.test.ts
```diff
@@ -15,12 +15,20 @@ test('gets address for name', async () => {
 
 test('name without address', async () => {
   await expect(
-    getEnsAddress(publicClient, { name: 'unregistered-name.eth' }),
+    getEnsAddress(publicClient, { name: 'unregistered--name.eth' }),
```
holic, line 18: the previous one was resolving for me and failing this test

TateB:
a name can still technically be valid hex, instead of modifying the `toBytes()` func i think you should just use `stringToBytes()` in `packetToBytes()`

e.g.
```ts
export function packetToBytes(packet: string): ByteArray {
  function length(value: string) {
    if (value === '.' || value === '..') return 1
    return stringToBytes(value.replace(/^\.|\.$/gm, '')).length + 2
  }

  const bytes = new Uint8Array(length(packet))
  // strip leading and trailing `.`
  const value = packet.replace(/^\.|\.$/gm, '')
  if (!value.length) return bytes

  let offset = 0
  const list = value.split('.')
  for (let i = 0; i < list.length; i++) {
    const encoded = stringToBytes(list[i])
    bytes[offset] = encoded.length
    bytes.set(encoded, offset + 1)
    offset += encoded.length + 1
  }

  return bytes
}
```

holic:
> a name can still technically be valid hex, instead of modifying the `toBytes()` func i think you should just use `stringToBytes()` in `packetToBytes()`

makes sense!

I feel like it still makes sense for utils like `toBytes` and `isHash` to lean on `isHex`, though? will let @jxom or @tmm weigh in on that (and can do separately if it makes sense)

jxom (approved):

PR #293, opened 2023-04-04
fix: misc ens bugs

TateB:
fixes:
- when ens resolver address func returns with null address, should return null from function rather than address
- if `addr()` does not exist on resolver, return null rather than trying to decode null result
- if `text()` does not exist on resolver, return null rather than trying to decode null result


src/actions/ens/getEnsAddress.ts
```diff
@@ -68,9 +69,14 @@ export async function getEnsAddress<TChain extends Chain | undefined,>(
     blockNumber,
     blockTag,
   })
-  return decodeFunctionResult({
+
+  if (res[0] === '0x') return null
```
jxom, line 73: Can we add a test for this line?

src/actions/ens/getEnsAddress.ts
```diff
@@ -68,9 +69,14 @@ export async function getEnsAddress<TChain extends Chain | undefined,>(
...
+
+  const address = decodeFunctionResult({
     abi: singleAddressResolverAbi,
     functionName: 'addr',
     data: res[0],
   })
+
+  return trim(address) === '0x' ? null : address
```
jxom, line 81:
```suggestion
  return trim(address) === '0x0' ? null : address
```

Can we add a test for this line?

src/actions/ens/getEnsText.ts
```diff
@@ -78,6 +78,8 @@ export async function getEnsText<TChain extends Chain | undefined>(
     blockTag,
   })
 
+  if (res[0] === '0x') return null
```
jxom, line 81: Can we add a test to this line?

PR #295, opened 2023-04-04
feat: return discriminated union type from decodeFunctionData

fubhy:
I'm probably missing sth. here as I've not spent that much time with `abitype` yet. 

Btw., I cannot emphasize enough how much you guys are knocking it out of the park here). Test suite, type harness, etc. everything is such a pleasant experience to work with. Refactoring, writing tests, adding features, it's hands down the easiest / lowest barrier of entry experience I've ever had to contributing to a library.

ref: https://github.com/wagmi-dev/viem/discussions/264

### DIFF
--- a/src/utils/abi/decodeFunctionData.test.ts
+++ b/src/utils/abi/decodeFunctionData.test.ts
@@ -106,7 +106,7 @@ test('getVoter((uint256,bool,address,uint256))', () => {
     functionName: 'getVoter',
   })
 
-  expectTypeOf(result).toMatchTypeOf<{
+  expectTypeOf(result).toEqualTypeOf<{
     functionName: 'getVoter'
     args: readonly [
       {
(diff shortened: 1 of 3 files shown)


src/utils/abi/decodeFunctionData.test.ts
```diff
@@ -58,43 +58,43 @@ test('bar(uint256)', () => {
...
-          name: 'getVoter',
-          outputs: [],
-          stateMutability: 'nonpayable',
-          type: 'function',
-        },
-      ] as const,
-      data: '0xf37414670000000000000000000000000000000000000000000000000000000000010f2c0000000000000000000000000000000000000000000000000000000000000001000000000000000000000000a5cc3c03994db5b0d9a5eedd10cabab0813678ac0000000000000000000000000000000000000000000000000000000000000029',
-    }),
-  ).toEqual({
+  const result = decodeFunctionData({
```
jxom, line 61: Can we add a test of what the type will look like with more than one ABI item?

fubhy: Yep, good idea.

fubhy: @jxom Added!

src/utils/abi/decodeFunctionData.ts
```diff
@@ -1,21 +1,37 @@
...
+import type { GetFunctionArgs, Hex } from '../../types'
 import { slice } from '../data'
 import { getFunctionSelector } from '../hash'
 import { decodeAbiParameters } from './decodeAbiParameters'
 import { formatAbiItem } from './formatAbiItem'
 
-export type DecodeFunctionDataParameters = {
-  abi: Abi | readonly unknown[]
+export type DecodeFunctionDataParameters<
+  TAbi extends Abi | readonly unknown[],
```
jxom, line 11:
```suggestion
  TAbi extends Abi | readonly unknown[] = Abi,
```

src/utils/abi/decodeFunctionData.ts
```diff
@@ -1,21 +1,37 @@
...
+> = {
+  abi: TAbi
   data: Hex
 }
 
-export function decodeFunctionData({
+export type DecodeFunctionDataReturnType<
+  TAbi extends Abi | readonly unknown[],
```
jxom, line 18:
```suggestion
  TAbi extends Abi | readonly unknown[] = Abi,
```

jxom (approved):
Just need a changeset, and we are good to go!

jxom (approved):

PR #309, opened 2023-04-04
chore: switch to nodenext module resolution

fubhy:
@jxom Alright, let's do this then! You raised a good point, let's just do `src` and not migrate the examples over. Let's leave all that untouched for now.

1. Migrate to nodenext module resolution and fix the surfaced type errors (ironically, I introduced an inferrence issue myself with the discriminated union `parseFunctionData` PR ;-))
2. Follow-up PR: Fix barrel imports
3. Follow-up PR: Let me re-propose the switch to `tsc` seperately (and sourcemaps, etc.) and we can evaluate that one in complete isolateion

How does that sound? :-)

This entire PR was done by my codemod: https://github.com/fubhy/node-to-nodenext


jxom:
Sounds good! Any particular reason for the implementation order? What if we:

1. Merge #282 without "NodeNext" resolution
2. Follow-up: Add "NodeNext" resolution
3. Follow-up: Fix barrel imports

fubhy:
Yeah we can also do that. Up to you. Re-creating this PR whenever we want it is now a matter of 10 seconds with the codemod so no hard feelings if we do it the other way round :)

jxom:
AI is the future ^

tmm:
I, for one, welcome our PR-summary AI overlords.

PR #320, opened 2023-04-06
feat: eip-1271

janek26:
Implements 2 new public actions:

- verifyMessage
- verifyTypedData

which have the same interface as their `utils` counterparts, but also support [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271)

Added some hints in the documentation (website and JSDoc) for developers which method they want to choose and why.

### <samp>🤖 Generated by Copilot at 0f87d3d</samp>

This pull request adds two new public actions, `verifyMessage` and `verifyTypedData`, to the public client and the public actions module. It also adds the documentation for these actions in the `site/docs` folder and updates the sidebar accordingly. Additionally, it updates the tests and the constants for the public actions, and imports the smart account contract abi for testing ERC-1271 verification.

---

## PR-Codex overview
This PR adds support for Contract Wallet signature verification (EIP-6492) via `publicClient.verifyMessage` & `publicClient.verifyTypedData`.

### Detailed summary
- Added support for Contract Wallet signature verification (EIP-6492) via `publicClient.verifyMessage` & `publicClient.verifyTypedData`
- Added `smartAccountAbi` and `universalSignatureValidatorAbi` to `abis.ts`
- Added `verifyMessage` and `verifyTypedData` functions to `verifyMessage.ts` and `verifyTypedData.ts`, respectively
- Updated `createPublicClient.test.ts` and `public.ts` to include `verifyMessage` and `verifyTypedData` in `PublicActions` type
- Added tests for `verifyMessage` and `verifyTypedData` in `verifyMessage.test.ts` and `verifyTypedData.test.ts`, respectively.

> The following files were skipped due to too many changes: `src/actions/public/verifyMessage.ts`, `src/clients/decorators/public.test.ts`, `src/actions/public/verifyTypedData.ts`, `src/actions/public/verifyHash.ts`, `contracts/src/DeploylessUniversalSigValidator.sol`, `site/docs/utilities/verifyTypedData.md`, `site/docs/actions/public/verifyMessage.md`, `site/docs/utilities/verifyMessage.md`, `src/constants/contracts.ts`, `src/actions/public/verifyHash.test.ts`, `site/docs/actions/public/verifyTypedData.md`


src/clients/decorators/public.ts
```diff
@@ -308,6 +320,8 @@ export const publicActions = <
   multicall: (args) => multicall(client, args),
   readContract: (args) => readContract(client, args),
   simulateContract: (args) => simulateContract(client, args),
+  verifyMessage: (args) => verifyMessage(client, args),
+  verifyTypedData: (args) => verifyTypedData(client, args),
```
fubhy, line 324: I'm wondering how this is going to affect tree shaking & bundle size if we keep adding stuff to the client factories. I don't have a counter proposal for a different pattern but the "fat client" approach kindof goes against our goal of keeping consumer bundle sizes small.

Maybe wen can split client factories into smaller chunks or encourage people to "roll their own" client?

fubhy (commented):
This looks really solid! Nice work.

src/clients/decorators/public.ts
jxom: viem was originally designed without actions on the clients at all. It was not until after feedback that we decided to attach actions to remove extraneous imports. 

You can still use viem in a “tree-shakable” way, check out the description of [this PR](https://github.com/wagmi-dev/viem/pull/77).

@tmm is working on some documentation for this approach (as well as tweaking the types to work better with it) in this open PR: #320

fubhy: Gotcha. I think for library authors in particular (libraries building on top of viem)  that should be the preferred pattern. Maybe also for some apps tbh. Documentation hinting at that would be great! Ty :)

jxom (commented):
Looks pretty good so far! Haven't looked at this too much in-depth yet, and will properly review this after the concerns I raised above have been addressed!

site/docs/actions/public/verifyMessage.md
```diff
@@ -0,0 +1,114 @@
...
+  - - meta
+    - property: og:description
+      content: Verifies if a signed message is valid for a given address.
+---
+
+# verifyMessage
+
+Verifies if a signed message is valid for a given address.
+
+::: info 💡 Why should I use this instead of `utils.verifyMessage`?
```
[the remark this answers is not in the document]
roninjin10: +1 my feedback here is basically this obfuscates what is happening too much via doing multiple things at once

src/utils/signature/verifyMessage.ts
```diff
@@ -10,6 +10,13 @@ export type VerifyMessageParameters = {
 
 export type VerifyMessageReturnType = boolean
 
+/**
+ * Can be used to verify a message signed by a private key
+ *
+ * @private Should not be used directly, please use `publicClient.verifyMessage` instead
+ * @param parameters
+ * @returns true if the message was signed by the private key of provided address (EOA)
+ */
```
roninjin10, line 19: This is awesome you went the extra mile to add jsdoc comments here

src/constants/abis.ts
```diff
@@ -81,3 +81,17 @@ export const singleAddressResolverAbi = [
...
+    name: 'isValidSignature',
+    type: 'function',
+    stateMutability: 'view',
+    inputs: [
+      { name: 'hash', type: 'bytes32' },
+      { name: 'signature', type: 'bytes' },
+    ],
+    outputs: [{ name: '', type: 'bytes4' }],
+  },
+] as const
```
roninjin10, line 97: nit unrelated to this pr these abis are way easier to discover if they were in different files

roninjin10 (commented):
I'm really excited about this pr thank you so much for doing this!

src/utils/signature/verifyMessage.ts
fubhy: Agreed. I think we should also add comments to the type definitions for function parameters. Particularly object parameters with multiple properties. That helps a great deal when typing those out in an IDE and getting popups with little doc snippets for those.

janek26:
Hey all! Just letting you know I’ll have a busy weekend and get back to you all after eastern! Enjoy everyone 🐰🥚

src/utils/signature/verifyMessage.ts
tmm: Big agree! This is on my list to do for the whole library, but happy to accept contributions. Put together some [basic guidelines](https://www.craft.do/s/tELgJ5UFsTjIzy) so things can be consistent.

src/clients/decorators/public.ts
janek26: I will resolve this convo, feel free to spin it out into an issue or discussion :)

src/utils/signature/verifyMessage.ts
janek26: updated the docs to match the guidelines

src/actions/public/verifyTypedData.ts
```diff
@@ -20,6 +20,23 @@ export type VerifyTypedDataParameters<
...
+ * @param client - The public client
+ * @param parameters - Object containing the typed data, signature and address to verify, plus optional blockTag and blockNumber for the onchain call
+ * @param parameters.address - The address to verify the typed data for
+ * @param parameters.signature - The signature to verify
+ * @param parameters.message - The typed data message
+ * @param parameters.primaryType - The typed data primary type
+ * @param parameters.types - The typed data types
+ * @param parameters.domain - The typed data domain
+ * @param parameters.blockTag - The block tag to use for the onchain call
+ * @param parameters.blockNumber - The block number to use for the onchain call
```
fubhy, line 37: What's the consensus for TSDoc / JSDoc when it comes to object parameters & documenting properties on those? I'd think that it would be nicer to document this on the underlying types (e.g. `VerifyTypedDataParameters` in this case) and then just enhance this with function specific documentation (if that's necessary) on the function that uses this as a parameter. That way, the documentation also appears in places where the type is used standalone.

I don't think this PR has to get that perfectly right btw. but when we go through the library to add those at some point (./cc @tmm), I think we should do that "inside out" (start with the underlying types, add tsdocs there, then enhance the function param types on top where (and only if) necessary.

site/docs/actions/public/verifyMessage.md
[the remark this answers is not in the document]
janek26: I believe there should be a single method for developers to verify a message or typed data. In my opinion, developers should not be concerned with how the signature is verified, but rather that it is done correctly. I want to prevent dapps from not supporting ERC4337 accounts, simply because they do not explicitly check with both types of signature verification.

```diff
- if (verifyMessage(msg) || publicClient.verifyContractMessage(msg)) {
+ if (publicClient.verifyMessage(msg)) {
  // go on
}
```

Achieving this may not be possible without either "deprecating" or, at the very least, clearly encouraging developers to use the alternative method instead of the verify methods that rely on address reconstruction.

I disagree with the notion that too much is happening in one method. If the goal of that method is to verify a signature, it should perform all the necessary tasks to do so. As I previously mentioned, I think it makes little sense to divide the logic into contract-related and EOA-related aspects, as doing so could lead to reduced smart wallet support. Additionally, the average developer should not be concerned with *how* the method achieves its goal, but rather that it does.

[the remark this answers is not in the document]
roninjin10: I think these are great points.   It is nice to have a method that “just works” and I as an application developer do not have to think about much beyond my goal.  

I think there is more to consider though

- what about libraries using viem as a meta library?   Shouldn’t they have the low level control?
- Do developers get the luxury of not caring about internals in non happy cases?
- Viem has a pretty consistent pattern of not doing magic and is super easy to use if you understand Ethereum.   Everything feels 1 to 1 with the core underlying apis like this is simply an Ethereum rpc sdk   Is it worth breaking that pattern here?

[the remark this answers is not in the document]
janek26: My biggest concern is that dapp developers will need to adopt their dapps to be compatible with smart contract wallets, as we had that problem before on a solidity level, and it sometimes persists until today.

- I'm a fan of exposing more low level functions to library developers, maybe through exporting a public action that does not get attached to the fat client object, as libraries should import these to take advantage of tree shaking.

- When verifyMessage consumes the client, I think it's quite obvious for a developer that he would need to take care of connectivity issues etc, just as with every other method in the client. 

- Keeping that 1 to 1 mapping for "people who know Ethereum" will be quite hard imo, imagine a Wallet instance, that instead of managing an EOA, implements and manages an ERC4337 wallet. It would probably take care of building and sending user operations etc, without requiring the developer to touch any of that. Developers would just know they have a wallet interface they can use, not caring about the implementation.

In the end it's depending on what's the goal for viem, and how we plan to target these different audiences (dapp developer, library developer, backend developer etc)

src/actions/public/verifyTypedData.ts
janek26: I think this is out of scope for this PR and should be discussed in a seperate github issue or discussion. This will also lead to more visibility for the question :)

tmm: Agree. I can clean it up to align with my work from earlier today

https://github.com/wagmi-dev/viem/pull/347

site/docs/actions/public/verifyMessage.md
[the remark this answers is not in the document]
jxom: Have been thinking about this more, and I agree. While viem is low-level, modular, 1-to-1, etc – we should make a reasonable effort to abstract (sensibly) for stuff like this. I think I'm more on the fence of _encouraging_ the user to use the Action instead of the Utility, instead of deprecating the Utility. A few thoughts:

Can we utilize [`getBytecode`](https://viem.sh/docs/contract/getBytecode.html) to determine if the wallet is a contract? This will break out of the EOA branching faster (as `eth_call` + abi encoding/decoding is an expensive call) and make the module more performant.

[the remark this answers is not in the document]
janek26: Sure! Happy to do so, I just thought 1 network request is better than 2.
You sure that `eth_call` is more expensive? I would expect it to return fast when there is no bytecode. Also, the return of getBytecode can be massive, no? Not sure that's worth it in low bandwidth environments.

I usually consider computation cheap and network requests/traffic expensive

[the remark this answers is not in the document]
jxom: Might be worth benchmarking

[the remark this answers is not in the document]
janek26: What do you think about https://github.com/AmbireTech/signature-validator/blob/main/index.ts and ERC-6492? Should we maybe just use that eth_call to handle all 3 cases in the EVM and just rely on the return?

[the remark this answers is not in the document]
jxom: That's pretty cool! Would probably fit a lot better – would be nice to support 6492 and resolve #337.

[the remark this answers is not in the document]
roninjin10: I think this is still a major concern but it could be handled with really good error messages and documentation 

“Do developers get the luxury of not caring about internals in non happy cases?”

janek26:
updated the PR

It's now using a deployless universal verifier contract, so it's sending 1 `eth_call` request and gets back a result for any kind of account (ERC-6492 -> ERC-1271 -> EOD in this sequence). 
Shoutout to @AmbireTech for the awesome ERC and the example implementation.

imo this is ready for reviews/merge 🙏

site/docs/actions/public/verifyMessage.md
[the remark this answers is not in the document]
janek26: Solved using the mentioned solution. Feel free to reopen/create a new convo as needed

PR #347, opened 2023-04-10
docs: tsdoc - public actions

tmm:
Adding [TSDoc](https://tsdoc.org) to all the things!

If anyone wants to help out, follow the rules laid out here: https://www.craft.do/s/tELgJ5UFsTjIzy (I'm currently going through `src/actions/public` so pick another area.)

image


tmm:
Thinking we probably want to [strip comments](https://www.typescriptlang.org/tsconfig#removeComments) from runtime source as they are increasing size (and just keep in declaration output).

https://github.com/wagmi-dev/viem/pull/347#issuecomment-1502030711

fubhy:
> Thinking we probably want to [strip comments](https://www.typescriptlang.org/tsconfig#removeComments) from runtime source as they are increasing size (and just keep in declaration output).
> 
> [#347 (comment)](https://github.com/wagmi-dev/viem/pull/347#issuecomment-1502030711)

Yeah that makes sense. We can also strip internals while we are at it.

fubhy:
Whoops, drunken misclick sorry

jxom (approved):

PR #353, opened 2023-04-11
chore: enable noImplicitOverride, noUnusedLocals and noUnusedParameters

fubhy:
I left a few `TODO` comments in there, please check those out. Not sure about those :)

... I'm checking the other tsconfig rules that I left `TODOs` for in the `TS5` upgrade PR.

### DIFF
--- a/src/errors/contract.ts
+++ b/src/errors/contract.ts
@@ -225,7 +225,7 @@ export class ContractFunctionZeroDataError extends BaseError {
 }
 
 export class RawContractError extends BaseError {
-  code = 3 // TODO: What's this code?
+  code = 3
   override name = 'RawContractError'
 
   data?: Hex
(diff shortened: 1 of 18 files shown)


src/errors/contract.ts
```diff
@@ -225,8 +225,8 @@ export class ContractFunctionZeroDataError extends BaseError {
 }
 
 export class RawContractError extends BaseError {
-  code = 3
-  name = 'RawContractError'
+  code = 3 // TODO: What's this code?
```
jxom, line 228: This comes from the node.
```suggestion
  code = 3 
```

src/errors/request.ts
```diff
@@ -159,8 +159,8 @@ export class JsonRpcVersionUnsupportedError extends RpcRequestError {
 }
 
 export class UserRejectedRequestError extends RequestError {
-  name = 'UserRejectedRequestError'
-  code = 4001
+  override name = 'UserRejectedRequestError'
+  code = 4001 // TODO: This has a code, should it be a RpcRequestError?
```
jxom, line 163: Yeah, we wanted consumers (ie. wagmi) to be able to pass in any arbitrary error, as it might not come from a network request (`RpcRequestError` assumes it was a network shaped error). Raised a PR at #355 to reduce ambiguity and make it more consistent.

PR #360, opened 2023-04-12
[WIP] chore: enable `exactOptionalPropertyTypes` and `noUncheckedIndexedAccess`

fubhy:
null

---

### <samp>🤖 Generated by Copilot at 8256dac</samp>

This pull request makes several changes to the type definitions of various actions and parameters in the `src` folder. The main purpose of these changes is to make the optional properties of the types more explicit by adding `| undefined` to their types, and to simplify the usage of the actions by allowing users to omit some of the properties and use default values instead. These changes affect the actions related to contracts, ENS, and public methods.


tmm:
@fubhy going to close with all these conflicts - might be better to start fresh.

PR #363, opened 2023-04-12
feat: add inference to `getLogs` `event`

tmm:
Adds type inference to `getLogs` `event` property

https://github.com/wagmi-dev/viem/discussions/359#discussioncomment-5594433

---

### <samp>🤖 Generated by Copilot at 3f8d114</samp>

This pull request enhances the type inference of the `getLogs` function, which fetches logs from a smart contract event. It uses the `Narrow` and `AbiEvent` types from `abitype` to narrow down the possible types of the `event` parameter and the returned logs. It also adds some type tests using `vitest` and documents the changeset in a markdown file.

### DIFF
--- a/.changeset/seven-panthers-admire.md
+++ b/.changeset/seven-panthers-admire.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added inference to `getLogs` `event` type.
--- a/src/actions/public/getLogs.test-d.ts
+++ b/src/actions/public/getLogs.test-d.ts
@@ -1,0 +1,96 @@
+import { expectTypeOf, test } from 'vitest'
+
+import { publicClient } from '../../_test/index.js'
+import { getLogs } from './getLogs.js'
+import type { AbiEvent } from 'abitype'
+
+test('event: const assertion', async () => {
+  const event = {
+    inputs: [
+      {
+        indexed: true,
+        name: 'from',
+        type: 'address',
+      },
+      {
+        indexed: true,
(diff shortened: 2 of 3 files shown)


fubhy:
Imho, there's also another issue with `getAbiItem`: We shouldn't specify generics to override the return type of functions. (`TReturnType` in this case).

This woudl be better imho:

```typescript
export function getAbiItem<
  TAbi extends Abi | readonly unknown[],
  TItemName extends string,
>({
  abi,
  args = [],
  name,
}: GetAbiItemParameters<TAbi, TItemName>): GetAbiItemReturnType<
  TAbi,
  TItemName
> {
  const abiItems = (abi as Abi).filter((x) => 'name' in x && x.name === name)

  if (abiItems.length === 0) return undefined as any

  for (const abiItem of abiItems) {
    if (!('inputs' in abiItem)) continue
    if (!args || args.length === 0) {
      if (!abiItem.inputs || abiItem.inputs.length === 0) return abiItem as any
      continue
    }
    if (!abiItem.inputs) continue
    if (abiItem.inputs.length === 0) continue
    const matched = (args as readonly unknown[]).every((arg, index) => {
      const abiParameter = 'inputs' in abiItem && abiItem.inputs![index]
      if (!abiParameter) return false
      return isArgOfType(arg, abiParameter as AbiParameter)
    })
    if (matched) return abiItem as any
  }
  return abiItems[0] as any
}
```

tmm:
@fubhy nice catch! Want to make that a separate PR?

PR #365, opened 2023-04-12
fix: do not use generic type variables for getAbiItem return type

fubhy:
It's imho good practice to avoid using generic type variables directly in return types. It creates a chicken & egg scenario when the return of such a function is then used in another generic function's input as seen in https://github.com/wagmi-dev/viem/discussions/359

---

### <samp>🤖 Generated by Copilot at db822ea</samp>

This pull request fixes the `getAbiItem` function to simplify its type signature and return type. It also adds a changeset file to document the patch version update and the bug fix for the `viem` package.

### DIFF
--- a/.changeset/silly-ravens-cross.md
+++ b/.changeset/silly-ravens-cross.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Fixed getAbiItem to not use a generic type variable for the return type
+Fixed `getAbiItem` to not use a generic type variable for the return type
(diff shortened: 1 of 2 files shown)


.changeset/silly-ravens-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed getAbiItem to not use a generic type variable for the return type
```
tmm, line 5:
```suggestion
Fixed `getAbiItem` to not use a generic type variable for the return type
```

tmm (approved):

PR #370, opened 2023-04-13
feat: allow `decodeFunctionData` to accept a single abi function

fubhy:
Mild annoyance that I ran into a couple of times... Would be great if we didn't have to wrap single abi declarations in an array needlessly?

@tmm @jxom If you agree with this, we could also do this for `decodeEventLog`, etc.

---

### <samp>🤖 Generated by Copilot at 2bca0e9</samp>

This pull request enhances the `decodeFunctionData` function in `src/utils/abi/decodeFunctionData.ts` to support single function ABIs and adds tests for this feature. It also improves the error handling for invalid function signatures.

### DIFF
--- a/src/utils/abi/decodeFunctionData.test.ts
+++ b/src/utils/abi/decodeFunctionData.test.ts
@@ -30,6 +30,18 @@ test('foo()', () => {
       data: '0xc2985578',
     }),
   ).toEqual({ args: undefined, functionName: 'foo' })
+  expect(
+    decodeFunctionData({
+      abi: {
+        inputs: [],
+        name: 'foo',
+        outputs: [],
+        stateMutability: 'nonpayable',
+        type: 'function',
+      } as const,
+      data: '0xc2985578',
+    }),
+  ).toEqual({ args: undefined, functionName: 'foo' })
 })
 
 test('bar(uint256)', () => {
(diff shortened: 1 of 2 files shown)


jxom:
Should we be allowing an ambigious `abi` type? I'm not sure if we should allow inserting a non-standard JSON ABI into `abi`. Think the implementation complexity might not be worth the trade-off than just placing a couple of square brackets beside the item.

fubhy:
> Should we be allowing an ambigious `abi` type? I'm not sure if we should allow inserting a non-standard JSON ABI into `abi`. Think the implementation complexity might not be worth the trade-off than just placing a couple of square brackets beside the item.

Hmm but isn't this a bit of an inconsistency then? https://github.com/wagmi-dev/viem/blob/841c885c1d801f5211c17e0df09aca07a2fb96fe/src/actions/public/getLogs.ts#L24

jxom:
That’s because we use the term “event” (`event`) in `getLogs` rather than “abi” (`abi`).

fubhy:
Ok but then I guess I'm just wondering why we don't expect people to do `getAbiItem` themselves and pass the result to that to a `event` or `function` parameter instead of passing the full abi and a `functionName`. Sorry, I'm just obsessed with DX and these seemingly small details ;-)

fubhy:
I guess I can see the benefit of this when you don't know the right function / event in advance. But then I'd probably want two functions. One where I know the function / event in advance and one where I pass in a range of possibilities. Dunno ...

fubhy:
Another issue I see here is that, if you have a large ABI, every single function call to `decodeFunctionData` loops through the entire ABI first, generates the signature for each item until it finds a match. That also seems a bit wasteful. Especially for cases where you already know the correct match beforehand.

If I was to create e.g. a long running decoding util that's going to process countless tx on a specific contract, I'd like generate a map of signatures & abis first and applying that selectively instead of looping through the whole ABI for every iteration.

jxom:
Could probably do a union between `abi`/`functionName` & `function`.

fubhy:
That would work for me.

fubhy:
Will close this and re-open with a new version as suggested when I get to it.

PR #390, opened 2023-04-19
refactor: replace isomorphic-ws with `unws`

sxzz:
unws (https://github.com/sxzz/unws) is designed for modern browsers and bundlers. 

`isomorphic-ws` is a lack of active maintenance, see https://github.com/heineiuo/isomorphic-ws/issues/33

---

## PR-Codex overview
This PR updates the `ws` package to version 8.12.0 and replaces `isomorphic-ws` with `unws`. It also updates the `abitype` package to version 0.7.1.

### Detailed summary
- Updated `ws` package to version 8.12.0
- Replaced `isomorphic-ws` with `unws`
- Updated `abitype` package to version 0.7.1.


fubhy:
Hm. We are already doing the import dynamically. So why not just do the differentiation ourselve within `viem` and not rely on a package for that at all?

sxzz:
@fubhy Good idea

sxzz:
@fubhy I tried that. But unfortunately, Webpack cannot be aware that `ws` will never be imported on the browser. So it will always bundle `ws` and cause an error.
```
Module not found: Error: Can't resolve 'zlib' in '/path/viem/node_modules/.pnpm/ws@8.12.0/node_modules/ws/lib'
```

fubhy:
> @fubhy I tried that. But unfortunately, Webpack cannot be aware that `ws` will never be imported on the browser. So it will always bundle `ws` and cause an error.
> 
> ```
> Module not found: Error: Can't resolve 'zlib' in '/path/viem/node_modules/.pnpm/ws@8.12.0/node_modules/ws/lib'
> ```

Too bad :-(. Thanks for trying though.

nissoh:
i've tried to riff around this issue, here's a log coming out of this code in Vite typescript environment with allowSyntheticDefaultImports = true

````typescript
  const WebSocket2 = await import('isomorphic-ws')
  const { WebSocket } = await import('isomorphic-ws')

  console.log(window.WebSocket) // ƒ WebSocket() { [native code] }
  console.log(WebSocket2) // Module {Symbol(Symbol.toStringTag): 'Module'}
  console.log(WebSocket) // undefined
````

it's possible that allowSyntheticDefaultImports is not working as intended on Vite, still unclear to me why `isomorphic-ws` cannot be dynamically imported

i was wondering however if ` const WebSocket = await import('isomorphic-ws')` would work on Webpack as-well so both would be compatible? @fubhy

fubhy:
> i've tried to riff around this issue, here's a log coming out of this code in Vite typescript environment with allowSyntheticDefaultImports = true and esModuleInterop = true (just for the kicks)
> 
> ```ts
>   const WebSocket2 = await import('isomorphic-ws')
>   const { WebSocket } = await import('isomorphic-ws')
> 
>   console.log(window.WebSocket) // ƒ WebSocket() { [native code] }
>   console.log(WebSocket2) // Module {Symbol(Symbol.toStringTag): 'Module'}
>   console.log(WebSocket) // undefined
> ```
> 
> i was wondering however if ` const WebSocket = await import('isomorphic-ws')` would work on Webpack as-well so both would be compatible? @fubhy

Yeah that could be related.

Our vitest test suite runs only in node environment atm. which isn't ideal. I've also discussed with @tmm and @jxom to add consumer acceptance tests by adding a few different example packages with different setups, environments, bundlers, typescript version, etc. and seeing if they compile & run properly.

fubhy:
@nissoh Can you maybe test whether it works for you with the replacement package that @sxzz has proposed?

nissoh:
> @nissoh Can you maybe test whether it works for you with the replacement package that @sxzz has proposed?

checked, unws. seems to working properly. package.json includes cjs, esModules and typescript d.ts
https://github.com/sxzz/unws/blob/main/package.json

with that said, i think the core issue is that https://github.com/wagmi-dev/viem/blob/136c4022c3ab4020f37333876bed788a111bff47/src/utils/rpc.ts#L147 doesn't use default export

check https://github.com/heineiuo/isomorphic-ws/blob/master/browser.js
looking at this file, only default is exported. not sure why `{ Webpack }` destrucutre works on webpack

fubhy:
Right. We should probably completely remove the `esModuleInterop` setting from our tsconfig to guard against this.

fubhy:
> > but i wonder if anyone has some insights on Vite
> 
> Vite or Rollup is an ES module bundler and it will strictly treat the modules. It won't convert or interop the code by default. But webpack will.
> 
> All in all, Webpack has more compatibility and Vite respects the specs more (even though some specs are very weird)

Yeah exactly. We should follow the strictest possible setup. 100% agree.

nissoh:
> > > but i wonder if anyone has some insights on Vite
> > 
> > 
> > Vite or Rollup is an ES module bundler and it will strictly treat the modules. It won't convert or interop the code by default. But webpack will.
> > All in all, Webpack has more compatibility and Vite respects the specs more (even though some specs are very weird)
> 
> Yeah exactly. We should follow the strictest possible setup. 100% agree.

are we good with @sxzz solution?

can we merge this?

sxzz:
@fubhy seems that `esModuleInterop` is unnecessary to be disabled. `esModuleInterop` is enabled by default in the default template of Vite projects.

sxzz:
FYI interopDefault is extremely complicated. See the table https://sokra.github.io/interop-test/

fubhy:
> @fubhy seems that `esModuleInterop` is unnecessary to be disabled. `esModuleInterop` is enabled by default in the default template of Vite projects.

It might not be necessary for vite projects. I just noticed that it was enabled and it imho shouldn't be. Libraries shouldn't assume any interop whatsoever imho.

> FYI interopDefault is extremely complicated. See the table https://sokra.github.io/interop-test/

Yeah. I deep dived into that very topic not long ago as I was bumping a few of my own projects to TS5, NodeNext, etc. and revamped the build setup along the way. Complicated indeed.

sxzz:
@fubhy I just checked. There's no import default or export default in viem codebase. So I guess it won't change anything, so we can safely disable this option. Done.

fubhy:
> @fubhy I just checked. There's no import default or export default in viem codebase. So I guess it won't change anything, so we can safely disable this option. Done.

Yeah. It'd basically act as a safe guard to prevent us from introducing sth. like that down the line by accident. I already opened https://github.com/wagmi-dev/viem/pull/391 :)

sxzz:
@jxom Is this PR still needed? I noticed #399 is fixed it.

jxom:
Let's put this PR on hold for now as `isomorphic-ws` is working now (at least I think lol). If anymore issues arise, we can consider switching to this!

nissoh:
@sxzz @jxom  i'm still having issues compiling a standalone project without a bunlder with both esModuleInterop and allowSyntheticDefaultImports turned on

this is reproducible by cloning https://github.com/GMX-Blueberry-Club/Puppet/tree/main/%40gambitdao-gmx-middleware

building by running `pnpm run build` in @gambitdao-gmx-middleware package would throw

`../node_modules/.pnpm/viem@0.3.31_typescript@5.0.4/node_modules/viem/dist/types/utils/rpc.d.ts:2:15 - error TS2595: 'WebSocket' can only be imported by using a default import.`

nissoh:
@jxom @fubhy ping fine sers (;

kolirt:
> @sxzz @jxom i'm still having issues transpiling with `tsc` a standalone project without a bunlder with both `esModuleInterop` and `allowSyntheticDefaultImports` turned on
> 
> this is reproducible by cloning https://github.com/GMX-Blueberry-Club/Puppet
> 
> building by running `pnpm run build` in @gambitdao-gmx-middleware package would throw
> 
> `../node_modules/.pnpm/viem@0.3.31_typescript@5.0.4/node_modules/viem/dist/types/utils/rpc.d.ts:2:15 - error TS2595: 'WebSocket' can only be imported by using a default import.`

same problem. any solutions?

PR #396, opened 2023-04-20
chore: spawn and manage jit anvil instances per worker

fubhy:
Okay, so this is a pretty cool solution imho... Here's how it works:

1. We spawn a http & ws server per worker / thread in `globalSetup`
2. This server is both a proxy and a manager of "JIT" (just-in-time) anvil instances
3. It exposes routing for `localhost:8545/[id]` where `id` is the id of the `vitest` worker
4. Whenever a request hits that endpoint, we spawn or reuse (JIT) an anvil instance and proxy the request to it
5. When the test suite finishes, it cleans up both the proxy and all spawned instances
6. There are no zombie processes

This means that we can parallelize the tests without changing much in the setup except for the rpc urls.

I think we can also extract this into a standalone package (e.g. `anvil-jit`). I think this would be super beneficial for other teams building js tests on top of anvil. Thoughts?

https://user-images.githubusercontent.com/1172528/233375033-6e995bb6-ef1e-4f70-aa4a-d394b5c9e7c2.mp4

---

## PR-Codex overview
This PR adds and modifies several test files, updates a constant file, and removes an Anvil installation step in a GitHub workflow. 

### Detailed summary
- Added `mine` import to `createContractEventFilter.ts`, `setNonce.test.ts`, `getLogs.test.ts`, `getFilterLogs.test.ts`, `getFilterChanges.test.ts`, `watchEvent.test.ts`
- Modified `localHttpUrl` and `localWsUrl` in `_test/constants.ts`
- Removed Anvil installation step in `github/workflows/main.yml`
- Updated `parseAvatarRecord.test.ts`, `getEnsText.test.ts`, `getEnsResolver.test.ts`, `getEnsAddress.test.ts`, `getEnsAvatar.test.ts`, `estimateContractGas.test.ts`, `watchPendingTransactions.test.ts`, `getBalance.test.ts`, `setNonce.test.ts`, `getLogs.test.ts`, `getFilterLogs.test.ts`, `getFilterChanges.test.ts`, `watchEvent.test.ts` to use `mine` function
- Updated `getBalance.test.ts` to use `await mine(testClient, { blocks: 1 })`
- Updated `watchPendingTransactions.test.ts` and `subscribe` function in `watchEvent.test.ts` to remove `retry` option from `waitForExpect`
- Updated `getEnsName.test.ts` and `getEnsResolver.ts` with `vitest` imports, `beforeAll`, and `test` functions
- Updated `parseAvatarRecord.test.ts` and `getEnsAvatar.test.ts` with `vitest` imports, `beforeAll`, `describe`, and `test` functions
- Updated `getFilterChanges.test.ts` and `getFilterLogs.test.ts` with `vitest` imports and `assertType` function

> The following files were skipped due to too many changes: `src/actions/public/watchEvent.test.ts`, `src/utils/promise/withRetry.test.ts`, `src/actions/public/watchContractEvent.test.ts`, `src/actions/public/multicall.test.ts`, `src/utils/buildRequest.test.ts`, `src/actions/public/getTransactionCount.test.ts`, `src/_test/utils.ts`, `src/actions/public/getTransactionReceipt.test.ts`, `src/clients/transports/http.test.ts`, `src/actions/test/inspectTxpool.test.ts`, `package.json`, `src/actions/public/simulateContract.test.ts`, `src/_test/setup.ts`, `src/_test/globalSetup.ts`, `src/_test/proxy.ts`, `src/_test/anvil.ts`, `src/utils/rpc.test.ts`, `src/actions/public/watchBlockNumber.test.ts`, `src/actions/public/watchBlocks.test.ts`, `pnpm-lock.yaml`, `src/actions/wallet/sendTransaction.test.ts`


kristofgazso:
Yes!!! Extracting this into a separate package would be incredibly useful, really annoying that there is no SDK for dealing with Anvil on Typescript

src/_test/globalSetup.ts
```diff
@@ -1,10 +1,25 @@
...
-import { setAutomine, setIntervalMining } from '../actions/index.js'
+import { portNumbers } from 'get-port'
+import { stopAnvilInstances, createAnvilProxy } from './proxy.js'
 
-export async function setup() {
-  if (!process.env.SKIP_GLOBAL_SETUP) {
-    await setBlockNumber(BigInt(parseInt(process.env.VITE_ANVIL_BLOCK_NUMBER!)))
-    await setAutomine(testClient, false)
-    await setIntervalMining(testClient, { interval: 1 })
+export default async function () {
```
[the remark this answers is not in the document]
fubhy: I didn't want to use an environment because I want people to have the option to use `jsdom` or whatever they like (because `globalSetup` shouldn't affect that). I'll just distribute the proxy & anvil command wrapper and then people can configure it how they like. Recommended way (after grinding my teeth on it for quite a while) will be the `globalSetup` approach and using `afterAll` or `afterEach` to reset the instance after each test file / test case.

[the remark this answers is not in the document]
0xOlias: Makes sense. Thanks! Look forward to adopting this.

src/_test/anvil.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export interface AnvilOptions {
+  port: number
+  forkUrl?: string | undefined
+  forkBlockNumber?: number | bigint | undefined
+  blockTime?: number
+  startUpTimeout?: number
+}
+
+export class Anvil {
```
jxom, line 12: what's this `class` keyword?! 

jk

src/_test/anvil.ts
```diff
@@ -0,0 +1,133 @@
+import { execa, type ExecaChildProcess } from 'execa'
+import { Writable } from 'node:stream'
+
+export interface AnvilOptions {
+  port: number
+  forkUrl?: string | undefined
+  forkBlockNumber?: number | bigint | undefined
+  blockTime?: number
+  startUpTimeout?: number
```
jxom, line 9:
```suggestion
  blockTime?: number
  forkUrl?: string | undefined
  forkBlockNumber?: number | bigint | undefined
  port: number
  startUpTimeout?: number
```

src/_test/anvil.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export class Anvil {
+  public readonly options: AnvilOptions
+  public readonly process: ExecaChildProcess
+  private readonly controller: AbortController
+  private readonly recorder: LogRecorder
+
+  public static async start(options: AnvilOptions) {
+    const opts = {
+      startUpTimeout: 10000,
```
jxom, line 20:
```suggestion
      startUpTimeout: 10_000,
```

src/_test/anvil.ts
```diff
@@ -0,0 +1,133 @@
+import { execa, type ExecaChildProcess } from 'execa'
+import { Writable } from 'node:stream'
+
+export interface AnvilOptions {
```
jxom, line 4:
```suggestion
export type AnvilOptions = {
```

src/_test/globalSetup.ts
```diff
@@ -1,10 +1,25 @@
...
+  const server = await createAnvilProxy({
```
jxom, line 5: Might be worth adding a comment here explaining why we are creating a proxy.

src/_test/globalSetup.ts
```diff
@@ -1,10 +1,25 @@
...
+    proxyPort: 8545,
+    proxyHostname: '::',
+    anvilOptions: {
+      portRange: portNumbers(8546, 8546 + 100),
```
jxom, line 9: Perhaps pull the `100` out into a variable? Might be good to implicitly describe what it does.

src/_test/proxy.ts
```diff
@@ -0,0 +1,131 @@
+import { Server, createServer } from 'node:http'
+import { createProxyServer } from 'http-proxy'
+import { Anvil, type AnvilOptions } from './anvil.js'
+import getPort from 'get-port'
+
+export interface AnvilProxyOptions {
```
jxom, line 6:
```suggestion
export type AnvilProxyOptions = {
```

src/_test/proxy.ts
```diff
@@ -0,0 +1,131 @@
...
+  proxyPort?: number
+  proxyHostname?: string
+  anvilOptions: Omit<AnvilOptions, 'port'> & {
+    portRange?: number[] | Iterable<number>
+  }
```
jxom, line 11:
```suggestion
  anvilOptions: Omit<AnvilOptions, 'port'> & {
    portRange?: number[] | Iterable<number>
  }
  proxyHostname?: string
  proxyPort?: number
```

src/_test/setup.ts
```diff
@@ -22,6 +22,48 @@ beforeEach(() => {
...
+  // Reset the anvil instance to the same state it was in before the tests started.
+  await Promise.all([
+    setBlockNumber(BigInt(Number(process.env.VITE_ANVIL_BLOCK_NUMBER))),
+    setAutomine(testClient, false),
+    setIntervalMining(testClient, { interval: 1 }),
+  ])
+})
+
+afterEach((context) => {
+  // Print the last log entries from anvil after eazch test.
```
jxom, line 37:
```suggestion
  // Print the last log entries from anvil after each test.
```

src/actions/public/watchBlockNumber.test.ts
```diff
@@ -365,7 +437,13 @@ describe('subscribe', () => {
...
         onBlockNumber: (blockNumber) => blockNumbers.push(blockNumber),
       })
-      await wait(3000)
+      await wait(200)
+      await mine(testClient, { blocks: 1 })
+      await wait(200)
+      await mine(testClient, { blocks: 1 })
+      await wait(200)
+      await mine(testClient, { blocks: 1 })
+      await wait(200)
```
jxom, line 446: Wonder if we could have a nice helper for this, e.g. 

```ts
await mineInterval({ blocks: 3, interval: 200 })
```

src/_test/anvil.ts
fubhy: Don't you worry. This won't stay long. Will extract that into it's own lib but I found it fitting here 😅

src/_test/anvil.ts
fubhy: Ah, the eternal battle of type vs. interface.

src/actions/public/watchBlockNumber.test.ts
fubhy: Yep. Makes sense. I'm wondering why you guys chose to default to interval mining & not to use auto mining btw.? Without knowing your rationale, I'd think it would be more ergonomics to:

- use auto mining for convenience
- don't use interval mining for test deterministism

Tests where auto mining wouldn't be desirable could then turn it off.

Would probably help to combat some flakiness?

fubhy: Let's do this in a follow-up (`mineInterval` util). But please let me know what you think about my previous comment anyways :-)

fubhy:
Ready for another review @jxom <3

src/_test/setup.ts
```diff
@@ -22,6 +22,48 @@ beforeEach(() => {
   cleanupCache.clear()
 })
 
-afterAll(() => {
-  vi.resetAllMocks()
+afterAll(async () => {
+  vi.restoreAllMocks()
```
fubhy, line 26: @jxom @tmm Btw. sidenote: This was previously incorrectly using `resetAllMocks`. Imho the naming of that is super confusing, but `resetAllMocks` basically only clears the mock history. It doesn't fully "undo" them. Took me a while to debug the errors caused by this because it's also easy to read past it even if you know the difference between the two because they read so similarly).

https://vitest.dev/api/vi.html#vi-resetallmocks
https://vitest.dev/api/vi.html#vi-restoreallmocks

Also important to remember this when undoing vi.spy / vi.mock in individual tests.

src/utils/rpc.test.ts
```diff
@@ -204,53 +205,54 @@ describe('http', () => {
...
-        Details: foo
-        Version: viem@1.0.2]
-      `,
-      )
-    }
+      Details: foo
+      Version: viem@1.0.2"
+    `)
+
+    mock.mockRestore()
```
fubhy, line 254: @jxom @tmm It's important to undo mocks / spies so they don't leak into other tests. I've not done that universally because at the end of each test file, we do this in `setup.ts` generally. But sometimes this will cause unexpected behavior if subsequent tests in the same file rely on an unmocked version of "the thing".

I've seen a few cases where we don't reset these mocks at the end of a test. Hence, why I'm bringing this up.

src/actions/public/watchBlockNumber.test.ts
jxom: Chose to do interval mining as it more closely represents real-world timing conditions. But agree it’s not suitable for all tests, however, I think it’s a reasonable default.

fubhy: Ok, fair enough. It just means we have to be concious with tests that can be affected by this and remember that this might be the cause for flaky tests whenever we see them (e.g. expecting certain logs at the "latest" block can fail if a block is mined inbetween).

tmm (approved):

jxom (approved):
An insane improvement! 🔥

PR #424, opened 2023-04-27
refactor: remove defineChain export

fubhy:
## PR-Codex overview
This PR removes the `defineChain` export from `viem/chains` and removes `defineChain` import from multiple files.

### Detailed summary
- Removed the `defineChain` export from `viem/chains`.
- Removed `defineChain` import from `src/chains.ts`.
- Removed `defineChain` import from `src/actions/wallet/sendTransaction.test.ts`.

### DIFF
--- a/.changeset/gold-bikes-reply.md
+++ b/.changeset/gold-bikes-reply.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Removed the `defineChain` export from `viem/chains`.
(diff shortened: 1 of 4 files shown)


jxom:
This is technically a breaking change. Don't think anyone would be using it yet though. Maybe we can move the export to the root entrypoint?

fubhy:
> This is technically a breaking change. Don't think anyone would be using it yet though. Maybe we can move the export to the root entrypoint?

Yeah. I think it's better to do this now than later. Also, it's a very small breaking change considering that all you'd have to do to update is to change the import. I'll add it to `index.ts` too.

fubhy:
Yeah, so imho this would be ready then? Unless I'm missing sth. :-)

PR #466, opened 2023-05-09
test: add failing test for formatting plain Solidity errors in error messages

holic:
Reverting with a plain custom error in Solidity (e.g. `revert PlainError()`) doesn't surface in error messages.

## PR-Codex overview
This PR adds new error types to the `ErrorsExample` contract and includes a test for a custom plain error function. 

### Detailed summary:
- Added `PlainError()`, `SimpleError(string message)`, and `ComplexError(Foo foo, string message, uint256 number)` error types to `ErrorsExample` contract
- Included a test for a custom plain error function in `src/errors/contract.test.ts`

### DIFF
--- a/contracts/src/ErrorsExample.sol
+++ b/contracts/src/ErrorsExample.sol
@@ -7,6 +7,7 @@ contract ErrorsExample {
         uint256 bar;
     }
 
+    error PlainError();
     error SimpleError(string message);
     error ComplexError(Foo foo, string message, uint256 number);
 
--- a/src/errors/contract.test.ts
+++ b/src/errors/contract.test.ts
@@ -378,6 +378,20 @@ describe('ContractFunctionRevertedError', () => {
 
       Version: viem@1.0.2]
     `)
+
+    expect(
+      new ContractFunctionRevertedError({
+        abi: errorsExampleABI,
+        data: '0x253eb71d',
+        functionName: 'customPlainError',
+      }),
+    ).toMatchInlineSnapshot(`
+      [ContractFunctionRevertedError: The contract function "customPlainError" reverted.
+
+      Error: PlainError()
+
+      Version: viem@1.0.2]
+    `)
   })
 
   test('data: zero data', () => {


holic:
I went to go fix this but there's a few design decisions that I don't feel empowered to make.

**Update `ContractFunctionRevertedError` to handle `undefined` `args` for plain custom errors**

There's an expectation in `ContractFunctionRevertedError` that [`args` will be filled in](https://github.com/wagmi-dev/viem/blob/41c78b3ac1687fcc92c33349e0d8e6e6bffe5103/src/errors/contract.ts#L175) when it's a custom error. We could instead check for `errorName` or `abiItem`, but types tell me that those are guaranteed to exist, so then we'd never reach [the final condition to show the original message](https://github.com/wagmi-dev/viem/blob/41c78b3ac1687fcc92c33349e0d8e6e6bffe5103/src/errors/contract.ts#L197).

**Update `decodeErrorResult` to return an empty `args` array for plain custom errors**

This makes the most sense to me, but [the `decodeErrorResult` tests tell me that this is intentional](https://github.com/wagmi-dev/viem/blob/41c78b3ac1687fcc92c33349e0d8e6e6bffe5103/src/utils/abi/decodeErrorResult.test.ts#L24) and I'm not sure the downstream effects of using an empty array here. I'm also not sure the difference between an empty `args` array only when ABI `inputs` is `[]` vs no ABI `inputs` defined.

Neither option feels great, so I'm looking for advice!

Separately, it feels scary to me to have a function call (`decodeErrorResult`) that can throw inside the constructor of an error object (which is called when throwing). I was considering an approach where `decodeErrorResult` would return nothing if no matching ABI exists, then check if there was an error ABI match in `ContractFunctionRevertedError`, but unsure how ya'll feel about that approach. It may still be prone to the original `message` issue mentioned above.

holic:
looks like this got added in https://github.com/wagmi-dev/viem/commit/9852bcdfb83bfdda43fb276b3a1266b30592d56e

PR #478, opened 2023-05-11
feat: change return value of getByteCode

HappySean2845:
Dicussion here: [https://github.com/wagmi-dev/viem/discussions/477](https://github.com/wagmi-dev/viem/discussions/477)

---

## PR-Codex overview
This PR removes the possibility of `undefined` being returned by `getBytecode` function in `getBytecode.ts` and now it always returns a hex string.

### Detailed summary
- `GetBytecodeReturnType` type is modified to only allow `Hex` type
- `getBytecode` function is modified to remove the possibility of returning `undefined`
- `getBytecode` function now always returns a hex string

### DIFF
--- a/.changeset/empty-needles-cheer.md
+++ b/.changeset/empty-needles-cheer.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
+Removed undefined from return value of getByteCode. Now it always return a hex string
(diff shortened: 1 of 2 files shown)


jxom:
Totally understand where you are coming from. viem aligns zero data ("0x") with `undefined` in other places in code, so if we do this here, then we will need to update other places. I also think that aligning zero data with `undefined` is better anyway as it generally leads to a better DX, ie. `if (!data)` instead of `if (data === '0x')`. Might be best to leave it as is for now, and if it becomes a hassle for more people in the future, we can reconsider.

HappySean2845:
Sounds resonable, thanks for reply

PR #487, opened 2023-05-12
fix: encodeEventTopics for leading non-indexed args

### LINKED ISSUE #485: bug: `encodeEventTopics` doesn't encode args correctly for some events

### Package Version

latest

### Current Behavior

`encodeEventTopics` that's used among others by `getLogs` doesn't properly encode arguments into topics for some events. Take a look at this code snippet:

https://github.com/wagmi-dev/viem/blob/1c539f82dd49d811ee771a030f2597eb1da86225/src/utils/abi/encodeEventTopics.ts#L62-L79

Let's assume `args` is an object, not an array. Firstly, `args_` is created by mapping over all `abiItem.inputs` - **this will include both indexed and not indexed args.** If we imagine an event that has 6 arguments, and only last of them is indexed and included in `args`, this can produce an example `const args_ = [undefined, undefined, undefined, undefined, undefined, 'value']`.

Next, `topics` is populated by mapping `abiItem.inputs`, **but this time only `indexed` ones.**. Then array index is used to access `args_`, but because `args_` is created from all the inputs, there are index mismatches. And instead using `'value'` as argument, `undefined` is accessed.

In conclusion, this function will work only if all the indexed event arguments are placed before any non-indexed arguments in event signature, which doesn't have to be the case. Hopefully my explanation makes sense, let me know if there's any confusion

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

kdembler:
Fixes #485

For more details, check out the issue.

In first commit, I've added a test for an event that has 2 arguments, and only second one is indexed. The test was failing. Then, small change was made to `encodeEventTopics` and now the test is passing.

---

## PR-Codex overview
This PR fixes the `encodeEventTopics` function to properly handle leading non-indexed arguments in event topics. 

### Detailed summary
- Added a new test case for `Bar(string,string)` with leading non-indexed arguments
- Modified the `encodeEventTopics` function to handle leading non-indexed arguments properly by filtering out non-indexed parameters and mapping indexed ones to their corresponding values in `args`

### DIFF
--- a/src/utils/abi/encodeEventTopics.ts
+++ b/src/utils/abi/encodeEventTopics.ts
@@ -61,21 +61,22 @@ export function encodeEventTopics<
 
   let topics: Hex[] = []
   if (args && 'inputs' in abiItem) {
+    const indexedAbiItemInputs = abiItem.inputs?.filter(
+      (param) => 'indexed' in param && param.indexed,
+    )
     const args_ = Array.isArray(args)
       ? args
-      : abiItem.inputs?.map((x: any) => (args as any)[x.name]) ?? []
+      : indexedAbiItemInputs?.map((x: any) => (args as any)[x.name]) ?? []
     topics =
-      abiItem.inputs
-        ?.filter((param) => 'indexed' in param && param.indexed)
-        .map((param, i) =>
-          Array.isArray(args_[i])
-            ? args_[i].map((_: any, j: number) =>
-                encodeArg({ param, value: args_[i][j] }),
-              )
-            : args_[i]
-            ? encodeArg({ param, value: args_[i] })
-            : null,
-        ) ?? []
+      indexedAbiItemInputs?.map((param, i) =>
+        Array.isArray(args_[i])
+          ? args_[i].map((_: any, j: number) =>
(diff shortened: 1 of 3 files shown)


src/utils/abi/encodeEventTopics.ts
```diff
@@ -61,21 +61,22 @@ export function encodeEventTopics<
 
   let topics: Hex[] = []
   if (args && 'inputs' in abiItem) {
+    const indexedAbiItemInputs = abiItem.inputs?.filter(
+      (param) => 'indexed' in param && param.indexed,
+    )
     const args_ = Array.isArray(args)
       ? args
-      : abiItem.inputs?.map((x: any) => (args as any)[x.name]) ?? []
+      : indexedAbiItemInputs?.map((x: any) => (args as any)[x.name]) ?? []
```
kdembler, line 69: This only works assuming that if `args` is already an array, it will contain only indexed arguments, otherwise there is still an error case. Is that guaranteed?

jxom:
ty!

Issue #493, opened 2023-05-13
bug: custom error not thrown on failed simulation


ubinatus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

^0.3.22

### Current Behavior

Hello Team,

Something is going on with the custom solidity errors on the contract. When simulating the contract transaction (`publicClient.simulateContract`) that is expected to be reverted with a custom error, the thrown error doesn't contain data nor the expected error.

Right now, whenever a contract write call gets reverted with a custom solidity error, it simply throws an error without an indication of what's the specific error. There is no `data` as well. The latter could be used to decode the custom error with `decodeErrorResult`

### Expected Behavior

The expected (or desired) behaviour is to receive the custom solidity error or the error data (in order to decode it with `decodeErrorResult`) in case the simulated transaction actually get reverted with a custom solidity error.

### Steps To Reproduce

1. Open the attached stackblitz example
2. Open your developer console and see the logs of the simulation error -> There will be no specific indication of the custom solidity error

### Link to Minimal Reproducible Example (StackBlitz, CodeSandbox, GitHub repo etc.)

https://stackblitz.com/edit/viem-getting-started-kizhyl?file=index.ts

#### Current contract method
![image](https://github.com/wagmi-dev/viem/assets/51177379/2f9b9067-ef66-444a-abaf-0ac86d9e11bc)

#### Current error thrown on simulation
![image](https://github.com/wagmi-dev/viem/assets/51177379/5b10e7b7-9e45-4f2e-b165-05c98fc0f8fc)

#### Contract address
https://polygonscan.com/address/0xAB56176Ab5Bb06A6aE90e9E3EDEEDD683c4B09B6#code#F19#L213

jxom:
Fixed!

ubinatus:
That was fast! Thanks a lot @jxom ! So the custom error for now is found on the `err.metaMessages[0]`. Good

jxom:
Yeah, good point, might be good to expose it as an attribute off the error instance, instead of extracting from the meta messages strings. Can look into that!

ubinatus:
Sound good! I'm about to sleep but tomorrow morning that I have some time could help you with a PR for that issue. Will check in tomorrow in case it's not resolved  ☺️

jxom:
Feel free! It will be quite trivial to add.

emretepedev:
[...]

PR #497, opened 2023-05-14
docs: add "Handle custom errors" section in `simulateContract`

ubinatus:
The intention of this PR is to expose the custom solidity error revertion on the Error instance when calling `readContracts`, which was motivated by https://github.com/wagmi-dev/viem/issues/493

Previously, in order to know if the simulation (or read) reverts due to a custom solidity error we had to lookup the `metaMessages` array attribute to see any coincidence:

```typescript
try {
  await publicClient.simulateContract(ops);
} catch (err) {
  // Checking if cause is as custom solidity error
  if (Array.isArray(err.metaMessages) && err.metaMessages.length > 0) {
    const possibleError = err.metaMessages[0] as string;
    const foundError = [
      "SimpleError(string message)",
      "SimpleErrorNoArgs()",
      "ComplexError((address sender, uint256 bar), string message, uint256 number)",
    ].find((x) => possibleError.includes(x));
    if (foundError) {
      // Simulation reverted because of a custom solidity error
      // >>> Show a user friendly message to the user or do something
    }
  }
}
```

With the current PR, the custom error is exposed with its formatted arguments called (in case it exists). The attribe has the following type:

```typescript
// Type
errorName?: string
errorArgs?: unknown[]
errorAbiItem?: AbiError
// Example 1: SimpleError
      err.errorName: 'SimpleError',
      err.errorArgs: ['bugger'],
      err.errorAbiItem: {
        inputs: [
          {
            internalType: 'string',
            name: 'message',
            type: 'string',
          },
        ],
        name: 'SimpleError',
        type: 'error',
      },
// Example 2: SimpleErrorNoArgs
      err.errorName: 'SimpleErrorNoArgs',
      err.errorArgs: undefined,
      err.errorAbiItem: {
        inputs: [],
        name: 'SimpleErrorNoArgs',
        type: 'error',
      },


["SimpleErrorNoArgs()", undefined]
// Example 3: ComplexError
       err.errorName: 'ComplexError',
       err.errorArgs: [
        {
          bar: BigInt(69),
          sender: '0x0000000000000000000000000000000000000000',
        },
        'bugger',
        BigInt(69),
      ],
      err.errorAbiItem: {
        inputs: [
          {
            components: [
              {
                internalType: 'address',
                name: 'sender',
                type: 'address',
              },
              {
                internalType: 'uint256',
                name: 'bar',
                type: 'uint256',
              },
            ],
            internalType: 'struct ErrorsExample.Foo',
            name: 'foo',
            type: 'tuple',
          },
          {
            internalType: 'string',
            name: 'message',
            type: 'string',
          },
          {
            internalType: 'uint256',
            name: 'number',
            type: 'uint256',
          },
        ],
        name: 'ComplexError',
        type: 'error',
      },
```

So, taking the previous example:

```typescript
try {
  await publicClient.simulateContract(ops);
} catch (err) {
  // Checking if cause is as custom solidity error
  if (err.errorName && ["Panic", "Error"].includes(err.errorName)) {
    // Simulation reverted because of a custom solidity error
    // >>> Throw a user friendly message to the user
    //
  }
}
```

---

## PR-Codex overview
This PR adds an example of how to handle custom errors in `simulateContract` function. 

### Detailed summary
- Adds example code for handling custom errors in `simulateContract` function.
- Shows how to catch a custom error thrown by `simulateContract` and access it through the `data` attribute of the error.
- Includes the custom error item in the contract `abi`.
- Provides two options for checking the instance of the error.

### DIFF
--- a/src/errors/contract.ts
+++ b/src/errors/contract.ts
@@ -1,4 +1,4 @@
-import type { Abi, AbiError, Address } from 'abitype'
+import type { Abi, Address } from 'abitype'
 
 import { parseAccount } from '../accounts/utils/parseAccount.js'
 import type { CallParameters } from '../actions/public/call.js'
@@ -81,9 +81,6 @@ export class ContractFunctionExecutionError extends BaseError {
   formattedArgs?: string
   functionName: string
   sender?: Address
-  errorAbiItem?: AbiError
-  errorName?: string
-  errorArgs?: unknown[]
 
   override name = 'ContractFunctionExecutionError'
 
@@ -130,19 +127,6 @@ export class ContractFunctionExecutionError extends BaseError {
       sender,
     })
 
-    let errorAbiItem
-    let errorName
-    let errorArgs
-    if (
-      cause instanceof ContractFunctionRevertedError &&
-      cause.data &&
-      !['Panic', 'Error'].includes(cause.data.errorName)
-    ) {
-      errorAbiItem = cause.data.abiItem
-      errorName = cause.data.errorName
-      errorArgs = cause.data.args as unknown[]
-    }
-
     super(
       cause.shortMessage ||
         `An unknown error occurred while executing the contract function "${functionName}".`,
@@ -162,9 +146,6 @@ export class ContractFunctionExecutionError extends BaseError {
     this.contractAddress = contractAddress
     this.functionName = functionName
     this.sender = sender
-    this.errorAbiItem = errorAbiItem
-    this.errorName = errorName
-    this.errorArgs = errorArgs
   }
 }
 
--- a/src/errors/contract.ts
+++ b/src/errors/contract.ts
@@ -1,4 +1,4 @@
-import type { Abi, Address } from 'abitype'
+import type { Abi, AbiError, Address } from 'abitype'
 
 import { parseAccount } from '../accounts/utils/parseAccount.js'
 import type { CallParameters } from '../actions/public/call.js'
@@ -73,8 +73,6 @@ export class CallExecutionError extends BaseError {
   }
 }
 
-type CustomErrorData = [string, string | undefined]
-
 export class ContractFunctionExecutionError extends BaseError {
   abi: Abi
   args?: unknown[]
@@ -83,7 +81,9 @@ export class ContractFunctionExecutionError extends BaseError {
   formattedArgs?: string
   functionName: string
   sender?: Address
-  customError?: CustomErrorData
+  errorAbiItem?: AbiError
+  errorName?: string
+  errorArgs?: unknown[]
 
   override name = 'ContractFunctionExecutionError'
 
@@ -130,10 +130,14 @@ export class ContractFunctionExecutionError extends BaseError {
       sender,
     })
 
-    const customError: CustomErrorData | undefined =
-      cause instanceof ContractFunctionRevertedError
-        ? cause.customError
-        : undefined
+    let errorAbiItem
+    let errorName
+    let errorArgs
+    if (cause instanceof ContractFunctionRevertedError && cause.data) {
+      errorAbiItem = cause.data.abiItem
+      errorName = cause.data.errorName
+      errorArgs = cause.data.args as unknown[]
+    }
 
     super(
       cause.shortMessage ||
@@ -154,7 +158,9 @@ export class ContractFunctionExecutionError extends BaseError {
     this.contractAddress = contractAddress
     this.functionName = functionName
     this.sender = sender
-    this.customError = customError
+    this.errorAbiItem = errorAbiItem
+    this.errorName = errorName
+    this.errorArgs = errorArgs
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/errors/contract.ts
```diff
@@ -186,6 +197,10 @@ export class ContractFunctionRevertedError extends BaseError {
               })
             : undefined
 
+        if (errorWithParams !== undefined) {
+          customError = [errorWithParams, formattedArgs]
```
jxom, line 201: Instead of `customError`, what if we add `abiItem`, `errorName` & `errorArgs` to the error instead?

This means that people will more deterministically be able to extract errors.

ubinatus: 100% agree with you. It's more helpful than just display the custom error. I've just pushed it. Will be changing the PR description to match this

ubinatus: Done! Also, ended up optimizing the `readContract` tests from my initial commit.

ubinatus: Currently, the `abiItem`, `errorName` and `errorArgs` are added not only on custom errors but, as well, on revert/assert/require errors. For those, the `errorName` is either "Panic" or "Error". Should I only expose the mentioned attributes on custom solidity errors?

ubinatus: @jxom Thinking that those attributes should be added to the error instance only if it's a custom solidity error for the following reasons:

1. For "Panic" exceptions (e.g. assert, overflows and divisions by zero): there is no need to extract the error since the `reason` already explains the panic message by suing the [panic via assert and error via require](https://docs.soliditylang.org/en/v0.8.16/control-structures.html#panic-via-assert-and-error-via-require) (which is handled on file `src/constants/solidity.ts`
2. For "Error" exceptions (i.e. doing `require(cond, message)`): the require message already explains the cause of the error.

Taking 1 and 2, there is no need to show the `errorAbiItem`, `errorName` and `errorArgs` for non custom solidity errors, as it will causes the developer to have to filter the `errorName` of type "Panic" and "Error" to find that the error is a custom error.

Let me know your thoughts :)

ubinatus:
Attributes `errorAbiItem`, `errorName` and `errorArgs` are now only exposed only if the contract transaction reverts with a custom solidity error.

jxom:
I just realised, could you not extract these off `error.cause` already? Meaning this PR may potentially not be needed?

ubinatus:
> I just realised, could you not extract these off `error.cause` already? Meaning this PR may potentially not be needed?

You're right! It can be extracted from the error.cause, as well as the `details`, `docsPath`, `metaMessages`, `shortMessage`, `version` and `reason`. 

However, it adds more convenience to the developer to access the custom error:
#### Main
```typescript
try {
  await call()
} catch (err) {
  if (
    err instanceof ContractFunctionExecutionError &&
    err.cause instanceof ContractFunctionRevertedError
  ) {
    // Check if custom error
    const customErrorList = ['CustomErrorA', 'CustomErrorB', 'CustomErrorC']
    if (customErrorList.includes(err.cause.data?.errorName ?? '')) {
      // Handle the custom error exposure to the user
      // >>
    }
  }
}
```
#### Current PR
```typescript
try {
  await call()
} catch (err) {
  if (err instanceof ContractFunctionExecutionError) {
    // Check if custom error
    const customErrorList = ['CustomErrorA', 'CustomErrorB', 'CustomErrorC']
    if (customErrorList.includes(err.errorName ?? '')) {
      // Handle the custom error exposure to the user
      // >>
    }
  }
}
```

Let me know your thoughts. In case this PR is not needed, feel free to close the PR. 😊 The `viem` user is still capable to extract the custom error through the `cause` attribute anyways.

jxom:
I think it might be better to leave it more verbose as it:

1. Errors were designed to be like this via `cause`, so folks can extract information out of each of them,
2. Makes things less ambiguious (ie. consumers may expect `errorName` to be populated in Error and Panic),
3. Keeps the code lean and pure (free of side-effects)

What do you think?

Plus, you can utilize `error.walk` to make it a bit more clearer:

```ts
try { ... } catch (err) {
  if (err.walk(err => err instanceof ContractFunctionRevertedError)) {
    err.data.errorName
  }
}
```

Though, it could be nice to have this in the documentation. Would you be happy to add some notes there if you would still like the contribution?

ubinatus:
@jxom Done! I added a section in the `Contracts < simulateContract` documentation. Please let me know if I should move it to another section. 

I think it's more relevant to place it there rather than under the readContract section. This is mainly because most custom errors are present in write methods due to their nature of being a gas-efficient way to revert errors. Additionally, one of the purposes of catching the custom error while building dApps is to rephrase it in a user-friendly way for better understanding of why the transaction failed to submit.

jxom:
Thank you!

jxom:
Ah good catch. You’re assumption is correct… open to a PR, go for it.

Alexsey:
So, the bug is in the docs or in the code? :)

jxom:
The code!

ubinatus:
Yes, the code. Walk function should return an instance of ContractFunctionRevertedError

Alexsey:
And what should it return when called without callback? At the moment, it will return the most bottom error. Should it work this way? Or should we make callback mandatory?

PR #500, opened 2023-05-14
refactor: lax log types

jxom:
Most of the time, the length of the data/topics passed to `decodeEventLog` will be non-deterministic. Even if we do know the event (and thus, the expected length of topics/data), the topics/data length is still not guarenteed as the event 4-byte selector is not aware of indexed/non-indexed args. Because of this, there is a chance that there could be a mismatch between indexed/non-indexed arguments with their corresponding topics/data, so it's not guarenteed that we can fully decode every argument. 

This PR aims to change the attributes within the `args` type to be optional to represent this, and as a result, we can just use the `AbiEventParametersToPrimitiveTypes` type instead.

---

## PR-Codex overview
This PR updates the `args` object in several files to make some properties optional. It also adds a new `AbiEventParametersToPrimitiveTypes` type to convert ABI event parameters to primitive types.

### Detailed summary
- Make some properties optional in `args` object in several files
- Add `AbiEventParametersToPrimitiveTypes` type to convert ABI event parameters to primitive types

> The following files were skipped due to too many changes: `src/types/contract.ts`

### DIFF
--- a/src/types/log.ts
+++ b/src/types/log.ts
@@ -93,7 +93,7 @@ type GetInferredLogValues<
         args: GetEventArgs<
           TAbi,
           TEventName,
-          { EnableUnion: false; IndexedOnly: false; Required: true }
+          { EnableUnion: false; IndexedOnly: false; Required: false }
         >
         /** The event name decoded from `topics`. */
         eventName: TEventName
@@ -105,7 +105,7 @@ type GetInferredLogValues<
           args: GetEventArgs<
             TAbi,
             string,
-            { EnableUnion: false; IndexedOnly: false; Required: true }
+            { EnableUnion: false; IndexedOnly: false; Required: false }
           >
           /** The event name decoded from `topics`. */
           eventName: TName
--- a/src/types/log.ts
+++ b/src/types/log.ts
@@ -93,7 +93,7 @@ type GetInferredLogValues<
         args: GetEventArgs<
           TAbi,
           TEventName,
-          { EnableUnion: false; IndexedOnly: false; Required: false }
+          { EnableUnion: false; IndexedOnly: false }
         >
         /** The event name decoded from `topics`. */
         eventName: TEventName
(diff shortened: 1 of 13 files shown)


src/utils/abi/decodeEventLog.test-d.ts
```diff
@@ -100,7 +100,11 @@ test('unnamed', async () => {
   })
 
   expectTypeOf(event).toEqualTypeOf<{
-    args: [Address, Address, bigint]
+    args:
+      | readonly []
+      | readonly [`0x${string}`]
+      | readonly [`0x${string}`, `0x${string}`]
+      | readonly [`0x${string}`, `0x${string}`, bigint]
```
tmm, line 107: 😍

src/types/log.ts
```diff
@@ -105,7 +105,7 @@ type GetInferredLogValues<
           args: GetEventArgs<
             TAbi,
             string,
-            { EnableUnion: false; IndexedOnly: false; Required: true }
+            { EnableUnion: false; IndexedOnly: false; Required: false }
```
tmm, line 108: Looks like `Required` is set to `false` everywhere now (except `contract.test-d.ts`). Should we can it?

tmm: Also looks like `EnableUnion` and `IndexedOnly` are always `false` too. Maybe should drop these options all together?

tmm (approved):
Couple thoughts that might allow us to remove a bit more code, but overall solid.

src/types/log.ts
jxom: Yep, that would be fantastic.

nissoh:
@jxom not sure if related to this PR, i i bumped Viem to 0.3.31 and event args became optional, accessing args will require extra check, not sure if intended

before
image

after
image

jxom:
It was intended. This is because the event signature does not honor the existence of indexed/non-indexed parameters.

For example, the two definitions share the same signature:

`Transfer(address indexed from, address indexed to, uint256 value)`

`Transfer(address from, address to, uint256 value`

So it makes it hard for viem to figure out what arguments in topics/data to decode. So if one contract doesn’t respect the intended definition, it can’t decode all args.

Maybe we could make this configurable if the consumer knows that the contract event definition is deterministic.

PR #505, opened 2023-05-15
feat: support loose abi decode

izayl:
## PR-Codex overview
This PR adds a new optional parameter to `decodeAbiParameters` that allows ignoring extra data at the end of the encoded data. It also updates the documentation to reflect this change.

### Detailed summary
- Added a new optional `loose` parameter to `decodeAbiParameters`
- Updated `decodeFunctionData` to include a note about truncating data if it's not aligned to 32 bytes
- Updated documentation for `decodeFunctionData` and `decodeAbiParameters` to reflect the new `loose` parameter

### DIFF
--- a/.changeset/witty-pianos-lick.md
+++ b/.changeset/witty-pianos-lick.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added loose abi decode to decodeAbiParameters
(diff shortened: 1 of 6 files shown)


izayl:
there is the real case who use postfix to data

https://github.com/reservoirprotocol/indexer/blob/main/packages/sdk/src/seaport-base/exchange.ts#L96

all the nft marketplace protocol fillOrder data added this postfix:

https://github.com/search?q=repo%3Areservoirprotocol%2Findexer%20%2B%20generateSourceBytes&type=code

jxom:
Thanks for the PR! I think we can get away with this without needing a `loose` attribute. Opened another PR at #506 and added you as a co-author of the commit.

PR #540, opened 2023-05-19
feat: proxy packages

jxom:
Some bundlers (Parcel, Metro, Webpack 4, etc) do not support `package.json#exports`. Would be nice to be compatible with them if folks do not want to migrate to a different bundler, or migration effort is too much.

---

## PR-Codex overview
This PR adds proxy packages to support bundlers that do not work with `package.json#exports`.

### Detailed summary
- Added proxy packages for each export
- Generates a package.json with only necessary fields
- Replaces `files` with `files_` to include proxy packages
- Supports bundlers not compatible with `package.json#exports`

### DIFF
--- a/scripts/prepublishOnly.ts
+++ b/scripts/prepublishOnly.ts
@@ -1,5 +1,9 @@
-import { readJsonSync, writeJsonSync } from 'fs-extra'
+import { outputFileSync, readJsonSync, writeJsonSync } from 'fs-extra'
 import path from 'path'
+
+type Exports = {
+  [key: string]: string | { types?: string; import: string; default: string }
+}
 
 generatePackageJson()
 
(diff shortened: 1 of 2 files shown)


scripts/prepublishOnly.ts
```diff
@@ -31,6 +35,34 @@ function generatePackageJson() {
...
+      `${key}/package.json`,
+      `{
+  ${Object.entries(value)
+    .map(([k, v]) => {
+      const key = (() => {
+        if (k === 'import') return 'module'
+        if (k === 'default') return 'main'
+        if (k === 'types') return 'types'
+        throw new Error('Invalid key')
+      })()
```
jxom, line 57: one can only dream 
https://github.com/tc39/proposal-pattern-matching

tmm (approved):

Issue #499, opened 2023-05-14
bug: transactionIndex from getFilterLog is statically typed as number but receives BigInt at runtime


deodad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.3.24

### Current Behavior

I am querying logs using `createContractEventFilter` and `getFilterLogs`. The `Log` type has a value of `number` for the `TIndex` generic, but at runtime I'm seeing a `bigint`.

### Expected Behavior

`getFilterLogs` return type has `TIndex` of `bigint` for the `Log` type.

Volchunovich:
Can I take it ?

jxom:
Go for it.

Volchunovich:
@jxom could you help me? I want to try to push my changes into my branch but got failed with denied permissions. What should I do to push and create PR?

jxom:
Fixed on #576

PR #552, opened 2023-05-22
Fix logIndex type in formatLog 

Volchunovich:
[Issue](https://github.com/wagmi-dev/viem/issues/499)

---

## PR-Codex overview
This PR focuses on fixing the return type of `getFilterLogs` function in `log.ts`. 

### Detailed summary
- Changed the return type of `logIndex` and `transactionIndex` from `BigInt` to `Number` in `getFilterLogs` function.


Volchunovich:
@tmm could you help me to understand why tests failed if I even don't touch these files?

jxom:
Instead, can we please change https://github.com/wagmi-dev/viem/blob/main/src/utils/formatters/log.ts#L12 to `Number()` instead of `BigInt()`. This was an oversight from us, it should be a number instead.

Volchunovich:
@jxom of course, I'll return the type and just change formatLog then

PR #556, opened 2023-05-22
fix: fallback to client.account for estimateGas

tmm:


jxom (approved):

jxom:
Could be worth adding a test case too!

tmm:
Will do!

PR #578, opened 2023-05-25
Fix: exception handling when rejecting a request sent to Metamask through walletConnect

siosio34:
This issue is a PR that fixes an issue where UserRejectedRequestError is not thrown when the user rejects the transaction when calling the sendTransaction function in metamask using walletconnect-legacy-provider.

---

## PR-Codex overview
This PR adds a new error type for when a transaction is rejected through WalletConnect on the Metamask mobile app.

### Detailed summary
- Added `UserRejectedRequestError` import
- Added check for `ProviderRpcError` with code `4001`
- Throws `UserRejectedRequestError` if above condition is met

### DIFF
--- a/src/actions/wallet/sendTransaction.ts
+++ b/src/actions/wallet/sendTransaction.ts
@@ -4,8 +4,10 @@ import type { WalletClient } from '../../clients/createWalletClient.js'
 import type { Transport } from '../../clients/transports/createTransport.js'
 import { AccountNotFoundError } from '../../errors/account.js'
 import type { BaseError } from '../../errors/base.js'
+import { UserRejectedRequestError } from '../../index.js'
 import type { GetAccountParameter } from '../../types/account.js'
 import type { Chain, GetChain } from '../../types/chain.js'
+import { ProviderRpcError } from '../../types/eip1193.js'
 import type { Formatter } from '../../types/formatter.js'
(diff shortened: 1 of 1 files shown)


jxom:
Hmm, this should be caught here: https://github.com/wagmi-dev/viem/blob/main/src/utils/buildRequest.ts#L87

siosio34:
@jxom 

I also thought so, and when connecting with the Metamask extension in the desktop web browser, I confirmed that an error was caught there as you say.

However, when using wallet-connect-legacy-provider, it seems that the error cannot be caught in that place, so I wrote this pull request.

If you know of any other way to make it work as intended using WalletConnect, please let me know!

jxom:
Does this happen with WalletConnect v2 as well? Not sure if it's worth fixing this if WalletConnect v1 is being shut down next month.

siosio34:
Currently, metamask does not support WALLECT CONNECT V2, so we cannot test it.
As far as I know, Metamask supports WALLETCONNECT V2 on June 15th.
At that time, I will connect WALLETCONNECT V2 and test it.
If the same issue does not occur, I will close this pull request.

So have a nice day ~

jxom:
Sounds great! Thank you for looking into this.

I am going to close this for now, but feel free to copy + paste this issue, and open another one if the issue still occurs.

PR #588, opened 2023-05-26
doc: add tuple notice

izayl:
## PR-Codex overview
This PR provides information about differences between viem and ethers, specifically regarding tuple expressions and the lack of support for fragments and interfaces in viem.

### Detailed summary:
- Viem only supports standard tuple expressions for Human Readable
- Fragments and interfaces are not supported in viem
- Encourages working with the ABI itself instead of using "fragments"

### DIFF
--- a/site/docs/ethers-migration.md
+++ b/site/docs/ethers-migration.md
@@ -800,6 +800,9 @@ decodeAbiParameters(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


izayl:
If you think the tuple keyword should be support, I'm willing to create a new PR

tmm:
> If you think the tuple keyword should be support, I'm willing to create a new PR

Viem supports first-class structs in human-readable ABIs. More info here: https://abitype.dev/api/human.html#structs

The tuple keyword isn't valid Solidity so trying to avoid and stay as close to Solidity as possible.

PR #595, opened 2023-05-29
feat: compatibility with new universal resolver

jxom:
The new `UniversalResolver` now reverts with `"Wildcard on non-extended resolvers is not supported"` if it cannot resolve an ENS name. This PR intends to catch those errors and return `null` to ensure behavior is the same.

---

## PR-Codex overview
This PR upgrades the ENS Universal Resolver contract address and improves error handling in the `getEnsAddress` function. 

### Detailed summary
- Upgrades the ENS Universal Resolver contract address
- Improves error handling in the `getEnsAddress` function

### DIFF
--- a/src/actions/ens/getEnsAddress.test.ts
+++ b/src/actions/ens/getEnsAddress.test.ts
@@ -13,7 +13,7 @@ import { http } from '../../clients/transports/http.js'
 import { getEnsAddress } from './getEnsAddress.js'
 
 beforeAll(async () => {
-  await setBlockNumber(16966585n)
+  await setBlockNumber(16966590n)
   await setVitalikResolver()
 })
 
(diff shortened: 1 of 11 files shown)


src/actions/ens/getEnsAddress.test.ts
```diff
@@ -153,7 +153,7 @@ test('universal resolver contract deployed on later block', async () => {
     "Chain \\"Localhost\\" does not support contract \\"ensUniversalResolver\\".
 
     This could be due to any of the following:
-    - The contract \\"ensUniversalResolver\\" was not deployed until block 16773775 (current block 14353601).
+    - The contract \\"ensUniversalResolver\\" was not deployed until block 16966585 (current block 14353601).
```
tmm, line 156: Unrelated to this PR: Wonder if we should remove/add a flag for skipping block number validation in `getChainContractAddress` since you should be able to use a block number before the universal resolver contract was deployed, right?

tmm (approved):
Needs a changeset, then lgtm

src/actions/ens/getEnsAddress.test.ts
jxom: Good question! Depends if `eth_call` against an old block would work. Would have to check.

Issue #606, opened 2023-05-30
bug: `signMessage` generates incorrect signature for `Hex` `message`


moldy530:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.3.39

### Current Behavior

Currently, `stringToBytes` (used by `hashMessage` when signing a message) takes the string and UTF-8 encodes the string and converts it to a byte array. This is a problem when trying to interface with a [`SimpleAccount`](https://github.com/eth-infinitism/account-abstraction/blob/main/contracts/samples/SimpleAccount.sol) because the generated signature is not correct. The expected signature is an `ERC-191` signed hash of the UserOp.

This was was only an issue after `0.3.17`

### Expected Behavior

I'd expect `stringToBytes` to check if the string is already a `Hex` and then just call `hexToBytes` instead of UTF-8 encoding the hex string.

### Steps To Reproduce

```ts
// This is one of the test private keys in this repo
const ownerAccount = privateKeyToAccount(
    "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
);

expect(
  await ownerAccount.signMessage({
    message: "0xa70d0af2ebb03a44dcd0714a8724f622e3ab876d0aa312f0ee04823285d6fb1b"
  }).toBe("0x1a7fd732f926be0a1dbfd1b4acbd7a6f7198fda1ab5ef69ef027a2eee89987785d210298e12119a1d62b5a7d2d9f24d075682675ae00f51a5602e725f0d180851b")
)
```

PR #607, opened 2023-05-30
fix: address #606 by converting hex string directly to byte array

moldy530:
This fixes https://github.com/wagmi-dev/viem/issues/606 by checking if the string passed to stringToBytes is a hex string already, and, if it is, converting it to a byte array directly instead of UTF-8 encoding the string.

---

## PR-Codex overview
This PR adds support for converting hexadecimal strings to byte arrays and uses it to sign messages. 

### Detailed summary
- Added `isHex` function to check if a string is a valid hexadecimal value.
- Added `hexToBytes` function to convert a hexadecimal string to a byte array.
- Modified `stringToBytes` function to use `hexToBytes` when the input is a valid hexadecimal string.
- Updated `signMessage` function to accept a hexadecimal string as the message input and convert it to a byte array before signing.


jxom:
Closing in favor of #610. Added you as a co-author of the commit.

PR #633, opened 2023-06-01
feat: support multiple events in getLogs

holic:
Allows for `event` arg in `getLogs` to accept an array of events.

Just wanted to get a gut check on this before I add tests+docs.

---

## PR-Codex overview
This PR focuses on improving type definitions and adding support for multiple events in the getLogs function.

### Detailed summary
- Updated type definitions in various files.
- Added support for multiple events in the getLogs function.

> The following files were skipped due to too many changes: `src/actions/public/getFilterChanges.test.ts`, `src/actions/public/getFilterChanges.test-d.ts`, `src/actions/public/watchEvent.test.ts`, `src/clients/decorators/public.ts`, `src/actions/public/getLogs.ts`, `src/actions/public/createEventFilter.ts`, `src/actions/public/watchEvent.ts`

### DIFF
--- a/src/actions/public/createEventFilter.test.ts
+++ b/src/actions/public/createEventFilter.test.ts
@@ -30,6 +30,27 @@ const event = {
     name: 'Transfer',
     type: 'event',
   },
+  approve: {
+    type: 'event',
+    name: 'Approval',
+    inputs: [
+      {
+        indexed: true,
+        name: 'owner',
+        type: 'address',
+      },
+      {
+        indexed: true,
+        name: 'spender',
+        type: 'address',
+      },
+      {
+        indexed: false,
+        name: 'value',
+        type: 'uint256',
+      },
+    ],
+  },
   unnamed: {
     inputs: [
       {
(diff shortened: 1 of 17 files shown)


jxom:
The only caveat I see with this is that we won't be able to decode the `topics` & `data` from the logs, as the `topics` you pass in are opaque.

What if we make `event` also support an array of `AbiEvent`s?

holic:
> What if we make `event` also support an array of `AbiEvent`s?

That could work for my specific use case! Just thought I'd do the more general thing.

> The only caveat to that approach will be that you can't filter on indexed arguments, but you can still decode logs.

Is that because `args` would still be separate? I wonder if instead of an array of events, it's an array of event/args pairs?

jxom:
> Is that because args would still be separate? I wonder if instead of an array of events, it's an array of event/args pairs?

It's because you can't filter on multiple events and filter by indexed arguments on the one `eth_getLogs` call. The alternative would be to use multiple `eth_getLogs` calls, but I don't see the point (you can just use multiple `publicClient.getLog`s).

holic:
Feeling a little out of my depth in terms of API design + wrangling types to get this to work. I briefly tried both an `abi` arg and `event: AbiEvent[]` arg, but it feels fragile. I pushed up a prototype with the latter.

I am wonder if, instead of raw `topics`, that the method supports either a single `event` + `args` or an array of `event`s? Then construct topics from the event array.

holic:
> What if we make `event` also support an array of `AbiEvent`s? The only caveat to that approach will be that you can't filter on indexed arguments, but you can still decode logs.

Just updated the PR with this approach after getting more familiar with `getLogs` usage. Is this what you had in mind?

holic:
I must have the types wrong because adding another event ABI to the array gives me an error.

image
(`event.secondary` represents the `Approval` event)

jxom:
Going to fix up the types in this PR, then we can look at merging.

jxom:
Closing in favor of #951 so we can get CI working.

PR #661, opened 2023-06-07
Fix tree shaking

jeetiss:
hello, folks

I fixed two tree-shaking issues with this PR

- `package.json` in `dist/esm` overrides default one so webpack doesn't see `sideEffects: false` option and disables optimizations. I added this field.
- webpack and other tools don't remove top-level function calls even when `sideEffects: false` is specified, so I added `/*#__PURE__*/` annotation with babel plugin

### size before:

```
  viem (cjs)
  Size limit:   60 kB
  Size:         59.77 kB with all dependencies, minified and gzipped
  Loading time: 1.2 s    on slow 3G
  Running time: 427 ms   on Snapdragon 410
  Total time:   1.6 s
  
  viem (tree-shaking check)
  Package size limit has exceeded by 16.75 kB
  Size limit:   1.5 kB
  Size:         18.25 kB with all dependencies, minified and gzipped
  Loading time: 357 ms   on slow 3G
  Running time: 203 ms   on Snapdragon 410
  Total time:   560 ms
  
  viem/chains
  Package size limit has exceeded by 1.76 kB
  Size limit:   1 kB
  Size:         2.76 kB with all dependencies, minified and gzipped
  Loading time: 54 ms   on slow 3G
  Running time: 98 ms   on Snapdragon 410
  Total time:   152 ms
```

### size after:

```
  viem (cjs)
  Size limit:   60 kB
  Size:         59.77 kB with all dependencies, minified and gzipped
  Loading time: 1.2 s    on slow 3G
  Running time: 404 ms   on Snapdragon 410
  Total time:   1.6 s
  
  viem (tree-shaking check)
  Size limit:   1.5 kB
  Size:         1.43 kB with all dependencies, minified and gzipped
  Loading time: 28 ms   on slow 3G
  Running time: 83 ms   on Snapdragon 410
  Total time:   111 ms
  
  viem/chains
  Size limit:   1 kB
  Size:         993 B with all dependencies, minified and gzipped
  Loading time: 20 ms on slow 3G
  Running time: 76 ms on Snapdragon 410
  Total time:   96 ms
```

---

## PR-Codex overview
This PR adds pure annotation to the codebase and updates some dependencies. 

### Detailed summary
- Adds pure annotation to the codebase using the `babel-plugin-annotate-pure-calls` plugin.
- Updates the `@babel/cli` and `@babel/core` dependencies.
- Updates the `@babel/plugin-syntax-typescript` dependency.
- Updates the `size-limit` configuration to include new size limits for different parts of the codebase.
- Adds `sideEffects` to the `package.json` file to optimize tree-shaking.
- Adds a new script to build a pure-annotations version of the codebase.
- Makes minor changes to code formatting in various files.

> The following files were skipped due to too many changes: `src/chains.ts`, `pnpm-lock.yaml`


fubhy:
Hey there. Good find!

> * `package.json` in `dist/esm` overrides default one so webpack doesn't see `sideEffects: false` option and disables optimizations. I added this field.

Is that a bug in webpack? Doesn't hurt to add that property here but feels like webpack should traverse to & inherit that from the package.json of the package.

> webpack and other tools don't remove top-level function calls even when sideEffects: false is specified, so I added /*#__PURE__*/ annotation with babel plugin
Instead of introducing another build step & babel for that, let's just add that annotation to the offending root level function calls in `viem/chains` and wherever else necessary?

jeetiss:
> Is that a bug in webpack?

I don't know exactly, but it seems as not. `esbuild` works the same way, and includes all classes when only `defineBlock` is imported https://bundlejs.com/?q=viem&treeshake=%5B%7B+defineBlock+%7D%5D&config=%7B%22esbuild%22%3A%7B%22minify%22%3Afalse%7D%7D

> another build step & babel

I wouldn't say I like this either, but this solution is bulletproof and typescript doesn't provide any way to achieve the same result.

Current code contains 110 pure annotations:

Снимок экрана 2023-06-07 в 16 02 04

Any usage without annotations would include the `function` and the `call` to bundle, so I think is better to use an automated solution.

But ofc, I can set comments manually if you don't like the costs of `babel`

fubhy:
> Any usage without annotations would include the function and the call to bundle, so I think is better to use an automated solution.

Oh okay, so the annotation is needed for more than just the `viem/chains` use of `defineChain`?

jeetiss:
yes, it is used for all define functions: `defineBlock`, `defineTransaction`, `defineChain`, `defineFormatter`, `defineTransactionReceipt`, `defineTransactionRequest`

Pure annotations affect `viem` and other entries too, but not as dramatically as `viem/chains`

jxom:
Just gave this a test run, and it seems to strip out all the sourcemap comments. Can we stop it from doing that?

fubhy:
@jeetiss Could you split this into two PRs? One for just adding the `sideEffects: false` to the package.jsons and the other for setting up the PURE annotations?

We are currently researching our options for that and don't have a super clear picture yet. A lot of the discussions both in the typescript repository are super recent and inconclusive:

- https://github.com/microsoft/TypeScript/issues/17181

We also found some interesting related discussions and recent activity for rollup, esbuild, etc. that are relevant:

- https://github.com/rollup/rollup/pull/5024
- https://github.com/evanw/esbuild/issues/3149

I dislike the proposed babel solution because it feels somewhat icky to post-process the build output with an additional step effectively fiddling with the dist artifacts after build. I'm worried about how that might potentially affect source maps too.

jeetiss:
@fubhy yes ofc, I will split this PR `sideEffects: false` gain most improvements here and are not much controversial as `pure` annotations.

I found that I can run `babel` over typescript sources, so I can do your suggestion automatically, it just adds pure to every top-level invocation

> Instead of introducing another build step & babel for that, let's just add that annotation to the offending root level function calls in viem/chains and wherever else necessary?

It looks like a good compromise for me before `NO_SIDE_EFFECTS` is settled well

What do you think?

fubhy:
> I found that I can run `babel` over typescript sources, so I can do your suggestion automatically, it just adds pure to every top-level invocation

Yeah that sounds good. Adding them to the source code would also give us an idea how many of those are necessary and how much maintenance burden that would add going forward.

@jxom Also identified a potential problem with how `tsc` is emitting code for `errors/base.ts`. We'd have to check that too but that could also be another follow-up.

src/utils/rpc.ts
```diff
@@ -144,7 +144,7 @@ export type Socket = WebSocket & {
   subscriptions: CallbackMap
 }
 
-const sockets = new Map<string, Socket>()
+const sockets = /*#__PURE__*/ new Map<string, Socket>()
```
fubhy, line 147: And this!

src/utils/promise/withCache.ts
```diff
@@ -1,5 +1,5 @@
-export const promiseCache = new Map()
-export const responseCache = new Map()
+export const promiseCache = /*#__PURE__*/ new Map()
+export const responseCache = /*#__PURE__*/ new Map()
```
fubhy, line 2: And these.

src/utils/encoding/toBytes.ts
```diff
@@ -6,7 +6,7 @@ import { pad } from '../data/pad.js'
 import { assertSize } from './fromHex.js'
 import { type NumberToHexOpts, numberToHex } from './toHex.js'
 
-const encoder = new TextEncoder()
+const encoder = /*#__PURE__*/ new TextEncoder()
```
fubhy, line 9: For these cases, are bundlers clever enough to identify these as dependencies of exported functions? If this gets eliminated by tree shaking but the function that uses this is imported, we got a runtime error. Let's ensure that this behaves as expected!

src/utils/encoding/toHex.ts
```diff
@@ -4,7 +4,7 @@ import { pad } from '../data/pad.js'
 
 import { assertSize } from './fromHex.js'
 
-const hexes = Array.from({ length: 256 }, (_v, i) =>
+const hexes = /*#__PURE__*/ Array.from({ length: 256 }, (_v, i) =>
```
fubhy, line 7: Also here.

src/utils/encoding/toHex.ts
```diff
@@ -201,7 +201,7 @@ export type StringToHexOpts = {
   size?: number
 }
 
-const encoder = new TextEncoder()
+const encoder = /*#__PURE__*/ new TextEncoder()
```
fubhy, line 204: And this.

src/utils/observe.ts
```diff
@@ -3,11 +3,11 @@ import type { MaybePromise } from '../types/utils.js'
 type Callback = ((...args: any[]) => any) | undefined
 type Callbacks = Record<string, Callback>
 
-export const listenersCache = new Map<
+export const listenersCache = /*#__PURE__*/ new Map<
   string,
   { id: number; fns: Callbacks }[]
 >()
-export const cleanupCache = new Map<string, () => void>()
+export const cleanupCache = /*#__PURE__*/ new Map<string, () => void>()
```
fubhy, line 10: And these.

src/utils/promise/createBatchScheduler.ts
```diff
@@ -29,7 +29,7 @@ export type CreateBatchSchedulerReturnType<
     : (args: TParameters) => Promise<Resolved<TReturnType>>
 }
 
-const schedulerCache = new Map<number | string, SchedulerItem[]>()
+const schedulerCache = /*#__PURE__*/ new Map<number | string, SchedulerItem[]>()
```
fubhy, line 32: And this.

jeetiss:
This PR contains a command to add `pure` annotations, if you want to run it again

close in favor of #687

PR #672, opened 2023-06-08
fix(tsconfig): explicit module constraints

sambacha:
These options are viral: enabling them in a package requires all downstream consumers to enable them as well.

Consumers can now opt into these semantics, it also does not require them to do so. Consumers can always safely use alternative import syntaxes (including falling back to `require() `and i`mport()`), or can enable these flags and opt into this behavior themselves.

---

## PR-Codex overview
This PR turns off `esModuleInterop` and `allowSyntheticDefaultImports` in the `tsconfig` file.

### Detailed summary
- `esModuleInterop` and `allowSyntheticDefaultImports` are set to `false` in `tsconfig.base.json`.

### DIFF
--- a/.changeset/khaki-coats-cry.md
+++ b/.changeset/khaki-coats-cry.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Turned off `esModuleInterop` & `allowSyntheticDefaultImports` in tsconfig.
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -24,6 +24,8 @@
     "checkJs": false,
 
     // Interop constraints
+    "esModuleInterop": false,
+    "allowSyntheticDefaultImports": false,
     "forceConsistentCasingInFileNames": true,
     "verbatimModuleSyntax": true,
 


sambacha:
> These options are viral: enabling them in a package requires all downstream consumers to enable them as well.
> 
> Consumers can now opt into these semantics, it also does not require them to do so. Consumers can always safely use alternative import syntaxes (including falling back to `require() `and i`mport()`), or can enable these flags and opt into this behavior themselves.
> 
> ## PR-Codex overview
> This PR focuses on updating the `tsconfig.base.json` file by setting `esModuleInterop` and `allowSyntheticDefaultImports` to `false`, and enabling `forceConsistentCasingInFileNames` and `verbatimModuleSyntax`.
> 
> ### Detailed summary
> * Set `esModuleInterop` to `false`
> * Set `allowSyntheticDefaultImports` to `false`
> * Enable `forceConsistentCasingInFileNames`
> * Enable `verbatimModuleSyntax`
> 
> > ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

```
> * Enable `forceConsistentCasingInFileNames`
> * Enable `verbatimModuleSyntax`
```

This was not enabled, look at the diff.

fubhy:
Yeah this is good.

fubhy:
Tried to do that with https://github.com/wagmi-dev/viem/pull/391/files but didn't realize that it's implicitly enabled due to NodeNext. Whoops.

fubhy (approved):

jxom:
> enabling them in a package **requires all downstream consumers to enable them as well**.

does this mean this is a breaking change? 😅

fubhy:
> > enabling them in a package **requires all downstream consumers to enable them as well**.
> 
> does this mean this is a breaking change? sweat_smile

Sam is disabling them here. Currently they are implicitly enabled. So as it stands, consumers are affected by it. Disabling it *should* not cause any breaking changes but simply be more lenient on consumers.

jxom:
Gotcha! I’m all good with it then!

PR #677, opened 2023-06-08
feat: `getFunctionSelector` for `AbiFunction`

hexcowboy:
https://github.com/wagmi-dev/viem/discussions/676

> Right now I don't see any utility to get a function selector from an AbiFunction. It would be useful to have this to avoid having to convert the AbiFunction to a string that can be parsed by the current getFunctionSelector, which only accepts ABI function strings right now.

This PR updates the `getFunctionSelector` only parameter to accept `string | AbiFunction` to get the hash of an ABI function item.

---

## PR-Codex overview

> The following files were skipped due to too many changes: `src/utils/hash/hashFunction.test.ts`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,8 +1,4 @@
-lockfileVersion: '6.1'
-
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
+lockfileVersion: '6.0'
 
 overrides:
   viem: workspace:*
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
 
 overrides:
   viem: workspace:*
--- a/src/utils/hash/getFunctionSelector.ts
+++ b/src/utils/hash/getFunctionSelector.ts
@@ -1,5 +1,12 @@
 import { slice } from '../data/slice.js'
+import { AbiFunction } from 'abitype'
 
-import { hashFunction } from './hashFunction.js'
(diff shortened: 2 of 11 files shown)


src/utils/hash/getFunctionSelector.ts
```diff
@@ -1,5 +1,12 @@
...
-import { hashFunction } from './hashFunction.js'
+import { hashAbiFunction, hashFunction } from './hashFunction.js'
 
-export const getFunctionSelector = (fn: string) => slice(hashFunction(fn), 0, 4)
+export const getFunctionSelector = (fn: string | AbiFunction) => {
+  if (typeof fn === 'string') {
+    return slice(hashFunction(fn), 0, 4)
+  } else {
+    return slice(hashAbiFunction(fn), 0, 4)
+  }
```
jxom, line 11:
```suggestion
  if (typeof fn === 'string')
    return slice(hashFunction(fn), 0, 4)
  return slice(hashAbiFunction(fn), 0, 4)
```

pnpm-lock.yaml
jxom, line 1: Let's revert this one.

jxom:
Looks good! Can we do the same for `AbiEvent` and `getEventSelector`?

pnpm-lock.yaml
hexcowboy: https://github.com/wagmi-dev/viem/pull/677/commits/6d6242d8d9d9530b7edf69e6d36bd781937db7e7

hexcowboy:
> Looks good! Can we do the same for `AbiEvent` and `getEventSelector`?

added in previous two commits!

jxom (approved):
lfg

hexcowboy:
i'm unsure what to do about codecov inaccurate "uncovered" lines. the tests cover all control flows.

also unsure about failing tests - not sure if they're relevant to this pr

PR #687, opened 2023-06-09
add __PURE__ annotations to source code

jeetiss:
this is the second part of
- #661

---

## PR-Codex overview
This PR adds `/*#__PURE__*/` annotations for better tree-shaking and makes other minor changes like using `Array.from` instead of spread syntax. 

### Detailed summary
- Added `/*#__PURE__*/` annotations for better tree-shaking
- Replaced spread syntax with `Array.from`
- Wrapped `defineBlock`, `defineTransactionReceipt`, `defineTransactionRequest`, and `defineTransaction` in `/*#__PURE__*/`
- Wrapped `Map` and `TextEncoder` in `/*#__PURE__*/`
- Added `formatters` object to `celo` chain definition

> The following files were skipped due to too many changes: `src/chains.ts`

### DIFF
--- a/src/utils/encoding/toBytes.ts
+++ b/src/utils/encoding/toBytes.ts
@@ -6,7 +6,7 @@ import { pad } from '../data/pad.js'
 import { assertSize } from './fromHex.js'
 import { type NumberToHexOpts, numberToHex } from './toHex.js'
 
-const encoder = new TextEncoder()
+const encoder = /*#__PURE__*/ new TextEncoder()
 
 export type ToBytesParameters = {
   /** Size of the output bytes. */
--- a/.changeset/chilled-llamas-tap.md
+++ b/.changeset/chilled-llamas-tap.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added pure annotatations for better tree-shaking
--- a/.changeset/chilled-llamas-tap.md
+++ b/.changeset/chilled-llamas-tap.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Added pure annotatations for better tree-shaking
+Added `/*#__PURE__*/` annotatations for better tree-shaking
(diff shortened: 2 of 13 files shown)


src/utils/encoding/toBytes.ts
```diff
@@ -6,7 +6,7 @@ import { pad } from '../data/pad.js'
 import { assertSize } from './fromHex.js'
 import { type NumberToHexOpts, numberToHex } from './toHex.js'
 
-const encoder = new TextEncoder()
+const encoder = /*#__PURE__*/ new TextEncoder()
```
fubhy, line 9: For these cases, are bundlers clever enough to identify these as dependencies of exported functions? If this gets eliminated by tree shaking but the function that uses this is imported, we got a runtime error. Let's ensure that this behaves as expected!

Also applies to all other usages of the `PURE` annotation for things that are not `defineXXX()` calls.

fubhy: I think we should not add these annotations here unless I'm missing sth. Please enlighten me :-)

jeetiss: > For these cases, are bundlers clever enough to identify these as dependencies of exported functions? 

Yes. But bundlers can't remove the `TextEncoder` invocation safe because it not mark as pure and can have some side effects.

I made a little demo, that shows that webpack can't delete a `Set` invocation from top-level scope without `#__PURE__` [here](https://stackblitz.com/edit/webpack-webpack-js-org-reoaaz?file=src%2Findex.js,src%2Fpure.js,dist%2Fmain.js,dist%2Fpure.js&terminal=build)

fubhy: Gotcha! Okay then :-)

fubhy (approved):
LGTM!

fubhy:
Good catch with the removeComments flag.

jxom (approved):

jxom:
download (4)

PR #666, opened 2023-06-07
Export Functions and Types for creating Custom Serializers

aaronmgdr:
As discussed in https://github.com/wagmi-dev/viem/discussions/623 

While viem supports formatting transactions with different properties the built in serializers do not have this capacity. This Pr  exports functions useful in creating a custom transaction serializer. Then using viem's `toAccount` with a custom signTransaction implementation calling that serializer we can provide viem the ability to handle the tx as a plugin. 

*this is part of a series of PRs aimed at extending viem's ability to support  EIP2718 tx types and creating of Accounts using `toAccount`* 

#### see also 

#665 

## example usage 

```ts
/*
 * Creates a local account with the same capabilities as a normal viem local account
 * as well as the ability to sign CIP42 transactions
 */
export function createCeloAccount(privateKey: `0x${string}`, customSerializer?: (transaction: TransactionSerializableIncludingCIP42) => Promise<string>) {
  const publicKey = privateKeyToPublicKey(privateKey)
  const address = publicKeyToAddress(publicKey)

  const celoAccountSource: CustomSource<TransactionSerializableIncludingCIP42> = {
    address,
    signMessage: async ({message}) => {
      return signMessage({message, privateKey})
    },
    signTypedData: async (typedData) => {
      return signTypedData({privateKey, ...typedData})
    },
    signTransaction: async (transaction) => {
      return signTransaction({privateKey, transaction}, customSerializer)
    }
  }

  return toAccount(celoAccountSource)
}
```

```ts
export function serializeTransaction(tx: TransactionSerializableIncludingCIP42, signature?: Signature): Promise<SerializedCIP42TransactionReturnType> {
  // handle celo's feeCurrency Transactions
  if (couldBeCIP42(tx)) {
    return serializeCIP42Transaction(tx, signature)
  }
  // handle rest of tx types
  return viemSerializeTransaction(tx, signature)
}
```

---

## PR-Codex overview
This PR adds new exports and types to enable alternative `signTransaction` implementations, and includes changes to `serializeTransaction` and `serializeAccessList` functions.

### Detailed summary
- Export `serializeAccessList`, `sign`, `SerializedTransactionReturnType`, `SignTransactionArgs`, and `SignTransactionReturnType`
- Add `SignTransactionArgs` and `SignTransactionReturnType` types to `accounts/index.ts`
- Add `serializeAccessList` function to `index.ts`
- Move `serializeTransaction` to `transaction/serializeTransaction.ts`
[...]

PR #691, opened 2023-06-09
Add Custom Chain Serializers

aaronmgdr:
## TLDR 

Allow account.signTransaction to take an optional transactionSerializer argument.

##Why?

Through custom chain formatters viem supports alternative standards. for formatting transactions but not yet for signing them

## How

1) add to signTransaction  an optional argument for a serializer which 

2) add an optional `serializer` property to chain. 

3) if the current chain has a serializer pass as argument to signTransaction function. 

*It is recommended that custom serializers check if the transaction is the type they support and if not call the default viem serializeTransaction function.*
  
 ## Example 
  
// create a customer Serializer ie 
```ts [serializer.ts]

import { SerializeTransactionFn, serializeTransaction } from 'viem/utils'
import { TransactionSerializable } from 'viem/types'

type CustomTX = TransactionSerializable {
  type: 'pip112' // a string representing the tx type this should serialize to a EIP2718 compliant TransactionType https://eips.ethereum.org/EIPS/eip-2718#transactiontype-only-goes-up-to-0x7f
  extraField: // some field the spec for this tx type supports
}

export const serializer: SerializeTransactionFn<CustomTX> = (transaction) => {
  if (specialCase(transaction)) {
    // serialization code here
  } else {
    return serializeTransaction(transaction)
  }
}

```
// Add it to the chain

``` ts
import {serializer } from './serializer'
import {celo} from 'wagmi/chains'

const chainWithSerializer = defineChain({...celo, serializer})

// use chain when creating client
const client = createWalletClient({
  account,
  chain: chainWithSerializer,
  transport: http()
})

```

## other ways to achieve this
 
 technically its possible to add an custom serializer by creating an account using the `toAccount` function. 
 
 This has 2 limitations. 
 1) you must recreated the `privateKeyToAccount`, `signTransaction` just to modify the `serializeTransaction` function. Since` mneumonicToAccount` and `HDKeyToAccount` use privateKeyToAccount they must also be recreated. 
 
 2) it doesnt give an elegent way for there to be multiple serializers such as different chains both having custom serializers. 
 

## related PRS 

# supersedes https://github.com/wagmi-dev/viem/pull/666

---

## PR-Codex overview
This PR adds support for custom chain serializers and EIP-2930 access lists. It also makes changes to several files, including `utils`, `types`, `accounts`, `actions`, and `transaction`.

### Detailed summary
- Added custom chain serializers via `chain.serializers`
- Added support for EIP-2930 access lists
- Removed `serializeTransaction` from some `utils` files
- Modified `signTransaction` in `accounts/utils/signTransaction.ts`
- Modified `serializeTransaction` in `utils/transaction/serializeTransaction.ts`
- Modified `Account` type in `accounts/types.ts`
- Modified `Chain` type in `types/chain.ts`
- Added tests for `serializeAccessList` and `signTransaction`

> The following files were skipped due to too many changes: `src/accounts/utils/signTransaction.test.ts`, `src/utils/transaction/serializeTransaction.ts`, `src/actions/wallet/sendTransaction.test.ts`, `site/docs/accounts/custom.md`

### DIFF
--- a/.changeset/calm-garlics-share.md
+++ b/.changeset/calm-garlics-share.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
+Added ability for chains to specify their own transaction serializer. Exported serializeAccessList and type SerializeTransactionFn.
--- a/.changeset/calm-garlics-share.md
+++ b/.changeset/calm-garlics-share.md
@@ -2,4 +2,4 @@
 "viem": minor
 ---
 
-Added ability for chains to specify their own transaction serializer. Exported serializeAccessList and type SerializeTransactionFn.
+Added custom chain serializers via `chain.serializers`. 
--- a/site/docs/accounts/custom.md
+++ b/site/docs/accounts/custom.md
@@ -36,7 +36,7 @@ const account = toAccount({
   async signMessage({ message }) {
     return signMessage({ message, privateKey })
   },
-  async signTransaction(transaction, serializer) {
+  async signTransaction(transaction, { serializer }) {
     return signTransaction({ privateKey, transaction, serializer })
   },
   async signTypedData(typedData) {
@@ -65,7 +65,7 @@ const account = toAccount({
   async signMessage({ message }) {
     return signMessage({ message, privateKey })
   },
-  async signTransaction(transaction, serializer) {
+  async signTransaction(transaction, { serializer }) {
     return signTransaction({ privateKey, transaction, serializer })
   },
   async signTypedData(typedData) {
@@ -85,7 +85,7 @@ const account = toAccount({
   async signMessage({ message }) { // [!code focus:3]
     return signMessage({ message, privateKey })
   },
-  async signTransaction(transaction, serializer) {
+  async signTransaction(transaction, { serializer }) {
     return signTransaction({ privateKey, transaction, serializer })
   },
   async signTypedData(typedData) {
@@ -104,7 +104,7 @@ const account = toAccount({
   async signMessage({ message }) {
     return signMessage({ message, privateKey })
   },
-  async signTransaction(transaction, serializer) {  // [!code focus:3]
+  async signTransaction(transaction, { serializer }) {  // [!code focus:3]
     return signTransaction({ privateKey, transaction, serializer })
   },
   async signTypedData(typedData) {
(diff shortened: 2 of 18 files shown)


aaronmgdr:
One issue Im running into is that the TS types for the return values of functions are very specific only allowing for specific strings. 

for instance 

```ts
 type TransactionSerialized<TType extends TransactionType = 'legacy'> =
  | (TType extends 'eip1559' ? TransactionSerializedEIP1559 : never)
  | (TType extends 'eip2930' ? TransactionSerializedEIP2930 : never)
  | (TType extends 'legacy' ? TransactionSerializedLegacy : never)

```
*and*

```ts
export type GetTransactionType<
  TTransactionSerializable extends TransactionSerializable = TransactionSerializable,
> =
  | (TTransactionSerializable['maxFeePerGas'] extends bigint
      ? 'eip1559'
      : never)
  | (TTransactionSerializable['maxPriorityFeePerGas'] extends bigint
      ? 'eip1559'
      : never)
  | (TTransactionSerializable['gasPrice'] extends bigint
      ? TTransactionSerializable['accessList'] extends AccessList
        ? 'eip2930'
        : 'legacy'
      : never)
  | (TTransactionSerializable['type'] extends string
      ? TTransactionSerializable['type']
      : never)
      
```

It seems like the solution will be for `TransactionType` to be something other than  `"eip1559" | "eip2930" | "legacy"` however i don't have a nice way to so so in mind yet. maybe its just adding 'eip2718' to mean tx is a kind of eip2718 tx but that doesnt seems right.

aaronmgdr:
ive writen up what I hope to be a doc on how to use this but Im not sure where in the docs it should go 

---

## Using custom transaction serializer

For Local Accounts `account.signTransaction` can take an optional second argument to serialize the transaction object before signing. If you are using `walletClient.sendTransaction` with a chain that has a serializer defined it will automatically be used to sign the transaction.

::: code-group

```ts [example.ts]
import { account, walletClient } from './config'
import { serializer } from './serializer'

const signedTransaction = await account.signTransaction(transaction, serializer)

```

```ts [serializer.ts]

import { SerializeTransactionFn, serializeTransaction } from 'viem/utils'
import { TransactionSerializable } from 'viem/types'

type CustomTX = TransactionSerializable {
  type: 'pip112' // a string representing the tx type this should serialize to a EIP2718 compliant TransactionType https://eips.ethereum.org/EIPS/eip-2718#transactiontype-only-goes-up-to-0x7f
  extraField: // some field the spec for this tx type supports
}

export const serializer: SerializeTransactionFn<CustomTX> = (transaction) => {
  if (specialCase(transaction)) {
    // serialization code here
  } else {
    return serializeTransaction(transaction)
  }
}

```

aaronmgdr:
/codex show me changes made by this pr

PR #694, opened 2023-06-10
docs(simulateContract): BaseError do not cause

nalinbhardwaj:
Check `error` to be an instance of `ContractFunctionExecutionError`, as BaseError don't have a `cause`, making the current example code incorrect?

---

## PR-Codex overview
This PR updates an import statement and changes an error checking condition in `simulateContract.md`. 

### Detailed summary:
- Updated import statement to use `ContractFunctionExecutionError` instead of `BaseError`
- Changed error checking condition to use `ContractFunctionExecutionError` instead of `BaseError`

### DIFF
--- a/site/docs/contract/simulateContract.md
+++ b/site/docs/contract/simulateContract.md
@@ -189,7 +189,7 @@ You can access the custom error through the `data` attribute of the error:
 ::: code-group
 
 ```ts {13-27} [example.ts]
-import { BaseError, ContractFunctionRevertedError } from 'viem';
+import { ContractFunctionExecutionError, ContractFunctionRevertedError } from 'viem';
 import { account, walletClient, publicClient } from './config'
 import { wagmiAbi } from './abi'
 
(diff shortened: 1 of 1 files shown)


fubhy:
`cause` is a property of the standard `Error` class: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

nalinbhardwaj:
> `cause` is a property of the standard `Error` class: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

Bizarre, if I use `BaseError` my editor complains, but changing it to `ContractFunctionExecutionError ` fixes it. Any idea why?
image

fubhy:
That sounds like your tsconfig is targeting a really old `target` or has misconfigured `lib`. Try setting your `target` to `es2022`  and make sure that your `lib` either isn't configured at all (defaults to `es2022` then too) or has `es2022` in the array too.

So e.g.

```json
{
  "compilerOptions": {
    "target": "es2022",
    "lib": ["es2022"],
  },
}
```

nalinbhardwaj:
@fubhy thanks, good catch; that seems to fix it. Closing.

PR #696, opened 2023-06-11
fix: decimals with leading zeroes when rounding up

xenoliss:
PR aiming at fixing the stripping of the decimal leading zeroes when using `parseUnits` when the last decimal is rounded-up.

Explanation: 
Simplified the logic by removing the `zeroes` variable and using the `length` of the `left` decimals part + 1 (the `unit` "slot") to handle the rounding.

---

## PR-Codex overview
This PR improves the `parseUnits` function in `src/utils/unit/parseUnits.ts` to handle decimal precision more accurately. 

### Detailed summary
- Replaces `Math.floor` with `Math.round` for more accurate rounding
- Simplifies code by removing unnecessary string manipulation
- Adds tests for edge cases with decimal precision


jxom:
good catch!

PR #744, opened 2023-06-21
export formatLog

holic:
I am playing around with different approaches for fetching logs from the RPC and realized that this method wasn't exposed yet.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -128,6 +128,7 @@ export {
   formatTransaction,
   transactionType,
 } from './formatters/transaction.js'
+export { formatLog } from './formatters/log.js'
 export {
   type FormattedTransactionReceipt,
   defineTransactionReceipt,


jxom:
May as well export this from the root too

holic:
> May as well export this from the root too

I originally did that but then noticed there weren't many of the formatter exports being used in the root, so I opted to keep it in just utils. I defer to you though!

holic:
Done!

Any reason in particular ya'll don't re-export everything from modules/categories, like
```
export * from './utils';
export * from './actions';
```
I assume the original intention was to be selective about what's exported but it seems in practice that ya'll eventually export ~everything?

jxom:
Yeah, that's correct! We have better visibility over what's being exported and being consumed by the public.

holic:
> Yeah, that's correct! We have better visibility over what's being exported and being consumed by the public.

Makes sense! Since ya'll already build separate files for certain modules, I figured they'd already pass the "is public" test and could just be wholesale exported. So the root just becomes an export of all individually built files/modules, plus any extra that aren't covered (e.g. actions).

But I also understand preferring less indirection!

PR #754, opened 2023-06-21
Fix `EIP712Domain` to support empty string name

jgalat:
Hi,

We are migrating our web app to support the latest version of `viem`. However we were blocked by being unable to sign messages using `signTypedData` with an empty string as `name` in the domain structure.

According to the [spec](https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator) fields should not be considered **only if absent**, but the current implementation doesn't take into account that `name` could be `''` which is a **valid string**.

Thanks!

---

## PR-Codex overview
This PR adds support for empty string in the `EIP712Domain` name field. It includes changes to `signTypedData.ts`, `hashTypedData.ts`, `hashTypedData.test.ts`, and `signTypedData.test.ts`.

### Detailed summary:
- Added support for empty string in `EIP712Domain` name field
- Changed comparison from `&&` to `!= null` in `signTypedData.ts` and `hashTypedData.ts`
- Added test case for empty name in `hashTypedData.test.ts` and `signTypedData.test.ts`

### DIFF
--- a/.changeset/chilly-dots-sparkle.md
+++ b/.changeset/chilly-dots-sparkle.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added support for empty string in EIP712Domain name field.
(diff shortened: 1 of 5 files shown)


jxom:
Thanks for the PR! Just recreated it on viem's origin (had trouble pushing to your origin), see #758.

Your commits are still retained, so you still get credit. :)

PR #771, opened 2023-06-23
Support sending simulate calls in batch via Multicall

### LINKED ISSUE #766: bug: Multicall Batching fails to batch calls with `simulate`

### Package Version

1.1.4

### Current Behavior

When using multicall batching config: 
```
{
    batch: {
      multicall: true,
    },
  }
```

And calling a set of functions:
```
const myContract = getContract(...)
const [fooRes, barRes, bazRes] = await Promise.all([
  myContract.read.foo()
  myContract.simulate.bar()
  myContract.simulate.baz()
])
```

The resulting network requests have a separate `eth_call` request for each `simulate`d call.

### Expected Behavior

Simulated calls, when batched with read calls, should result in a single multicall call.

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

arjun-io:
Fixes #766 by simply removing the `batch: false` hardcode in the `call`.

---

## PR-Codex overview
This PR removes the `batch` property from a function call in `simulateContract.ts` and updates the corresponding test in `simulateContract.test.ts`.

### Detailed summary
- Removed `batch` property from function call in `simulateContract.ts`
- Updated corresponding test in `simulateContract.test.ts` to remove `batch` property and add `data` property with a specific value

### DIFF
--- a/src/actions/public/simulateContract.ts
+++ b/src/actions/public/simulateContract.ts
@@ -127,7 +127,6 @@ export async function simulateContract<
   } as unknown as EncodeFunctionDataParameters<TAbi, TFunctionName>)
   try {
     const { data } = await call(client, {
-      batch: false,
       data: `${calldata}${dataSuffix ? dataSuffix.replace('0x', '') : ''}`,
       to: address,
       ...callRequest,
--- a/src/actions/public/simulateContract.test.ts
+++ b/src/actions/public/simulateContract.test.ts
@@ -168,7 +168,6 @@ test('args: dataSuffix', async () => {
   })
   expect(spy).toHaveBeenCalledWith(publicClient, {
     account: accounts[0].address,
-    batch: false,
     data: '0x1249c58b12345678',
     to: wagmiContractConfig.address,
   })


src/actions/public/simulateContract.ts
```diff
@@ -127,7 +127,6 @@ export async function simulateContract<
   } as unknown as EncodeFunctionDataParameters<TAbi, TFunctionName>)
   try {
     const { data } = await call(client, {
-      batch: false,
```
jxom, line 130: I wonder if this should rather be hoisted as an argument on `simulateContract`, and have it default to `true`. Some consumers may want opt-out of multicall in some cases?

arjun-io: From my read it seems like the `batch` option is omitted for read contract params: https://github.com/wagmi-dev/viem/blob/main/src/actions/public/readContract.ts#L26 (the `Pick` doesn't include `batch`)

Should these args stay consistent with each other? In my view a `simulate` is just a `read` for `payable | nonpayable` functions

jxom:
Sorry! Will take a proper look when I get back from vacation (Aug 18)!

jxom:
Think I might have to close this one. The reasoning is that when we are calling contracts from the multicall contract, the `msg.sender` value will be populated with the multicall address. Currently, we do not batch calls with an explicit `from` attribute, however, when a `from` is not specified, we would expect `msg.sender` to equal a zero address and not the multicall contract address. This is dangerous and will produce some unwanted side-effects for contracts that depend on `msg.sender` (and we cannot deterministically identify if a contract function depends on `msg.sender`). [See more](https://github.com/mds1/multicall#batch-contract-writes)

PR #840, opened 2023-07-03
[Docs] Fix function used in snippet

fvictorio:
`transferFrom` both doesn't make sense and doesn't match what the other tab does.

---

### Detailed summary

- The `transferFrom` function in the `contract` object is replaced with the `mint` function.
- The `from` address in the `watchEvent.Transfer` function is changed to `0xd8da6bf26964af9d7eed9e03e53415d37aa96045`.


jxom:
oops ty

PR #853, opened 2023-07-06
Allow EIP-1559 transactions on chains with baseFeePerGas of 0

johngrantuk:
Currently a baseFeePerGas of 0 fails truthy check and results in `Chain does not support EIP-1559 fees.` error.

This was highlighted when trying to refactor Lattice/MUD project to use Viem instead of Ethers because they use a local Anvil node with with 0 base fee for ease of development.

---

## PR-Codex overview
This PR focuses on allowing the use of EIP-1559 transactions on chains with a base fee of 0.

### Detailed summary
- Added support for using EIP-1559 transactions on chains with 0 base fee.
- Updated the condition to check for a non-null value of `baseFeePerGas` in the `prepareRequest.ts` file.
- Added an error check to ensure that the chain does not support legacy `gasPrice`.

### DIFF
--- a/.changeset/loud-poems-burn.md
+++ b/.changeset/loud-poems-burn.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Allow using EIP-1559 transactions on chains with 0 base fee.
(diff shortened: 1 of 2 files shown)


jxom:
Good catch!

jxom (approved):

PR #876, opened 2023-07-11
docs: simulateContract/Handling Custom Errors - simplify search of `revertError`

Alexsey:
Remove redundant code from the example of finding `revertError`

---

### Detailed summary
- The focus of this PR is to refactor error handling in `simulateContract.md` file.
- It replaces the option 1 of checking the instance of the error with option 2 using the `walk` method from `BaseError`.
- It updates the code to check if `revertError` is an instance of `ContractFunctionRevertedError` before accessing its `data` property.
- The code for handling `errorName` remains the same.

### DIFF
--- a/site/docs/contract/simulateContract.md
+++ b/site/docs/contract/simulateContract.md
@@ -201,11 +201,12 @@ try {
     account,
   })
 } catch (err) {
-  const revertError = err instanceof BaseError
-    && err.walk(err => err instanceof ContractFunctionRevertedError)
-  if (revertError instanceof ContractFunctionRevertedError) {
-    const errorName = revertError.data?.errorName ?? "";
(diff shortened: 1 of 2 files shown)


jxom:
Not sure if the return type of `walk` would infer the condition would it? Could try make use of TypeScript `asserts` for this: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

Alexsey:
Yes, it would be really nice, but after playing a little bit with it, it didn't work for me :(

Anyway, the initial docs have the same problem for both old and new `walk` implementation

Alexsey:
Ok, I have updated the code in the docs, so it will not have the type error. It's not as pretty, but it can be copy-passed now

PR #908, opened 2023-07-19
feat: add support for resolving ens addresses using coinType

moldy530:
ENS address records support coinType keys as defined in [ENSIP-9](https://docs.ens.domains/ens-improvement-proposals/ensip-9-multichain-address-resolution) so that one record can hold addresses for multiple chains. This also applies to evm-based chains so that the record `moldy.eth` on mainnet can hold an optimism address entry. This is supported via the `addr(name, coinType)` method on the Resolver

This PR adds an optional parameter for `coinType` to the `getENSAddress` method which can be used during resolution of an address

---

### Focus of this PR:
Add coin type support for `getEnsAddress` function.

### Detailed summary:
- Added `coinType` parameter to the `getEnsAddress` function in `src/actions/ens/getEnsAddress.ts`.
- Updated the function implementation to handle the `coinType` parameter and fetch the address accordingly.
- Added a new test case in `src/actions/ens/getEnsAddress.test.ts` to test the function with the `coinType` parameter.
- Updated the `addressResolverAbi` in `src/constants/abis.ts` to include a new function definition for `addr` with `coinType` parameter.
- Updated the documentation in `site/docs/ens/actions/getEnsAddress.md` to include the `coinType` parameter description.

### DIFF
--- a/src/actions/ens/getEnsAddress.ts
+++ b/src/actions/ens/getEnsAddress.ts
@@ -3,8 +3,7 @@ import type { Address } from 'abitype'
 import type { Client } from '../../clients/createClient.js'
 import type { Transport } from '../../clients/transports/createTransport.js'
 import {
-  coinTypeAddressResolverAbi,
-  singleAddressResolverAbi,
+  addressResolverAbi,
   universalResolverResolveAbi,
 } from '../../constants/abis.js'
 import type { Chain } from '../../types/chain.js'
@@ -27,7 +26,7 @@ export type GetEnsAddressParameters = Prettify<
     /** Name to get the address for. */
     name: string
     /** ENSIP-9 compliant coinType used to resolve addresses for other chains */
-    coinType?: bigint
+    coinType?: number
     /** Address of ENS Universal Resolver Contract. */
     universalResolverAddress?: Address
   }
@@ -88,18 +87,13 @@ export async function getEnsAddress<TChain extends Chain | undefined,>(
   }
 
   try {
-    const functionData =
-      coinType != null
-        ? encodeFunctionData({
-            abi: coinTypeAddressResolverAbi,
-            functionName: 'addr',
-            args: [namehash(name), coinType],
-          })
-        : encodeFunctionData({
-            abi: singleAddressResolverAbi,
-            functionName: 'addr',
-            args: [namehash(name)],
-          })
+    const functionData = encodeFunctionData({
+      abi: addressResolverAbi,
+      functionName: 'addr',
+      ...(coinType != null
+        ? { args: [namehash(name), BigInt(coinType)] }
+        : { args: [namehash(name)] }),
+    })
 
     const res = await readContract(client, {
       address: universalResolverAddress,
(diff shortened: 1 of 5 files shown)


src/constants/abis.ts
```diff
@@ -112,6 +112,19 @@ export const singleAddressResolverAbi = [
   },
 ] as const
 
+export const coinTypeAddressResolverAbi = [
```
jxom, line 115: Might be better if we move this into `singleAddressResolverAbi` (and rename that to `addressResolverAbi`), then we can rely on viem to discriminate the abi items based on the overloaded args.

```diff
-export const singleAddressResolverAbi = [
+export const addressResolverAbi = [
  {
    name: 'addr',
    type: 'function',
    stateMutability: 'view',
    inputs: [{ name: 'name', type: 'bytes32' }],
    outputs: [{ name: '', type: 'address' }],
  },
+ {
+   name: 'addr',
+   type: 'function',
+   stateMutability: 'view',
+   inputs: [
+     { name: 'name', type: 'bytes32' },
+     { name: 'coinType', type: 'uint256' },
+   ],
+   outputs: [{ name: '', type: 'bytes' }],
+ },
] as const

...

const res = await readContract(client, {
  address: universalResolverAddress,
  abi: universalResolverResolveAbi,
  functionName: 'resolve',
  args: [
    toHex(packetToBytes(name)),
    encodeFunctionData({
-     abi: singleAddressResolverAbi,
+     abi: addressResolverAbi,
      functionName: 'addr',
      args: [
        namehash(name),
+       ...(coinType != null ? [coinType] : [])
      ],
    }),
  ],
  blockNumber,
  blockTag,
})
```

src/actions/ens/getEnsAddress.ts
```diff
@@ -25,6 +26,8 @@ export type GetEnsAddressParameters = Prettify<
   Pick<ReadContractParameters, 'blockNumber' | 'blockTag'> & {
     /** Name to get the address for. */
     name: string
+    /** ENSIP-9 compliant coinType used to resolve addresses for other chains */
+    coinType?: bigint
```
jxom, line 30: We can probably make `coinType` a `number`

src/actions/ens/getEnsAddress.ts
```diff
@@ -84,31 +88,44 @@ export async function getEnsAddress<TChain extends Chain | undefined,>(
...
+            data: res[0],
+          })
+        : decodeFunctionResult({
+            abi: singleAddressResolverAbi,
+            functionName: 'addr',
+            data: res[0],
+          })
 
-    return trim(address) === '0x00' ? null : address
+    return new Set(['0x00', '0x']).has(trim(address)) ? null : address
```
jxom, line 128: Might be better if we just use a couple of equality checks here performance wise

src/actions/ens/getEnsAddress.ts
```diff
@@ -84,31 +87,39 @@ export async function getEnsAddress<TChain extends Chain | undefined,>(
   }
 
   try {
+    const functionData = encodeFunctionData({
+      abi: addressResolverAbi,
+      functionName: 'addr',
+      ...(coinType != null
```
moldy530, line 93: @jxom defining this as:
```ts
args: [
        namehash(name),
       ...(coinType != null ? [coinType] : [])
      ],
```

resulted in a ts error because it was evaluating to `[Hex, ...bigint[]]`. I also tried using type casting here:
```
args: [
        namehash(name),
        ...(coinType != null ? ([coinType] as [bigint]) : ([] as [])),
      ],
```
but that results in:
```
Type '[`0x${string}`] | [`0x${string}`, bigint]' is not assignable to type 'readonly [`0x${string}`, bigint]'.
          Type '[`0x${string}`]' is not assignable to type 'readonly [`0x${string}`, bigint]'.
```

src/actions/ens/getEnsAddress.ts
```diff
@@ -84,31 +87,39 @@ export async function getEnsAddress<TChain extends Chain | undefined,>(
...
 
-    const address = decodeFunctionResult({
-      abi: singleAddressResolverAbi,
-      functionName: 'addr',
-      data: res[0],
-    })
+    const address = trim(
+      decodeFunctionResult({
+        abi: addressResolverAbi,
+        args: coinType != null ? [namehash(name), BigInt(coinType)] : undefined,
```
moldy530, line 112: also odd TS typing issue here too, the typing here for args was `[Hex, bigint] | undefined` might need to dig into how the args typing is generated

jxom (approved):

Issue #836, opened 2023-07-02
bug: watchContractEvent infinitely tries to get updates on a non-existent filter when the filter has expired


iuriiiurevich:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

1.2.7 (latest)

### Current Behavior

There are cases in which a filter (created with `eth_newFilter`) is no longer available (`eth_getFilterChanges` throws `filter not found)`, for example:

1. The filter has expired due to being inactive for 5 minutes (for example, if the user was offline).
2. The node on which the filter was located was dropped because of the changes in infrastructure.

Related: [QuickNode docs](https://support.quicknode.com/hc/en-us/articles/10838914856977-Error-code-32000-message-filter-not-found-), [Alchemy docs](https://docs.alchemy.com/reference/eth-newfilter).

This causes `watchContractEvent` to endlessly try to get updates on a non-existent filter.

### Expected Behavior

As stated in the [QuickNode docs](https://support.quicknode.com/hc/en-us/articles/10838914856977-Error-code-32000-message-filter-not-found-), it would be better to recreate a filter if it throws the "filter not found" error.

### Anything else?

Related: https://github.com/wagmi-dev/viem/discussions/818

Gift-Stack:
@jxom Can I take this?

jxom:
yes

jxom:
Fixed via #954

PR #954, opened 2023-07-30
Add re-initialization logic for event filters

jaxernst:
Here's my proposal to address #836.  Filters can timeout or get into a bad state where reinitialization is necessary to continue getting filter changes on the polling interval. This request adds fallback reinitialization, so if 'getFilterChanges' fails repeatedly, it will attempt to recreate the filter.

---

### Focus of the PR:
This PR focuses on adding filter reinitialization logic for `watchContractEvent` and `watchEvent` when a filter has been uninstalled.

### Detailed summary:
- Added filter reinitialization logic for `watchContractEvent` and `watchEvent` when a filter has been uninstalled.
- Imported `InvalidInputRpcError` in `watchEvent.ts` and `watchContractEvent.ts`.
- Added error handling and filter reinitialization logic in `watchEvent.ts` and `watchContractEvent.ts` if a filter has been set and gets uninstalled.
- Added a test case to verify the filter reinitialization logic in `watchEvent.test.ts` and `watchContractEvent.test.ts`.

### DIFF
--- a/.changeset/brown-horses-design.md
+++ b/.changeset/brown-horses-design.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Add filter reinitialization logic for watchContractEvent and watchEvent
--- a/.changeset/brown-horses-design.md
+++ b/.changeset/brown-horses-design.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Add filter reinitialization logic for watchContractEvent and watchEvent
+Added filter reinitialization logic for `watchContractEvent` and `watchEvent` for when a filter has been uninstalled.
--- a/.changeset/gold-days-juggle.md
+++ b/.changeset/gold-days-juggle.md
@@ -1,5 +1,0 @@
----
-"viem": patch
----
-
-Added event filter reinitialization logic to watchContractEvent
--- a/.changeset/gold-days-juggle.md
+++ b/.changeset/gold-days-juggle.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added event filter reinitialization logic to watchContractEvent
--- a/src/actions/public/watchContractEvent.test.ts
+++ b/src/actions/public/watchContractEvent.test.ts
@@ -596,4 +596,33 @@ describe('errors', () => {
     },
     { retry: 3 },
   )
+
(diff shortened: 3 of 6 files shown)


jxom:
Instead of checking if the filter has errored n times, maybe we can check the error instead for “filter not found”.

Most nodes support this, examples:
https://github.com/ethereum/go-ethereum/blob/master/eth/filters/api.go#L38
https://github.com/paradigmxyz/reth/blob/f41386d28e89dd436feea872178452e5302314a5/crates/rpc/rpc/src/eth/filter.rs#L447

jaxernst:
> Instead of checking if the filter has errored n times, maybe we can check the error for “filter not found”.
> 
> Most nodes support this, examples: https://github.com/ethereum/go-ethereum/blob/master/eth/filters/api.go#L38 https://github.com/paradigmxyz/reth/blob/f41386d28e89dd436feea872178452e5302314a5/crates/rpc/rpc/src/eth/filter.rs#L447

This does make more sense. I was hesitant to do this because I wasn't sure if "filter not found" was standardized, so I looked a bit further. Every client I've checked returns this error except for Erigon:

Besu: https://github.com/hyperledger/besu/blob/main/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/response/RpcErrorType.java#L45

Nethermind:
https://github.com/NethermindEth/nethermind/blob/a099f0fe2158b969ca6b81861c5097ecc5f6c0b1/src/Nethermind/Nethermind.JsonRpc/Modules/Eth/EthRpcModule.cs#L553

Erigon:
https://github.com/ledgerwatch/erigon/blob/56934395210c48585ae8d7f77404d7c9b9ba3a30/turbo/jsonrpc/eth_filters.go#L78


Should we still check for this even if not all clients return it?

jxom:
Thought so. Maybe we can keep the current implementation as a fallback, and check the error first?

jaxernst:
Although Erigon doesn't throw a 'filter not found', they do all appear to return a -32000 (Invalid input) error code when the filter isn't found. It looks like Viem already catches these as "InvalidInputRpcError", so might we just reinitialize the filter if we catch an InvalidInputRpcError?

jaxernst:
Implemented this with tests for watchContractEvent and watchEvent

jxom (approved):
Looks good! Will merge it in soon.

PR #961, opened 2023-08-01
feat: support a string or array of keys for `getEnsText`

gskril:
This allows `getEnsText` to accept an array of keys to resolve via the ENS Universal Resolver. The return type is dynamic based on the `key` parameter. It can still take a string, causing no API changes.

Usage:
```ts
const ensText = await publicClient.getEnsText({
  name: normalize('wagmi-dev.eth'),
  key: 'com.twitter',
})
```

or

```ts
const ensText = await publicClient.getEnsText({
  name: normalize('wagmi-dev.eth'),
  key: ['com.twitter', 'description', 'url'],
})
```

---

### Detailed summary
- Added `universalResolverResolveArrayAbi` constant in `src/constants/abis.ts` and updated its usage in `src/actions/ens/getEnsText.ts`.
- Updated `getEnsText` function in `src/actions/ens/getEnsText.ts` to handle multiple text records.
- Updated documentation in `site/docs/ens/actions/getEnsText.md` to reflect the changes.

### DIFF
--- a/src/actions/ens/getEnsText.ts
+++ b/src/actions/ens/getEnsText.ts
@@ -20,7 +20,9 @@ import {
   readContract,
 } from '../public/readContract.js'
 
-export type GetEnsTextParameters<TKeys extends  string | readonly string[] = string | readonly string[]> = Prettify<
+export type GetEnsTextParameters<
+  TKeys extends string | readonly string[] = string | readonly string[],
+> = Prettify<
   Pick<ReadContractParameters, 'blockNumber' | 'blockTag'> & {
     /** ENS name to get Text for. */
     name: string
@@ -31,8 +33,10 @@ export type GetEnsTextParameters<TKeys extends  string | readonly string[] = str
   }
 >
 
-// if a single key is passed, return a single string or null. otherwise, return an array of strings or nulls.
-export type GetEnsTextReturnType<TKeys extends  string | readonly string[] = string | readonly string[]> =
+// If a single key is passed, return a single string or null. Otherwise, return an array of strings or nulls.
+export type GetEnsTextReturnType<
+  TKeys extends string | readonly string[] = string | readonly string[],
+> =
   | (TKeys extends string ? string | null : never)
   | (TKeys extends readonly string[] ? (string | null)[] | null : never)
 
@@ -135,7 +139,9 @@ export async function getEnsText<
       decodedRecords.length === 1 ? decodedRecords[0] : decodedRecords
     ) as GetEnsTextReturnType<TKeys>
   } catch (err) {
-    if (isNullUniversalResolverError(err, 'resolve')) return null
+    if (isNullUniversalResolverError(err, 'resolve')) {
+      return null as GetEnsTextReturnType<TKeys>
+    }
     throw err
   }
 }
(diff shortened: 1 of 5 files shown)


jxom:
👑

src/actions/ens/getEnsText.ts
```diff
@@ -20,21 +20,24 @@ import {
   readContract,
 } from '../public/readContract.js'
 
-export type GetEnsTextParameters = Prettify<
+export type GetEnsTextParameters<TKeys> = Prettify<
```
tmm, line 23: Adding constraint + default

```suggestion
export type GetEnsTextParameters<TKeys extends  string | readonly string[] = string | readonly string[]> = Prettify<
```

src/actions/ens/getEnsText.ts
```diff
@@ -62,16 +65,19 @@ export type GetEnsTextReturnType = string | null
  * })
  * // 'wagmi_sh'
  */
-export async function getEnsText<TChain extends Chain | undefined>(
+export async function getEnsText<
+  TChain extends Chain | undefined,
+  TKeys extends string | string[],
```
tmm, line 70: Usually we try to use `readonly` arrays if we can.

```suggestion
  TKeys extends string | readonly string[],
```

src/actions/ens/getEnsText.ts
```diff
@@ -20,21 +20,24 @@ import {
...
     /** Address of ENS Universal Resolver Contract. */
     universalResolverAddress?: Address
   }
 >
 
-export type GetEnsTextReturnType = string | null
+// if a single key is passed, return a single string or null. otherwise, return an array of strings or nulls.
+export type GetEnsTextReturnType<T> = T extends string
+  ? string | null
+  : (string | null)[] | null
```
tmm, line 37: Removing conditional type:

```suggestion
export type GetEnsTextReturnType<TKeys extends  string | readonly string[] = string | readonly string[]> =
  | (TKeys extends string ? string | null : never)
  | (TKeys extends readonly string[] ? (string | null)[] | null : never)
```

tmm (commented):
Looking solid! Few TS-related comments. Can add a changeset with `pnpm changeset` when you are ready too.

gskril:
What do you guys think the return type should be when an array of keys is passed? 

`(string | null)[]` in the same order as the supplied keys (what it is now) or `Record<Key, string | null>` so that the values can be accessed by key (which I think would also bring autocomplete via intellisense)?

@lucemans was [tinkering around with the 2nd option](https://github.com/gskril/viem/pull/1/files) but I figured people might want this to remain consistent with other parts of viem so thought I'd ask first

scottrepreneur:
Would love to batch these lookups. Anything we can do to get this unblocked?

jxom:
Wonder if this PR is still relevant as Viem has [built-in multicall batching](https://viem.sh/docs/clients/public.html#eth-call-aggregation-via-multicall)? 

Since this PR is a draft anyway, going to close for now! If you think this PR might otherwise be meaningful, feel free to raise a discussion!

scottrepreneur:
Added one here in `wagmi`. Thanks
- https://github.com/wagmi-dev/wagmi/discussions/3192

PR #963, opened 2023-08-01
set fees to zero for zero base fee chains (e.g. anvil)

holic:
We use anvil locally with `--block-base-fee-per-gas 0`, because we often create arbitrary burner accounts and don't want to do all the plumbing to fund new accounts per local instance.

However, viem isn't accounting for this in its fee calculation, always sending 1.5 gwei unless you 1) fund the account or 2) manually set each transaction call to `{ maxFeePerGas: 0n, maxPriorityFeePerGas: 0n }`.

We'd like to avoid the former, and the latter would leave us with wrapping viem or our own contract abstraction or something, but would be nice to keep to plain ol' viem.

See related issues that were fixed in foundry/anvil:
- https://github.com/foundry-rs/foundry/issues/5161
- https://github.com/foundry-rs/foundry/pull/5163

---

## PR-Codex overview
This PR focuses on implementing EIP-1559 fees and handling zero base fee. 

### Detailed summary:
- Added logic to set fees to zero when running with zero base fee.
- Implemented a buffer of 1.2x on top of the base fee to account for fluctuations.
- Added tests for zero base fee and legacy fees.

### DIFF
--- a/src/utils/transaction/prepareRequest.test.ts
+++ b/src/utils/transaction/prepareRequest.test.ts
@@ -70,6 +70,42 @@ describe('prepareRequest', () => {
     `)
   })
 
+  test('zero base fee', async () => {
+    await setup()
+    await setNextBlockBaseFeePerGas(testClient, {
+      baseFeePerGas: 0n,
+    })
+
+    const {
+      maxFeePerGas,
+      nonce: _nonce,
+      ...rest
+    } = await prepareRequest(walletClient, {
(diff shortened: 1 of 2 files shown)


holic:
I tried to add a test for this but calling `setNextBlockBaseFeePerGas` to `0n` doesn't seem to be doing what I expect?

image

jxom:
Closing in favor of #1006

PR #989, opened 2023-08-06
Added webSocket reconnect functionality on error event

sb-saksham:
Added the webSocket reconnect functionality on the occurrence of an "error" event to handle the case where the user's internet connection is unstable.
The user can provide different parameters to control the reconnection and also an option to disable the functionality altogether.
The options are:
- `reconnect : boolean ` the default value is **true** so the webSocket will reconnect automatically but if the user wants he can disable it by setting `reconnect=false`
- `reconnectAttempt : number` the default is **5**, user can provide a number higher or lower.
- `reconnectTimeout : number` the default is **5000ms**, the user can provide the timeout of when to attempt the subsequent reconnection attempt.

All changes are made to `getSocket` function so all the rest functionality is operational without any other tweaks.

---

## PR-Codex overview
This PR focuses on adding reconnection functionality to the `getSocket` function in `rpc.ts`.

### Detailed summary
- Added `reconnect`, `reconnectAttempts`, and `reconnectTimeout` parameters to the `getSocket` function.
- Added an `onError` function to handle reconnection attempts.
- Added an `error` event listener to the WebSocket to trigger reconnection on error.

### DIFF
--- a/src/utils/rpc.ts
+++ b/src/utils/rpc.ts
@@ -146,7 +146,11 @@ export type Socket = WebSocket & {
 
 const sockets = /*#__PURE__*/ new Map<string, Socket>()
 
-export async function getSocket(url_: string) {
+export async function getSocket(
+	url_: string, 
+	reconnect=true,
+	reconnectAttempts=5,
+	reconnectTimeout=5000) {
   const url = new URL(url_)
   const urlKey = url.toString()
 
@@ -177,7 +181,15 @@ export async function getSocket(url_: string) {
 
       // Set up a cache for subscriptions (eth_subscribe).
       const subscriptions = new Map<Id, CallbackFn>()
-
+      const onError = () => {
+	if (reconnect && reconnectAttempts > 0) {
+          // delay reconnection
+          setTimeout(() => {
+            reconnectAttempts -= 1
+            schedule().catch(console.error) // log error during reconnection
+          }, reconnectTimeout) // delay for 5 second default
+        }
+      }	
       const onMessage: (event: MessageEvent) => void = ({ data }) => {
         const message: RpcResponse = JSON.parse(data as string)
         const isSubscription = message.method === 'eth_subscription'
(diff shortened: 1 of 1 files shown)


src/utils/rpc.ts
```diff
@@ -146,7 +146,11 @@ export type Socket = WebSocket & {
 
 const sockets = /*#__PURE__*/ new Map<string, Socket>()
 
-export async function getSocket(url_: string) {
+export async function getSocket(
+	url_: string, 
+	reconnect=true,
+	reconnectAttempts=5,
+	reconnectTimeout=5000) {
```
holic, line 153: should these args be an object?

src/utils/rpc.ts
```diff
@@ -177,7 +181,15 @@ export async function getSocket(url_: string) {
 
       // Set up a cache for subscriptions (eth_subscribe).
       const subscriptions = new Map<Id, CallbackFn>()
-
+      const onError = () => {
+	if (reconnect && reconnectAttempts > 0) {
+          // delay reconnection
+          setTimeout(() => {
+            reconnectAttempts -= 1
+            schedule().catch(console.error) // log error during reconnection
```
holic, line 189: should we reset the `reconnectAttempts` after successful connection, so that this doesn't just die after 5 different instances of successfully reconnecting?

0xRapid: @sb-saksham

jxom:
Closing for now as this PR is becoming stale. Please re-open once @holic's comments are addressed! Thanks.

src/utils/rpc.ts
ITZSHOAIB: Reolved in [wevm:main](https://github.com/wevm/viem/tree/main)

src/utils/rpc.ts
ITZSHOAIB: Resolved in https://github.com/wevm/viem/pull/1894

PR #991, opened 2023-08-06
Fix estimateContractGas' error docsPath

norswap:


tmm (commented):
Can you add a changeset with `pnpm changeset`? Then looks good.

norswap:
> Can you add a changeset with `pnpm changeset`? Then looks good.

Done 👍

jxom:
Will also need to update snapshots in the test file. Can do this when I get to a computer if you don’t want to go through the ceremony of setting up your environment.

PR #1033, opened 2023-08-16
fix: ensure that getEnsAddress returns the correct padded address length

### LINKED ISSUE #1032: bug: getEnsAddress returns invalid address (when returned address starts with 0s)

### Package Version

1.6.0

### Current Behavior

It looks like getEnsAddress returns address with 0s to the left being trimmed when the returned address starts with 0s. This throws `InvalidAddressError` errors when you pass it on to contracts 

Example: When you have an address of `0x00A59Ec1F4BF9718EeE07078141b540272BAB807`, it returns `0xA59Ec1F4BF9718EeE07078141b540272BAB807` instead.

### Expected Behavior

The returned address string should be complete and in a format that contracts will accept it as address arguments.

### Steps To Reproduce

```js
const publicClient = createPublicClient({
  chain: mainnet,
  transport: http(),
});

const address = await publicClient.getEnsAddress({ name: 'skeith.eth'});
console.log({ address }); // Prints { address: '0xA59Ec1F4BF9718EeE07078141b540272BAB807' }
```

https://stackblitz.com/edit/viem-getting-started-hezseq?file=index.ts

### Anything else?

Tracing back, the last correct version was 1.3.0.
Versions 1.4.0, 1.5.0, and 1.6.0 all return the invalid address.

skeithc:
## Introduction
Because the trim function is used on the value being returned by `getEnsAddress`, the returned address may sometimes become too short to be a valid Ethereum address if the address has 0s at the beginning. 

## Changes
This change adds a pad to the address returned by getEnsAddress() to ensure that it is always the correct length. It also adds a test for an ens name that resolves to an address that has 0s at the beginning.

## Related
Resolves #1032

---

### Focus of the PR:
Fixing the `getEnsAddress` function to handle addresses that start with `0`.

### Detailed summary:
- Fixed `getEnsAddress` for addresses that start with `0`.
- Updated the test to get the address for a name that starts with `0s`.
- Removed unnecessary code in `getEnsAddress.ts`.

### DIFF
--- a/.changeset/tame-kids-poke.md
+++ b/.changeset/tame-kids-poke.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `getEnsAddress` for addresses that start with `0`.
--- a/src/actions/ens/getEnsAddress.test.ts
+++ b/src/actions/ens/getEnsAddress.test.ts
@@ -25,6 +25,14 @@ test('gets address for name', async () => {
   )
 })
 
+test('gets address that starts with 0s for name', async () => {
+  await expect(
+    getEnsAddress(publicClient, { name: 'skeith.eth' }),
+  ).resolves.toMatchInlineSnapshot(
+    '"0x00A59Ec1F4BF9718EeE07078141b540272BAB807"',
+  )
+})
+
 test('gets address for name with coinType', async () => {
   await expect(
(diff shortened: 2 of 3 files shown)


jxom:
Not sure why we were trimming `decodeFunctionResult`... The address gets trimmed internally.

PR #1045, opened 2023-08-21
fix: In the case of WalletConnect, add logic to handle when a user reject transaciton

siosio34:
hello. After connecting metamask through walletconnect, sending a transaction, and then canceling the transaction directly, UserRejectedRequestError was expected, but UnknownRpcError occurs as shown in the picture below.

image

After figuring out the cause of the error, there was an issue where the error code was 5000 instead of 4001 when walletconnect rejected the transaction.
https://docs.walletconnect.com/2.0/specs/clients/sign/error-codes#rejected-caip-25

Referring to the code below, when a user reject error occurs in the walletconnect environment, a UserRejectedRequestError is raised.
https://github.com/wagmi-dev/references/blob/849d7a5d2c630c1b55f3414cea034303e84250f9/packages/connectors/src/walletConnect.ts#L159C32-L159C32

I don't know the project structure well, so it may be the direction of modification that I don't want. Feel free to edit this pull request at any time.

Thank you for making such a great project and saving me a lot of time. Thank you every time.

---

## PR-Codex overview
This PR focuses on handling a specific error scenario when the user rejects a transaction after connecting the wallet through WalletConnect. 

### Detailed summary
- Added import for `UserRejectedRequestError` and `ProviderRpcError`
- Added condition to check for user rejection error and return `UserRejectedRequestError`

### DIFF
--- a/src/utils/errors/getTransactionError.ts
+++ b/src/utils/errors/getTransactionError.ts
@@ -2,6 +2,7 @@ import type { Account } from '../../accounts/types.js'
 import type { SendTransactionParameters } from '../../actions/wallet/sendTransaction.js'
 import type { BaseError } from '../../errors/base.js'
 import { TransactionExecutionError } from '../../errors/transaction.js'
+import { UserRejectedRequestError, type ProviderRpcError } from '../../index.js'
 import type { Chain } from '../../types/chain.js'
 
 import {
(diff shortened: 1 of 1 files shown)


jxom:
Might be worth adding a case here instead for `5000`: https://github.com/wagmi-dev/viem/blob/169a695edeb7a0d1b6a9885509c3da2683cfe7ba/src/utils/buildRequest.ts#L111

siosio34:
@jxom  sorry. Could you please explain in more detail what you want?

jxom:
Ended up resolving via #1136! Thanks for the PR.

PR #1081, opened 2023-08-29
fix: multicall perfomance

witem:
It looks like `viem` has big performance degradation because of calling `chunkedCalls.flat()` for each item from `results` in the `multicall` function. Based on the code I think we can move this `flat` out of the `map` cycle, which will significantly improve performance, especially for big multicalls.

Test scenario around 2k `balanceOf` with `batchSize: 30 * 1024`:
Flame chart before:
image
Flame chart after: 
image

---

## PR-Codex overview
This PR focuses on improving multicall performance. 

### Detailed summary
- Refactored code to remove redundant variable declaration.

### DIFF
--- a/.changeset/pink-beers-share.md
+++ b/.changeset/pink-beers-share.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-multicall performance improve
+Improved multicall performance.
--- a/.changeset/pink-beers-share.md
+++ b/.changeset/pink-beers-share.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+multicall performance improve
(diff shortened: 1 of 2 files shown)


.changeset/pink-beers-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"viem": patch
+---
+
+multicall performance improve
```
tmm, line 5:
```suggestion
Improved multicall performance.
```

tmm (approved):

.changeset/pink-beers-share.md
witem: updated, thanks!

PR #1096, opened 2023-08-31
Correct the RPC error handling

MOZGIII:
The error code used to detect revert is set to `3`, and it not documented where this value comes from. This does not look right, the real code should be `-32603` as per https://eips.ethereum.org/EIPS/eip-1474#error-codes.

This is a wip, published to collect feedback.

A few concerns I have here:
- we can't just use the new code, as people relying on the old code will suffer if we just drop it; checking both seems harmless though.
- this is a breaking change for those who expected the error to be `ContractFunctionExecutionError`  - after this change they may get `ContractFunctionRevertedError` where the didn't expect.
- why are there some fields that are in `ContractFunctionExecutionError` but not in `ContractFunctionRevertedError`? why is not data available at `ContractFunctionExecutionError` if it was passed but not parsed, but other params are copied there?

CC: @tonycdot

---

### Detailed summary

This PR introduces a new constant `EXECUTION_REVERTED_ERROR_CODE_NEW` and updates the `getContractError` function to handle the new error code. The changes include:
- Adding a new constant `EXECUTION_REVERTED_ERROR_CODE_NEW`
- Updating the condition in `getContractError` to include the new error code

### DIFF
--- a/src/utils/errors/getContractError.ts
+++ b/src/utils/errors/getContractError.ts
@@ -10,6 +10,7 @@ import {
 } from '../../errors/contract.js'
 
 const EXECUTION_REVERTED_ERROR_CODE = 3
+const EXECUTION_REVERTED_ERROR_CODE_NEW = -32603
 
 export function getContractError(
   err: BaseError,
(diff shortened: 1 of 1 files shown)


jxom:
The error code `3` comes from most execution clients (geth, erigon, reth) to indicate an execution error. 

What RPC Provider throws a `-32603` so I can test?

MOZGIII:
> The error code `3` comes from most execution clients (geth, erigon, reth) to indicate an execution error.
> 
> What RPC Provider throws a `-32603` so I can test?

I see, this is supposed to be the legacy behavior before the EIP-1474, as prior to that codes were not specified, and clients used to set them as they please.

---

Thinking about the error model, it might make sense to ignore the code altogether, and attempt parsing data if it is present regardless of the code. The code itself can be exposed for the application level inspection if needed - but most apps will only need the message/custom data.

jxom:
This should be okay for the time being! We can refactor at a later point.

PR #1088, opened 2023-08-30
Add OP Stack contracts to ChainConstants.contracts

wilsoncusack:
This will unlock improved APIs and typing in [op-viem](https://github.com/base-org/op-viem).

---

## PR-Codex overview
This PR focuses on adding OP Stack contracts to chain contracts type. 

### Detailed summary
- Added OP Stack contracts to chain contracts type in `src/types/chain.ts`
- Added contracts for L1 OP Stack and L2 OP Stack
- Added addresses for L2 OP Stack contracts

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #1110, opened 2023-09-01
feat: amend chain type to support OP Stack contracts

jxom:
See https://github.com/wagmi-dev/viem/pull/1088.

This PR enables downstream consumers of viem (ie. `op-viem`) to add their own contracts onto chains.

---

## PR-Codex overview
This PR focuses on amending the `Chain` type to allow arbitrary `contracts` that can be dependent on the chain.

### Detailed summary
- Amended `Chain` type to allow arbitrary `contracts` (as well as contracts that could be dependent on chain).

### DIFF
--- a/src/types/chain.ts
+++ b/src/types/chain.ts
@@ -34,6 +34,8 @@ export type ChainConstants = {
   }
   /** Collection of contracts */
   contracts?: {
+    [key: string]: ChainContract | { [chainId: number]: ChainContract }
+  } & {
     ensRegistry?: ChainContract
     ensUniversalResolver?: ChainContract
     multicall3?: ChainContract
--- a/.changeset/calm-terms-bathe.md
+++ b/.changeset/calm-terms-bathe.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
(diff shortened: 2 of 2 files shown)


wilsoncusack:
This will help. Also might be nice if contracts wasn't optional. Minimally an empty dict? 🤷

src/types/chain.ts
```diff
@@ -34,6 +34,8 @@ export type ChainConstants = {
   }
   /** Collection of contracts */
   contracts?: {
+    [key: string]: ChainContract | { [chainId: number]: ChainContract }
+  } & {
```
tmm, line 38: Should be like this, right?

```suggestion
    [key: string]: ChainContract | { [chainId: number]: { [key: string]: ChainContract } }
  } & {
```

jxom: The named part is already at L37:5-17

wilsoncusack: Yeah the format is {"optimismPortal": {8543: "0x..."}}

I'm not sure we need the "deployed at" info but I guess good to stay consistent.

wilsoncusack: Ofc we _could_ also do it as chainId>name>address, instead

PR #1188, opened 2023-09-12
fix: support integer for `baseFeeMultiplier`

Holybasil:
I set `baseFeeMultiplier` to `2` for the chain.
However, got the error
```
TypeError: Cannot read properties of undefined (reading 'length')
    at internal_estimateFeesPerGas (bundle.js:112518:62)
```

---

### Detailed summary
- The focus of this PR is to handle the case when `baseFeeMultiplier` is an integer without decimals.
- Notable changes:
  - Added a check to see if `baseFeeMultiplier` is an integer.
  - If `baseFeeMultiplier` is an integer, the `multiply` function will return the result of `base` multiplied by `baseFeeMultiplier`.
  - If `baseFeeMultiplier` has decimals, the `multiply` function will calculate the result using the previous logic.

### DIFF
--- a/src/actions/public/estimateFeesPerGas.ts
+++ b/src/actions/public/estimateFeesPerGas.ts
@@ -106,10 +106,17 @@ export async function internal_estimateFeesPerGas<
   })()
   if (baseFeeMultiplier < 1) throw new BaseFeeScalarError()
 
-  const decimals = baseFeeMultiplier.toString().split('.')[1].length
-  const denominator = 10 ** decimals
-  const multiply = (base: bigint) =>
-    (base * BigInt(baseFeeMultiplier * denominator)) / BigInt(denominator)
+  const multiply = (base: bigint) => {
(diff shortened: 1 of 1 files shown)


Holybasil:
In addition, the decimal calculations here still suffer from precision issues.
When I set `baseFeeMultiplier` to `2.01`, the error is
```
RangeError: The number 200.99999999999997 cannot be converted to a BigInt because it is not an integer
```
This is because
```js
2.01 * 100 // = 200.99999999999997 not 201
```
How about importing [decimal.js](https://github.com/MikeMcl/decimal.js) to solve the problem?

jxom:
Thanks for the PR! Fixed in #1194 – I have added you as a co-author so you are still credited for the contribution. :)

Holybasil:
I think `Math.round` is more suitable than `Math.ceil`.

```js
1.09 * 100 //  = 109.00000000000001
```
@jxom

PR #1214, opened 2023-09-17
docs: remove duplicate sidebar item

Sabnock01:
### Detailed summary

This PR removes the `sendRawTransaction` entry from the sidebar in `site/.vitepress/sidebar.ts`.

- Removed `sendRawTransaction` entry from the sidebar in `site/.vitepress/sidebar.ts`.

### DIFF
--- a/site/.vitepress/sidebar.ts
+++ b/site/.vitepress/sidebar.ts
@@ -188,10 +188,6 @@ export const sidebar: DefaultTheme.Sidebar = {
               link: '/docs/actions/public/getTransactionReceipt',
             },
             {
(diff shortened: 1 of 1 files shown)


jxom:
This was actually intentional (you can use it on both public and wallet client)

Sabnock01:
> This was actually intentional (you can use it on both public and wallet client)

Ah, would you prefer it be added back and simply point to the other page? Or just have a duplicate page? (as opposed to throwing a 404?)

jxom:
Didn’t realise it was 404ing! Yes, pointing to the wallet one sounds good!

PR #1215, opened 2023-09-17
perf/refactor: rlp

jxom:
This PR aims to refactor our RLP coding implementation to be _a lot_ faster by using cursors instead of concating arrays (ie. copies).

h/t to @fubhy for the `Object.create` perf boost in `cursor.ts`.

Screenshot 2023-09-18 at 8 59 41 am

Screenshot 2023-09-18 at 9 01 19 am

---

### Focus of this PR:
Refactoring RLP coding for improved performance and adding test vectors.

### Detailed summary:
- Refactored RLP coding for better performance.
- Added test vectors for RLP and transaction encoding/decoding.
- Added functions for reading gzipped JSON files.
- Added error classes for encoding and cursor operations.
- Updated package.json scripts for running vectors tests.
- Updated size-limit configuration.
- Updated gitignore and workflow files.

> The following files were skipped due to too many changes: `src/utils/encoding/toBytes.test.ts`, `src/utils/encoding/toHex.ts`, `vectors/src/transaction.vectors.test.ts`, `vectors/src/rlp.ts`, `src/utils/cursor.ts`, `src/utils/encoding/toRlp.ts`, `src/utils/encoding/toRlp.bench.ts`, `src/utils/encoding/fromRlp.ts`, `vectors/src/transaction.ts`, `src/utils/cursor.test.ts`, `src/utils/encoding/fromRlp.bench.ts`, `src/utils/encoding/fromRlp.test.ts`, `src/utils/encoding/toRlp.test.ts`, `src/utils/encoding/__snapshots__/toRlp.test.ts.snap`


src/utils/encoding/toBytes.bench.ts
```diff
@@ -2,13 +2,22 @@ import { toBeArray } from 'ethers'
 import { bench, describe } from 'vitest'
 
 import { hexToBytes } from './toBytes.js'
+import { bytesToHex } from './toHex.js'
+
+const generateBytes = (length: number) => {
```
fubhy, line 7: You could consider using https://www.npmjs.com/package/fast-check for your generating, fuzzing and bench input scenarios.

src/utils/encoding/toBytes.ts
```diff
@@ -117,6 +117,12 @@ export function hexToBytes(hex_: Hex, opts: HexToBytesOpts = {}): ByteArray {
   let hexString = hex.slice(2) as string
   if (hexString.length % 2) hexString = `0${hexString}`
 
+  if ('Buffer' in globalThis && typeof Buffer.from === 'function') {
+    if (!isHex(hex_))
+      throw new BaseError(`${hexString} is not a valid hex value.`)
+    return Uint8Array.from(Buffer.from(hexString, 'hex'))
+  }
```
fubhy, line 124: This is good. As discussed, given wrongly encoded input to `Buffer.from` doesn't fail as it should (imho). So I appreciate that `isHex` check before that call for correctness.

To be sure that we are decoding correctly and the output between Buffer.from and the custom implementation is identical I'd suggest to add a test that fuzzes both. But that would require splitting it up here.

src/utils/encoding/toHex.ts
```diff
@@ -114,10 +114,20 @@ export type BytesToHexOpts = {
...
+        value.byteOffset,
+        value.byteLength,
+      ).toString('hex')
+
+    let string = ''
+    for (let i = 0; i < value.length; i++) {
+      string += hexes[value[i]]
+    }
+    return string
+  })()
```
fubhy, line 130: This function can be externalized. Safe some memory & performance.

Also, I'd put the check for (`'Buffer' in globalThis && typeof Buffer.from === 'function'`) into a const somewhere and use that in these two functions (and wherever else it might be used). Checking a boolean is always faster :)

src/utils/encoding/toRlp.ts
```diff
@@ -1,49 +1,118 @@
...
+        cursor.pushByte(0xc0 + bodyLength)
+      } else {
+        cursor.pushByte(0xc0 + 55 + sizeOfBodyLength)
+        if (sizeOfBodyLength === 1) cursor.pushUint8(bodyLength)
+        else if (sizeOfBodyLength === 2) cursor.pushUint16(bodyLength)
+        else if (sizeOfBodyLength === 3) cursor.pushUint24(bodyLength)
+        else cursor.pushUint32(bodyLength)
+      }
+      list.forEach((x) => x.encode(cursor))
+    },
```
fubhy, line 77: This could also use `Object.create`

src/utils/encoding/toRlp.ts
```diff
@@ -1,49 +1,118 @@
...
+      } else {
+        cursor.pushByte(0x80 + 55 + sizeOfBytesLength)
+        if (sizeOfBytesLength === 1) cursor.pushUint8(bytes.length)
+        else if (sizeOfBytesLength === 2) cursor.pushUint16(bytes.length)
+        else if (sizeOfBytesLength === 3) cursor.pushUint24(bytes.length)
+        else cursor.pushUint32(bytes.length)
+        cursor.pushBytes(bytes)
+      }
+    },
+  }
```
fubhy, line 109: Same here.

vectors/src/rlp.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const generateList = (length: number) => {
+  const bytes: `0x${string}`[] = []
+  for (let i = 0; i < length; i++)
+    bytes.push(bytesToHex(generateBytes(Math.floor(Math.random() * 128))))
+  return bytes
+}
+
+export async function generateRlpVectors() {
```
fubhy, line 19: Check out https://www.npmjs.com/package/fast-check

fubhy (changes requested):

src/utils/encoding/toBytes.ts
fubhy: You could also consider doing sth. like

```ts
export const hexToBytes: (hex_: Hex, opts: HexToBytesOpts = {}): ByteArray = hasBuffer ? hexToBytesBuffer : hexToBytesCustom;

paulmillr:
Could you add benchmarks against community-standard https://www.npmjs.com/package/@ethereumjs/rlp ?

jxom:
@paulmillr – good idea, can do!

jxom:
Seems pretty on-par for encoding:

Screenshot 2023-09-24 at 2 08 47 pm

Maybe some room for improvement for decoding:

Screenshot 2023-09-24 at 2 09 31 pm

These benches are for `Uint8Array` → `Uint8Array`.

Could be worth contributing back to ethereumjs here – can look more closely when I find some time.

paulmillr:
Thanks, that sounds interesting!

paulmillr:
@jxom How does this `Buffer` addition work with bundlers? In my experience some bundlers could see this and auto-include buffer polyfills.

I would strongly recommend against buffers, because they have security issues.

jxom:
Fair call. Don't think the perf gain is much of an improvement after your change. Will remove.

PR #1249, opened 2023-09-24
Speed-up hexToBytes 6x.

paulmillr:
## PR-Codex overview
This PR focuses on refactoring the `hexToBytes` function to improve its performance. 

### Detailed summary
- Refactored `hexToBytes` to be up to 6x more performant.
- Introduced a `charCodeMap` object to map character codes to their base16 values.
- Implemented the `charCodeToBase16` function to convert character codes to base16 values.
- Replaced the previous implementation of `hexToBytes` with a more optimized approach using the `charCodeToBase16` function.
- Improved error handling for invalid byte sequences.


src/utils/encoding/toBytes.ts
```diff
@@ -88,6 +88,25 @@ export type HexToBytesOpts = {
...
 
+// We use very optimized technique to convert hex string to byte array
+enum HexC {
+  ZERO = 48, // 0
+  NINE = 57, // 9
+  A_UP = 65, // A
+  F_UP = 70, // F
+  A_LO = 97, // a
+  F_LO = 102, // f
+}
```
tmm, line 99: Can we use a const asserted object here? Try to stay away for enums.

paulmillr: it would be slow. probably much slower.

paulmillr: actually i've used `const enum` which is automatically compiled to its values, not sure what happened, linter maybe

paulmillr: hmmm, interesting, changing to object doesn't impact perf much. I wonder why is this

src/utils/encoding/toBytes.ts
```diff
@@ -88,6 +88,25 @@ export type HexToBytesOpts = {
...
+const HexC = {
+  ZERO: 48, // 0
+  NINE: 57, // 9
+  A_UP: 65, // A
+  F_UP: 70, // F
+  A_LO: 97, // a
+  F_LO: 102, // f
+} as const
+
+function charCodeToBase16(char: number) {
```
jxom, line 101: I love these optimizations so much lol

PR #1058, opened 2023-08-24
feat: `signTransaction`, `sendRawTransaction`, `prepareTransactionRequest`

jxom:
Added [`prepareTransactionRequest`](https://viem-6c4wqz00t-wagmi-dev.vercel.app/docs/actions/wallet/prepareTransactionRequest.html), [`signTransaction`](https://viem-6c4wqz00t-wagmi-dev.vercel.app/docs/actions/wallet/signTransaction.html), and [`sendRawTransaction`](https://viem-6c4wqz00t-wagmi-dev.vercel.app/docs/actions/wallet/sendRawTransaction.html).

Ref: https://github.com/wagmi-dev/viem/discussions/785

---

### Focus of this PR:
Adding new functions `prepareTransactionRequest`, `signTransaction`, and `sendRawTransaction` to the codebase.

### Detailed summary:
- Added `prepareTransactionRequest`, `signTransaction`, and `sendRawTransaction` functions.
- Updated imports and types in multiple files.
- Deprecated `prepareRequest` function and replaced it with `prepareTransactionRequest` in exports.
- Updated documentation and sidebar links.

> The following files were skipped due to too many changes: `src/actions/public/getBlock.test.ts`, `src/index.ts`, `src/types/chain.ts`, `src/actions/public/watchContractEvent.ts`, `src/actions/wallet/sendTransaction.ts`, `src/clients/decorators/public.test.ts`, `src/actions/public/estimateFeesPerGas.ts`, `site/docs/actions/wallet/sendRawTransaction.md`, `src/actions/wallet/sendRawTransaction.ts`, `src/clients/createPublicClient.test.ts`, `src/clients/decorators/wallet.test.ts`, `src/chains/celo/formatters.ts`, `src/utils/formatters/transactionRequest.test.ts`, `src/clients/createWalletClient.test.ts`, `src/clients/decorators/public.ts`, `src/chains/celo/types.ts`, `src/actions/wallet/signTransaction.ts`, `src/chains/celo/formatters.test-d.ts`, `src/clients/decorators/wallet.ts`, `src/utils/transaction/prepareRequest.ts`, `site/docs/actions/wallet/signTransaction.md`, `site/docs/actions/wallet/prepareTransactionRequest.md`, `src/actions/wallet/signTransaction.test.ts`, `src/utils/transaction/prepareRequest.test.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #1044, opened 2023-08-21
feat: `estimateFeesPerGas`, `estimateMaxPriorityFeePerGas`

jxom:
- Added `estimateFeesPerGas` & `estimateMaxPriorityFeesPerGas` Actions.
- Added `estimateFeesPerGas` & `baseFeeMultiplier` to the Chain Fees configuration (`chain.fees`).

Docs: 

- [estimateFeesPerGas](https://viem-a4sw0wg1f-wagmi-dev.vercel.app/docs/actions/public/estimateFeesPerGas.html)
- [estimateMaxPriorityFeePerGas](https://viem-a4sw0wg1f-wagmi-dev.vercel.app/docs/actions/public/estimateMaxPriorityFeePerGas.html)
- [Chain Configuration](https://viem-a4sw0wg1f-wagmi-dev.vercel.app/docs/clients/chains.html#chain-configuration)

---

### Focus of this PR:
This PR focuses on adding new actions and types related to fee estimation and configuration.

### Detailed summary:
- Added `estimateFeesPerGas` and `estimateMaxPriorityFeesPerGas` actions.
- Added `FeeValuesType` type for fee values.
- Added `estimateFeesPerGas` and `baseFeeMultiplier` to the Chain Fees configuration.
- Added new errors: `BaseFeeScalarError`, `Eip1559FeesNotSupportedError`, and `MaxFeePerGasTooLowError`.
- Updated documentation and tests related to fee estimation and configuration.

> The following files were skipped due to too many changes: `src/clients/createPublicClient.test.ts`, `site/docs/actions/public/estimateMaxPriorityFeePerGas.md`, `src/actions/wallet/sendTransaction.test.ts`, `site/docs/actions/public/estimateFeesPerGas.md`, `src/index.ts`, `src/actions/public/estimateMaxPriorityFeePerGas.ts`, `src/actions/public/estimateMaxPriorityFeePerGas.test.ts`, `src/clients/decorators/public.ts`, `src/utils/transaction/prepareRequest.test.ts`, `src/utils/transaction/prepareRequest.ts`, `src/actions/public/estimateFeesPerGas.ts`, `src/types/chain.ts`, `src/actions/public/estimateFeesPerGas.test.ts`, `site/docs/clients/chains.md`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #1006, opened 2023-08-09
feat: default priority fee per-chain

jxom:
Added `fees` to `chain` config that includes a `getDefaultPriorityFee` for setting a default priority fee for a chain.

```ts
import type { Chain } from 'viem'

export const example = {
  // ...
  fees: {
    getDefaultPriorityFee: () => 1_000_000n, // 0.001 gwei
  },
  // ...
} as const satifies Chain
```

Also cleaned up the chain serializers & formatters types. Probably makes more sense to have them in `types/chain.ts` rather than separate files since they aren't related to anything else apart from the chain. Will also help if we scale to more than just `fees`, `serializers` & `formatters` on the chain config (and also when we chose to move the chains into a separate package).

---

### Focus of this PR:
This PR focuses on adding a `fees` field to the `chain` configuration, which includes a `defaultPriorityFee` for setting a default priority fee (`maxFeePerGas`) for a chain.

### Detailed summary:
- Added `fees` field to `chain` config
- `fees` field includes a `defaultPriorityFee` for setting a default priority fee (`maxFeePerGas`) for a chain
- Updated multiple files to include the `fees` field in the `chain` configuration

> The following files were skipped due to too many changes: `src/utils/formatters/block.ts`, `src/index.ts`, `src/actions/wallet/sendTransaction.ts`, `src/actions/wallet/writeContract.ts`, `src/chains/index.ts`, `src/types/chain.ts`, `src/utils/transaction/prepareRequest.ts`, `src/utils/transaction/prepareRequest.test.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

PR #1280, opened 2023-10-03
fix: zero base fee

holic:
I just bumped viem from 1.6.0 to 1.14.0 and I am seeing "intrinsic gas too high" when using `writeContract` on an anvil chain running with `--base-fee 0` and [viem chain set to `fees: { defaultPriorityFee: 0n }`](https://github.com/latticexyz/mud/blob/f99e889872e6881bf32bcb9a605b8b5c1b05fac4/packages/common/src/chains/mudFoundry.ts#L7).

Looking at request logs, I can see a new call to `eth_maxPriorityFeePerGas` before `eth_estimateGas` that returns `0x3b9aca00`, which is used as the value for `maxFeePerGas` and `maxPriorityFeePerGas` in the `eth_estimateGas` call.

Before the version bump, there was no `eth_maxPriorityFeePerGas` call and the `0n` from the chain config's `defaultPriorityFee` was used as the value of `maxFeePerGas` and `maxPriorityFeePerGas`.

I am not sure if I am misunderstanding the new fee calculation but this seems like a regression from the previous behavior that respected the chain config's `defaultPriorityFee`? I think this behavior changed in https://github.com/wagmi-dev/viem/pull/1058 but haven't read it thoroughly.

I've added a test for this here, but can't seem to get tests running locally to verify that it fails + add a fix.

---

### Focus of the PR:
This PR focuses on improving the handling of defaultPriorityFee in the code.

### Detailed summary:
- Added a check for typeof chain?.fees?.defaultPriorityFee to handle undefined values.
- Added tests for chain override with zero base fee and async zero base fee.
- Updated package.json to include typesVersions for better type definitions.

### DIFF
--- a/src/actions/public/estimateMaxPriorityFeePerGas.test.ts
+++ b/src/actions/public/estimateMaxPriorityFeePerGas.test.ts
@@ -154,7 +154,7 @@ test('client: chain `defaultPriorityFee` override', async () => {
     transport: http(),
   })
   expect(await estimateMaxPriorityFeePerGas(client_4)).toBe(0n)
-  
+
   // async zero base fee
   const client_5 = createPublicClient({
     chain: {
--- a/src/actions/public/estimateMaxPriorityFeePerGas.test.ts
+++ b/src/actions/public/estimateMaxPriorityFeePerGas.test.ts
@@ -74,6 +74,36 @@ test('args: chain `defaultPriorityFee` override', async () => {
       },
     }),
   ).toBe(69420n)
+
+  // zero base fee
+  const client_4 = createPublicClient({
+    transport: http(localHttpUrl),
+  })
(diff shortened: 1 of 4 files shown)


holic:
I think I found it. We had an `typeof defaultPrioritFee !== 'undefined'` check in https://github.com/wagmi-dev/viem/pull/1006/files#diff-e055abee919d609c1a367c63ac0c8877d34a82f21e687f8901db46bb7eb4a18bR90 but that changed to a falsey check in https://github.com/wagmi-dev/viem/pull/1044/files#diff-cb3ad47420815660f8a1c71a9587607f6ad26342fd4220688031852b1d327c7dR71

jxom (approved):

PR #1309, opened 2023-10-07
feat: add chain url avaliable check

izayl:
### Focus of the PR:
This PR focuses on adding chain checking functionality and updating configuration files.

### Detailed summary:
- Added `tsconfig.json` to include the `test/src` directory for local development and type checking.
- Added `test/tsconfig.json` with compiler options for `bun-types` and path mapping.
- Updated `src/utils/promise/withTimeout.ts` to make the `errorInstance` parameter optional.
- Added `github/workflows/chaincheck.yml` for checking chains in the CI workflow.
- Updated `package.json` to include a `chaincheck` script.
- Added `test/chains/check.test.ts` for testing chain URLs and block explorers.


izayl:
seems not suitable to check it on Actions, as it has maximum only 5 min runtime

jxom (commented):
Looks good! Instead of running this in a Bun script, do you think we could just run them as Bun tests? 

This means we can get rid of the `for` loop, console logs, etc.

izayl:
looks pretty
image

izayl:
should we use snapshot test, to make sure the new added chain is available

PR #1339, opened 2023-10-12
fix: `decodeEventLogs` strict mode

### LINKED ISSUE #1336: decodeEventLog in strict: true mode fails to enforce ABI conformity on indexed arguments

### Package Version

^1.15.1

### Current Behavior

I've encountered an issue where `decodeEventLog` when used with `strict: true` is not enforcing the ABI conformity regarding the number of indexed arguments. According to the documentation, strict: true should throw an error when there's a mismatch between the number of indexed/non-indexed arguments and the data length. However, it seems to be allowing partial decoding without throwing an error when there's a mismatch, especially concerning the indexed argument `nounId` in the `AuctionCreated` event.

`decodeEventLog` with `strict: true` allows partial decoding of the `AuctionCreated` event, missing the indexed nounId argument, without throwing an error.

### Expected Behavior

`decodeEventLog` with `strict: true` should either throw an error due to a mismatch or correctly decode all arguments, including indexed arguments like nounId.

### Steps To Reproduce

Use decodeEventLog with strict: true to decode the AuctionCreated event from the following transaction: [Etherscan Transaction](https://etherscan.io/tx/0xb4b02bd9f1df0fbddd91a694b9b20b7656656a713f216dec6bdfb474e4d41fb2#eventlog)

Observe that nounId is missing from the decoded event, and no error is thrown for the mismatch.

`const decoded = decodeEventLog({
  abi,
  data,
  strict: true,
  eventName,
  topics,
});

// Output missing `nounId`, no error thrown
`

_No response_

### Anything else?

The documentation mentions that setting strict: false should allow for partial decoding, but strict: true should enforce strict ABI conformity and throw an error on mismatch. The current behavior deviates from this expectation.

jxom:
Fixes #1336.

---

### Detailed summary
- Fixed topics mismatch in `decodeEventLogs` strict mode.

### DIFF
--- a/.changeset/plenty-meals-visit.md
+++ b/.changeset/plenty-meals-visit.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed topics mismatch in `decodeEventLogs` strict mode.
--- a/src/actions/public/getFilterChanges.test.ts
+++ b/src/actions/public/getFilterChanges.test.ts
@@ -356,7 +356,7 @@ describe('contract events', () => {
         'Transfer'
       >[]
     >(logs)
-    expect(logs.length).toBe(784)
+    expect(logs.length).toBe(783)
   })
 
   test('args: singular `from`', async () => {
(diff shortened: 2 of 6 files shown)


rocketmanh:
boom ty sir now time to rerun my indexer

PR #1379, opened 2023-10-23
feat: add support for celo cip 64 transaction type

yerdua:
This PR adds support for the new [CIP-64](https://github.com/celo-org/celo-proposals/blob/master/CIPs/cip-0064.md) transaction type on Celo, which is similar to CIP-42, but with `gatewayFee` and `gatewayFeeRecipient` fields removed.

---

### Focus of this PR:
This PR focuses on adding support for Celo CIP-64 transactions.

### Detailed summary:
- Added support for Celo CIP-64 transactions
- Updated `src/chains/utils/index.ts` to include `RpcTransactionCIP64`, `RpcTransactionRequestCIP64`, `TransactionCIP64`, `TransactionRequestCIP64`, `TransactionSerializableCIP64`, and `TransactionSerializedCIP64`
- Updated `src/chains/celo/formatters.test.ts` to include tests for CIP-64 transactions
- Updated `src/chains/celo/formatters.test-d.ts` to include type assertions for CIP-64 transactions
- Updated `src/chains/celo/formatters.ts` to include formatters for CIP-64 transactions
- Updated `src/chains/celo/parsers.ts` to include parsers for CIP-64 transactions
- Updated `src/chains/celo/types.ts` to include types for CIP-64 transactions

> The following files were skipped due to too many changes: `src/chains/celo/types.ts`, `src/chains/celo/serializers.ts`, `src/chains/celo/parsers.test.ts`, `src/chains/celo/serializers.test.ts`


src/chains/celo/formatters.test.ts
```diff
@@ -237,30 +237,30 @@ describe('block', () => {
...
+      {
+        "blockHash": "0xac8c9bc3b84e103dc321bbe83b670e425ff68bfc9a333a4f1b1b204ad11c583d",
+        "blockNumber": 16645775n,
+        "chainId": undefined,
+        "ethCompatible": false,
+        "feeCurrency": null,
+        "from": "0x045d685d23e8aa34dc408a66fb408f20dc84d785",
+        "gas": 1527520n,
+        "gasPrice": 562129081n,
+        "gatewayFee": "0x0",
```
yerdua, line 249: This changed from `"gatewayFee:": 0n` to `"gatewayFee": "0x0"` because of a change in src/chains/celo/formatters.ts.

I'm uncertain which value is expected and correct for legacy types.

src/chains/celo/formatters.ts
```diff
@@ -35,10 +61,15 @@ export const formattersCelo = {
...
-          gatewayFeeRecipient: transaction.gatewayFeeRecipient,
+
+          ...(transaction.type === '0x7c'
+            ? {
+                gatewayFee: transaction.gatewayFee
+                  ? hexToBigInt(transaction.gatewayFee)
+                  : null,
+                gatewayFeeRecipient: transaction.gatewayFeeRecipient || null,
+              }
+            : {}),
```
yerdua, line 72: Previously, the `gatewayFee` fields were always added to the return value for this function. Here, I've changed it to only include them if the type is 0x7c.

However, maybe this is incorrect, since other transaction types exist ('0x0', '0x1', '0x2')? On the other hand... maybe these aren't relevant in this function outside of tests?

Would love some guidance here on what the correct behavior should be.

aaronmgdr: I could see this go both ways.

src/chains/celo/formatters.test.ts
jxom: I'd expect this to be the same as before!

src/chains/celo/formatters.ts
jxom: I don't really know enough about Celo protocol to provide an answer here. Maybe we can leave it how it is now?

jxom (commented):
PR looks solid! Happy to merge once outstanding comments are addressed!

src/chains/celo/formatters.ts
yerdua: In recent changes, I restored the original behavior for other types.

jxom (approved):

PR #1382, opened 2023-10-23
feat: infer actions from extended client

jxom:
Added internal `getAction` utility which allows the ability for Actions (i.e. `readContract`) to infer their internal/dependant Actions (i.e. `call`) from the optionally extended Client.

For instance, if an extended Client has overridden the `call` Action, then the `readContract` Action will use that instead of Viem's internal `call` Action.

---

## PR-Codex overview
This PR focuses on adding the ability for Actions to infer their internal/dependent Actions from the extended Client. 

### Detailed summary
- Added ability for Actions to infer their internal/dependent Actions from the extended Client.
- Updated `getAction.ts` to include the new functionality.
- Updated multiple action files to use the `getAction` function instead of the previous implementation.

> The following files were skipped due to too many changes: `src/actions/public/estimateFeesPerGas.ts`, `src/actions/public/watchBlocks.ts`, `src/actions/wallet/prepareTransactionRequest.ts`, `src/actions/public/watchPendingTransactions.ts`, `src/actions/wallet/sendTransaction.ts`, `src/actions/public/estimateMaxPriorityFeePerGas.ts`, `src/utils/getAction.test.ts`, `src/actions/public/watchEvent.ts`, `src/actions/public/watchContractEvent.ts`, `src/actions/getContract.ts`, `src/actions/public/waitForTransactionReceipt.ts`, `src/clients/createClient.ts`, `src/clients/createClient.test-d.ts`

### DIFF
--- a/src/clients/createClient.ts
+++ b/src/clients/createClient.ts
@@ -1,8 +1,9 @@
 import type { Address } from 'abitype'
 
-import type { Account, JsonRpcAccount } from '../accounts/types.js'
+import type { JsonRpcAccount } from '../accounts/types.js'
 import type { ParseAccountErrorType } from '../accounts/utils/parseAccount.js'
 import type { ErrorType } from '../errors/utils.js'
+import type { Account } from '../types/account.js'
 import type { Chain } from '../types/chain.js'
 import type {
   EIP1193RequestFn,
@@ -12,6 +13,8 @@ import type {
 import type { Prettify } from '../types/utils.js'
 import { parseAccount } from '../utils/accounts.js'
 import { uid } from '../utils/uid.js'
+import type { PublicActions } from './decorators/public.js'
+import type { WalletActions } from './decorators/wallet.js'
 import type { Transport } from './transports/createTransport.js'
 
 export type ClientConfig<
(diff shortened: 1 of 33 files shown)


src/utils/getAction.ts
```diff
@@ -0,0 +1,13 @@
...
+export function getAction<params extends {}, returnType extends {}>(
+  client: Client,
+  action: (_: any, params: params) => returnType,
+) {
+  return (params: params): returnType =>
+    (
+      client as Client & {
+        [key: string]: (params: params) => returnType
+      }
+    )[action.name]?.(params) ?? action(client, params)
```
tmm, line 12: Do we want this to fall back to `action(client, params)` if `client[action.name](params)` result is nullish?

src/clients/createClient.ts
```diff
@@ -53,6 +56,34 @@ export type ClientConfig<
   type?: string | undefined
 }
 
+type ProtectedActions = Pick<
```
tmm, line 59: Comment here about what makes a protected action would be good.

src/utils/getAction.ts
jxom: Think `null` would be a valid response in a lot of cases (ie. `getEnsAddress`) – falling back could be an option, but behaviour might be unpredictable (ie. extra RPC call when it's not really needed)?

PR #1411, opened 2023-10-30
fix(chains)!: renamed polygon mainnet

ennioVisco:
Addresses [the wrong name used for polygon mainnet](https://github.com/wagmi-dev/wagmi/issues/3171).

- Warning 1: I had to disable the pre-commit hook since I'm on Windows and currently bun is not an option
- Warning 2: This is a significant breaking change for people using polygon. Perhaps an alias could be added to avoid this.

---

## PR-Codex overview

> The following files were skipped due to too many changes: `pnpm-lock.yaml`


jxom:
I’m a bit hesitant to merge this change just to make MetaMask happy. What if other wallets also implement such functionality but expect “Polygon”?

jxom:
Many Wallets and Dapps use the terminology "Polygon" (Rainbow, Opensea, Coinbase Wallet, Uniswap, etc), and I don't think I've seen any application apart from chainlist display a "Mainnet" suffix, so I believe that appending "Mainnet" may cause a bit of inconsistency & confuse mainstream users. 

Raised a bug report here: https://github.com/MetaMask/metamask-extension/issues/21609

PR #1415, opened 2023-10-31
Fix batch scheduler sort results

0xKheops:
**Focus of the PR** : sort batched rpc responses in http() transport, for cases where the RPC provides them in an unexpected order

**Context of the PR** : migrating Talisman browser extension wallet from ethers to viem

When batching rpc requests, some RPCs do not return responses in the same order as requests.
Viem expects them to be in the same order, which causes a bunch of random bugs in the wallet.

In the example below, `client.getGasPrice()` breaks as it tries to parse the rpc response of `client.getBlock()`.
It can be reproduced once out of 10 batches, approximately, when using https://rpc.ankr.com/bsc

Batch requests
image

Batch responses (first 2 responses aren't in the wrong order)
image

Fixed it by adding a an optional sort method to the `createBatchScheduler` method
I hope you'll like it, please let me know if anything needs to be changed.

Note : I couldn't get `bun run test` to run on my machine, getting a "Anvil failed to start in time" error systematically.

---

### Focus of the PR:
The focus of this PR is to add a sorting mechanism to the batch scheduler.

### Detailed summary:
- Added a `sort` parameter to the `createBatchScheduler` function.
- Added a `BatchResultsCompareFn` type.
- Modified the `createBatchScheduler` function to sort the data if the `sort` parameter is provided.

### DIFF
--- a/.changeset/stupid-news-wink.md
+++ b/.changeset/stupid-news-wink.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added sorting mechanism to batch scheduler.
--- a/src/clients/transports/http.ts
+++ b/src/clients/transports/http.ts
@@ -98,6 +98,7 @@ export function http(
                 fetchOptions,
                 timeout,
               }),
+            sort: (a, b) => a.id - b.id,
           })
 
           const fn = async (body: RpcRequest) =>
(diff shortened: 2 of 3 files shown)


jxom:
Seems fine! Thanks for the PR.

PR #1499, opened 2023-11-15
fix: CIP-64 are sent as EIP-1559 transactions

shazarre:
This PR fixes bug where CIP-64 transactions are sent as EIP-1559 transactions.

It reverts code changes to `src/chains/celo/utils.ts` from [this commit](https://github.com/wagmi-dev/viem/commit/494dc538129e49fa8fa01b7e6d073605b0f00bde) and repurposes a test that started failing because of the change.

# Context

When `prepareTransactionRequest` [is called](https://github.com/wagmi-dev/viem/blob/main/src/actions/wallet/sendTransaction.ts#L167) and type is not explicitly provided then it [will be determined](https://github.com/wagmi-dev/viem/blob/main/src/actions/wallet/prepareTransactionRequest.ts#L148) by `getTransactionType` and because for CIP-64 all EIP-1559 fields are provided it will be set to `eip1559` as per:

```typescript
if (
    typeof transaction.maxFeePerGas !== 'undefined' ||
    typeof transaction.maxPriorityFeePerGas !== 'undefined'
  )
    return 'eip1559' as GetTransactionType<TTransactionSerializable>
```

[Previously the logic of "forcing" the type](https://github.com/wagmi-dev/viem/commit/c0da695a76225c7f24ca263f9109e34d8cd93dba#diff-f51139e09908a7b2fef2022c35f6c86577299a5db761fcc5db4d8cc8c10561dfR62) in `src/chains/celo/utils.ts` allowed it only if the type was explicitly set to `cip64` and otherwise it would try to determine if it's a CIP-64 transaction based on the provided fields:

```typescript
return (
    isEIP1559(transaction) &&
    isPresent(transaction.feeCurrency) &&
    isEmpty(transaction.gatewayFee) &&
    isEmpty(transaction.gatewayFeeRecipient)
)
```

After the aforementioned commit it won't ever get to determine the type based on the fields as it will always go into `if (transaction.type)` branch as the `eip1559` type is injected hence resulting in sending wrong transaction type to the blockchain.

---

### Detailed summary
- Fixes an issue where CIP-64 transactions were mistakenly sent as EIP-1559 transactions.
- Updated the condition to check if `feeCurrency` is a token address for CIP-64 transactions.
- Updated the logic to determine if a transaction is CIP-64 or CIP-42.
- Added tests to ensure proper recognition of CIP-64 transactions.
- Updated the `sendTransaction()` function to provide valid transaction params for CIP-64 and CIP-42 transactions.

### DIFF
--- a/.changeset/friendly-numbers-rest.md
+++ b/.changeset/friendly-numbers-rest.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixes an issue where CIP-64 are sent as EIP-1559 transactions
--- a/src/chains/celo/sendTransaction.test.ts
+++ b/src/chains/celo/sendTransaction.test.ts
@@ -1,0 +1,99 @@
+import { describe, expect, test, vi } from 'vitest'
+import { accounts } from '~test/src/constants.js'
+import { privateKeyToAccount } from '~viem/accounts/privateKeyToAccount.js'
+import {
+  type EIP1193RequestFn,
+  type PublicRpcSchema,
+  type WalletRpcSchema,
+  createTransport,
+  createWalletClient,
+} from '../../index.js'
+import { celo } from '../index.js'
+
+describe('sendTransaction()', () => {
+  // We need a local account
+  const account = privateKeyToAccount(accounts[0].privateKey)
+  const toAddress = account.address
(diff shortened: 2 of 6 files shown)


shazarre:
@jxom could you have a look at this pls?

PR #1520, opened 2023-11-20
Patch ccipFetch method when URL includes {sender}

lucemans:
As specified in [EIP-3668 # Gateway Interface](https://eips.ethereum.org/EIPS/eip-3668#gateway-interface), if the URL template contains the {data} substitution parameter, the client MUST send a GET request after replacing the substitution parameters.

image

This should not however include the `{sender}` parameter as its safe to use in `POST` requests (see example below)

image

This PR updates the `ccipFetch` method such that URLs with `{sender}` execute as `POST` instead of `GET` in accordance with spec.

---

### Focus of the PR:
The focus of this PR is to patch the `ccipFetch` method in order to use POST by default when the URL includes `{sender}`.

### Detailed summary:
- Patched the `ccipFetch` method in `ccip.ts` to use POST by default when the URL includes `{sender}`.
- Updated the logic in the `for` loop in `ccip.ts` to determine the method based on whether the URL includes `{data}` or not.

### DIFF
--- a/.changeset/slow-chicken-leave.md
+++ b/.changeset/slow-chicken-leave.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Patch ccipFetch method when URL includes {sender}
+Patched `ccipFetch` method to use POST by default when URL includes `{sender}`.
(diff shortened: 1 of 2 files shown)


jxom:
Is this PR ready to review?

lucemans:
Yup, ready for review, just doing some last minute duedilligence but **can confirm** this is the right way!

jxom (approved):

PR #1596, opened 2023-12-17
test(accounts): add better test case for generating private keys

neuodev:
### Focus of this PR:
This PR focuses on adding a test case for the `generatePrivateKey` function.

### Detailed summary:
- Added imports for `secp256k1`, `expect`, `it`, and `vi`.
- Imported type `Hex` from `~viem/index.js`.
- Updated the test case to use the `it` function instead of `test`.
- Added a new test case to check if the `generatePrivateKey` function generates a random private key correctly.

### DIFF
--- a/src/accounts/generatePrivateKey.test.ts
+++ b/src/accounts/generatePrivateKey.test.ts
@@ -1,7 +1,22 @@
-import { expect, test } from 'vitest'
+import { secp256k1 } from '@noble/curves/secp256k1'
+import { expect } from 'chai'
+import { it, vi } from 'vitest'
 
+import type { Hex } from '~viem/index.js'
 import { generatePrivateKey } from './generatePrivateKey.js'
 
-test('default', () => {
-  expect(generatePrivateKey()).toBeDefined()
+it('should generate random private key', () => {
+  const hex: Hex =
+    '0xb4eb3c48376e1dc6e236c2004aea6f70d598de97825d15b7e97eaf89057ff9b9'
+
+  vi.spyOn(secp256k1.utils, 'randomPrivateKey').mockImplementationOnce(() =>
+    Uint8Array.from(
+      Buffer.from(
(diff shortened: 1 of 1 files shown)


src/accounts/generatePrivateKey.test.ts
```diff
@@ -1,7 +1,22 @@
-import { expect, test } from 'vitest'
+import { secp256k1 } from '@noble/curves/secp256k1'
+import { expect } from 'chai'
```
jxom, line 2: We don't use `chai`. I think you mean `vitest`?

src/accounts/generatePrivateKey.test.ts
```diff
@@ -1,7 +1,22 @@
...
+import { it, vi } from 'vitest'
 
+import type { Hex } from '~viem/index.js'
 import { generatePrivateKey } from './generatePrivateKey.js'
 
-test('default', () => {
-  expect(generatePrivateKey()).toBeDefined()
+it('should generate random private key', () => {
```
jxom, line 8:
```suggestion
test('default', () => {
```

src/accounts/generatePrivateKey.test.ts
```diff
@@ -1,7 +1,22 @@
...
+  const hex: Hex =
+    '0xb4eb3c48376e1dc6e236c2004aea6f70d598de97825d15b7e97eaf89057ff9b9'
+
+  vi.spyOn(secp256k1.utils, 'randomPrivateKey').mockImplementationOnce(() =>
```
jxom, line 12: Can we do this without mocking the implementation? 

```
const spy = vi.spyOn(secp2561.utils, 'randomPrivateKey')

expect(spy).toBeCalled()
```

jxom:
Closing for now as this PR is becoming stale. Please re-open when the comments above are addressed! Thanks.

PR #1602, opened 2023-12-20
feat: add alchemy as rpc url for base sepolia

avasisht23:
## PR-Codex overview
This PR focuses on adding RPC URLs to the Base Sepolia chain definition.

### Detailed summary
- Added RPC URLs for Alchemy HTTP and WebSocket endpoints to the Base Sepolia chain definition.

### DIFF
--- a/.changeset/curvy-tips-fetch.md
+++ b/.changeset/curvy-tips-fetch.md
@@ -1,5 +1,5 @@
 ---
-"viem": minor
+"viem": patch
 ---
 
 feat: add alchemy as rpc url for base sepolia
--- a/.changeset/curvy-tips-fetch.md
+++ b/.changeset/curvy-tips-fetch.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-feat: add alchemy as rpc url for base sepolia
+Added RPC URLs to Base Sepolia
--- a/.changeset/curvy-tips-fetch.md
+++ b/.changeset/curvy-tips-fetch.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
+feat: add alchemy as rpc url for base sepolia
--- a/src/chains/definitions/baseSepolia.ts
+++ b/src/chains/definitions/baseSepolia.ts
@@ -8,6 +8,10 @@ export const baseSepolia = /*#__PURE__*/ defineChain(
     name: 'Base Sepolia',
     nativeCurrency: { name: 'Sepolia Ether', symbol: 'ETH', decimals: 18 },
     rpcUrls: {
+      alchemy: {
+        http: ['https://base-sepolia.g.alchemy.com/v2'],
+        webSocket: ['wss://base-sepolia.g.alchemy.com/v2'],
+      },
       default: {
         http: ['https://sepolia.base.org'],
       },


tmm:
FYI [Viem v2](https://github.com/wevm/viem/pull/1429) will be released soon and will no longer accept these type of pull requests because we are moving away from hardcoding provider URLs and are only going to keep track of public resources.

cainaleaouk:
Hey guys, this a huge blocker for us at the moment. Any change we could get this merged soon?

avasisht23:
@tmm so how would devs use RPC providers like Alchemy in the future? hard-coding the URLs themselves? I found this util in viem to be quite helpful.

tmm:
You would copy the RPC URL from your providers dashboard. You already need to go there to get the API key. This way we don't need to hardcode anything.

cainaleaouk:
> You would copy the RPC URL from your providers dashboard. You already need to go there to get the API key. This way we don't need to hardcode anything.

So just making the devs life a bit harder then, for not real gain?

cainaleaouk:
Also... Any chance this can get merged, still waiting on this to unblock the migration from goerli to sepolia

tmm:
> So just making the devs life a bit harder then, for not real gain?

@cainaleaouk the only difference is you copy-paste the second or third option instead of the first.

000278_Firefox_2023-12-21-10 27 06@2x

> Also... Any chance this can get merged, still waiting on this to unblock the migration from goerli to sepolia

Feel free to copy the code and define the chain in your repo manually while you wait! Pretty simple workaround and will unblock you.

tmm (changes requested):
Please add a changeset with `bun changeset`

cainaleaouk:
> > So just making the devs life a bit harder then, for not real gain?
> 
> @cainaleaouk the only difference is you copy-paste the second or third option instead of the first.
> 
> 000278_Firefox_2023-12-21-10 27 06@2x
> 
> > Also... Any chance this can get merged, still waiting on this to unblock the migration from goerli to sepolia
> 
> Feel free to copy the code and define the chain in your repo manually while you wait! Pretty simple workaround and will unblock you.



> > So just making the devs life a bit harder then, for not real gain?
> 
> @cainaleaouk the only difference is you copy-paste the second or third option instead of the first.
> 
> 000278_Firefox_2023-12-21-10 27 06@2x
> 
> > Also... Any chance this can get merged, still waiting on this to unblock the migration from goerli to sepolia
> 
> Feel free to copy the code and define the chain in your repo manually while you wait! Pretty simple workaround and will unblock you.

So how about people who don't use the API key? We connect to alchemy provider via JWT, and then (not sure but that's my understanding) will take care of appending it to the url as long as it exists (in this case it doesnt).

Also your answer is not complete enough, is not copying the second link only, I need to then inject that into the chain information as I'm doing locally now to unblock myself like you pointed out.

```
const baseSepolia = SupportedChains.get(config.chainIds.baseSepolia.numId);
baseSepolia.rpcUrls.alchemy = { http: ['https://base-sepolia.g.alchemy.com/v2'] };
```

But I can't help to think it's silly to have to do that for every project we have since we already had a nice util. And the reason for not having the utils is, to me, quite weak.

Could you elaborate, so I can understand, what is your beef with the current util? Why does it hurt to have hardcoded values in the lib?

jxom:
Setting an RPC URL is something you do once and shouldn’t need to touch again. Surely that isn’t too hard? 😅

tmm:
> Why does it hurt to have hardcoded values in the lib?

Because we don't want to be the source of truth for this information and there are tens of providers (and if we let Alchemy hardcode their values, then all of them will want to).

cainaleaouk:
> Setting an RPC URL is something you do once and shouldn’t need to touch again. Surely that isn’t too hard? 😅

Not hard, just yak shaving

avasisht23:
added @tmm

.changeset/curvy-tips-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"viem": minor
```
tmm, line 2:
```suggestion
"viem": patch
```

.changeset/curvy-tips-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"viem": patch
+---
+
+feat: add alchemy as rpc url for base sepolia
```
tmm, line 5:
```suggestion
Added RPC URLs to Base Sepolia
```

PR #1712, opened 2024-01-18
Update base sepolia explorer url

0xOlias:
It looks like Base may be migrating from a Blockscout explorer (https://base-sepolia.blockscout.com) to Etherscan (https://sepolia.basescan.org).

EDIT: 

@tmm I saw that you recently removed non-default explorer URLs. FWIW, this may be a case where it would be best if both were included - Jesse from Base says both explorers will be around long term (https://warpcast.com/jessepollak/0x067ed15a).

Also, let me know if you'd rather Ponder not use `viem/chains` as a ~registry for explorer URLs in this way!

---

## PR-Codex overview
This PR focuses on updating the Base Sepolia explorer URL. 

### Detailed summary
- Updated the name of the default block explorer to "Basescan".
- Updated the URL of the default block explorer to "https://sepolia.basescan.org".
- Updated the API URL of the default block explorer to "https://api-sepolia.basescan.org/api".

### DIFF
--- a/src/chains/definitions/baseSepolia.ts
+++ b/src/chains/definitions/baseSepolia.ts
@@ -16,7 +16,7 @@ export const baseSepolia = /*#__PURE__*/ defineChain({
   },
   blockExplorers: {
     default: {
-      name: 'Blockscout',
+      name: 'Basescan',
       url: 'https://sepolia.basescan.org',
       apiUrl: 'https://api-sepolia.basescan.org/api',
     },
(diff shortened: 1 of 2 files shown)


src/chains/definitions/baseSepolia.ts
```diff
@@ -17,8 +17,8 @@ export const baseSepolia = /*#__PURE__*/ defineChain({
   blockExplorers: {
     default: {
       name: 'Blockscout',
```
jxom, line 19:
```suggestion
      name: 'Basescan',
```

PR #1733, opened 2024-01-24
fix: getAddress

tmm:
Calling `getAddress` on multiple addresses (e.g. `['0x...', '0x...'].map(getAddress)`) in different orders doesn't return consistent results.

Minimal reproduction: https://stackblitz.com/edit/viem-getting-started-vn6ea3?file=index.ts

Related https://github.com/wevm/wagmi/issues/3507

---

### Detailed summary
- Added a new test case for handling multiple addresses in the `getAddress` utility function.
- Created two arrays of addresses and mapped them using the `getAddress` function.
- Asserted that the two arrays of addresses are equal using the `toEqual` matcher.

### DIFF
--- a/src/utils/address/getAddress.test.ts
+++ b/src/utils/address/getAddress.test.ts
@@ -57,3 +57,16 @@ describe('errors', () => {
     `)
   })
 })
+
+test('multiple addresses', () => {
+  const addresses = [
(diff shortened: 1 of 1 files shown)


tmm:
lol this is because `['0x...', '0x...'].map(getAddress)` implicitly passes `index` as `chainId` (`getAddress` second parameter) 🥲

PR #1769, opened 2024-02-02
feat: make parseUnits handle scientific notation 

hskang9:
Viem breaks numbers when the input turns into scientific notation after calculating amounts to return in a decentralized exchange. 

if the input is parseUnits("4.5e-7", 6), the result is BigInt(`4.5e-7000000`). 

I add this function to properly handle the case when Typescript is unable to detect when scientific notation is used with its glorified type.

---

### Detailed summary
- Added a new test case for parsing scientific notations in `parseUnits.test.ts`
- Implemented `scientificToBigInt` function in `parseUnits.ts` to handle scientific notation values
- Updated `parseUnits` function to handle scientific notation values


hskang9:
parseEther should be updated accordingly.

src/utils/unit/parseUnits.ts
```diff
@@ -14,8 +14,20 @@ export type ParseUnitsErrorType = ErrorType
  * // 420000000000n
  */
 export function parseUnits(value: string, decimals: number) {
+  const scientificToBigInt = (scientificNotation) => {
+    const [coefficient, exponent] = scientificNotation.toLowerCase().split('e');
+    const shift = BigInt(exponent);
+    const shiftedValue = BigInt(coefficient.replace('.', '')) * 10n ** shift;
```
jxom, line 20: Wouldn't this break for negative exponents (ie. `4.5e-7`)?

jxom (commented):
Thanks for the PR. If we can add test cases for both positive & negative exponents, that would be great.

src/utils/unit/parseUnits.ts
hskang9: here is the correct code
```
const scientificToBigInt = (scientificNotation) => {
        const [coefficient, exponent] = scientificNotation.toLowerCase().split('e');
        const shift = BigInt(exponent);
        const shiftedValue = (shift >= 0n  ? BigInt(coefficient.replace('.', '')) * 10n ** shift :  BigInt(coefficient.replace('.', '')) / (10n ** (-shift)));
        return shiftedValue;
      };
      ```
      Still testing

Comment (hskang9) [reply src/utils/unit/parseUnits.ts:20]: if negative exponent's absolute value is more than the (unit argument + 1), then yes.

Comment (hskang9) [reply src/utils/unit/parseUnits.ts]: @jxom  I added both negative & positive exponent cases.

When I tried to test the case with command `yarn test parseUnits.test.ts`, the codebase requires some generated abi from a script.

Can you help me how to remove this error?

```
 FAIL  src/utils/unit/parseUnits.test.ts [ src/utils/unit/parseUnits.test.ts ]
Error: Failed to load url ../contracts/generated.js (resolved id: ../contracts/generated.js) in /Users/hyungsukkang/hskang9/viem/test/src/utils.ts. Does the file exist?
 ❯ loadAndTransform node_modules/vite/dist/node/chunks/dep-9A4-l-43.js:49731:21
 ```

jxom:
Thanks for the updates. Will review tomorrow.

src/utils/unit/parseUnits.ts
hskang9: Done. I linted and formatted the code.

jxom:
Going to have #1790 supersede this. When I was testing this, I noticed that it didn't handle decimal output (ie. `12.345e2`), and only bigints. I also noticed some edge cases weren't being handled (ie. rounding). I think it might be nicer if we can first convert the scientific notation string into a decimal string, and then feed that through the existing logic.

Your commits are still preserved on #1790, so you still will get the contribution/credit for your efforts on this PR!

hskang9:
> Thanks so much for the PR!
> 
> I'm going to have #1790 supersede this. When I was testing this, I noticed that it didn't handle fractional components (ie. `parseUnits('0.6959e2', 9)`). I also noticed some edge cases weren't being handled (ie. rounding). I think it might be nicer if we can first convert the scientific notation string into a decimal string, and then feed that through the existing logic (see here: https://github.com/wevm/viem/pull/1790/files#diff-5d352c8e3cb3ef0d9e5e97305068adbbc533454139f38b2052abe2442ce33e9eR17).
> 
> Your commits are still preserved on #1790, so you still will get the contribution/credit for your efforts on this PR!

If it is fractional and it is below the represented decimal, I think the number should be zero. What other test cases can you think?

hskang9:
Sorry for late response, I had to update to new viem version to my app. I deeply feel sad it often breaks the original code, but anyways I strongly agree to your thought.

hskang9:
I think it would be also good to test cases where fractional numbers are cut in case of 1.0000….015. When last digit cannot be represented, it should be rounded down or just remove it, but preserve integer part.

PR #1790, opened 2024-02-08
feat: handle scientific notation in `parseUnits`

jxom:
### Focus of this PR:
Refactoring the `parseUnits` function to handle scientific notation.

### Detailed summary:
- Renamed the `value` parameter in the `parseUnits` function to `value_`.
- Added a new function `scientificToDecimal` to convert numbers in scientific notation to decimal notation.
- Updated the `parseUnits` function to use `scientificToDecimal` if the input value contains 'e'.
- Added tests for the new functionality.


hskang9:
You did not fix the bug and crashed the indexer. 

I even put the code and the test, and I don't see why you have added this feature.

0xSimbo:
Was this ever deployed? I just ran into a bug where scientific notation was not handled.

Using viem: ^2.21.2
Example:
```const value  = 9.955189695275411e-7;
const result = parseUnits(value.toString(), 6)
console.log(result)
//Prints 9955180n, when it should print 0 or 1 most likely due to rounding.```

Probably a naive way to handle but i refactored the parseUnits function  to look like this:

```export function parseUnits(value: string, decimals: number,options?:{}) {
    if(value.includes("e-")) {
        //Strip the values on the left of the decimal point, and add the same amount of 0's to the right, and then put the left units back right after the zeros
        const [numWithoutScientificNotation, scientificNotationNegativeExponent] = value.split("e-")
        const [left, right] = numWithoutScientificNotation.split(".")
        const numZeroes = Number(scientificNotationNegativeExponent)-1
        const zeros = "0".repeat(numZeroes)
        value = `0.${zeros}${left}${right}`
        console.log(value);
    }
  let [integer, fraction = '0'] = value.split('.')
  console.log(integer, fraction)

  const negative = integer.startsWith('-')
  if (negative) integer = integer.slice(1)

  // trim trailing zeros.
  fraction = fraction.replace(/(0+)$/, '')

  // round off if the fraction is larger than the number of decimals.
  if (decimals === 0) {
    if (Math.round(Number(`.${fraction}`)) === 1)
      integer = `${BigInt(integer) + 1n}`
    fraction = ''
  } else if (fraction.length > decimals) {
    const [left, unit, right] = [
      fraction.slice(0, decimals - 1),
      fraction.slice(decimals - 1, decimals),
      fraction.slice(decimals),
    ]

    const rounded = Math.round(Number(`${unit}.${right}`))
    if (rounded > 9)
      fraction = `${BigInt(left) + BigInt(1)}0`.padStart(left.length + 1, '0')
    else fraction = `${left}${rounded}`

    if (fraction.length > decimals) {
      fraction = fraction.slice(1)
      integer = `${BigInt(integer) + 1n}`
    }

    fraction = fraction.slice(0, decimals)
  } else {
    fraction = fraction.padEnd(decimals, '0')
  }

  return BigInt(`${negative ? '-' : ''}${integer}${fraction}`)
}
```

The diff is here,
```    if(value.includes("e-")) {
        //Strip the values on the left of the decimal point, and add the same amount of 0's to the right, and then put the left units back right after the zeros
        const [numWithoutScientificNotation, scientificNotationNegativeExponent] = value.split("e-")
        const [left, right] = numWithoutScientificNotation.split(".")
        const numZeroes = Number(scientificNotationNegativeExponent)-1
        const zeros = "0".repeat(numZeroes)
        value = `0.${zeros}${left}${right}`
        console.log(value);
    }```
    and it only handles e- , not e without a negative.

**Comment by jxom** [comment]
Scientific notation must be handled before providing it as input:

```typescript
const units = new Intl.NumberFormat('en-US', { useGrouping: false, style: 'decimal', maximumFractionDigits: 18 }).format('1e21')
// 1000000000000000000000
```

0xSimbo:
Ok, this would have been good to know, or at least have an error thrown because it lead to a critical bug in our production code. If governance hadn't caught it in time, it could gave had financial consequences.

jxom:
Got it. We can throw an error for the time being.

jxom:
Added https://github.com/wevm/viem/commit/19a66282e6e1d2a7702cb82f1f12b69141389b76. Will be out in 2.21.3.

0xSimbo:
Got it. That would be great to have it auto-formatting, but if that doesen't work out, having an error thrown would have prevented this bug for us.

Issue #1803, opened 2024-02-13
bug: conform `isAddress` to EIP-55 checksum validation


dawsbot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

2.7.6

### Current Behavior

The `isAddress` function does not perform checksum validation.

### Expected Behavior

An address with an invalid checksum should fail the `isAddress` function.

### Steps To Reproduce

1. Change this unit-test to falsy instead of truthy:
https://github.com/dawsbot/viem/blob/12d61c5c4b965f11e5e819fdff9dd87dd9eb9d55/src/utils/address/isAddress.test.ts#L6
2. Run `./node_modules/.bin/vitest /Users/db/code/scratch/viem/src/utils/address/isAddress.test.ts`
3. See that the tests now fail because the first test is an address with an invalid checksum (`'0xa5cc3c03994db5b0d9a5eEdD10Cabab0813678ac'` is invalid and `'0xa5cc3c03994DB5b0d9A5eEdD10CabaB0813678AC'` is valid)

### Link to Minimal Reproducible Example (StackBlitz, CodeSandbox, GitHub repo etc.)

https://stackblitz.com/edit/viem-getting-started-43axcq?file=index.ts

### Anything else?

If you change the import to `ethers` in the unit-testing, you'll see that the current viem library does not match the behavior of `ethers`

<img width="984" alt="image" src="https://github.com/wevm/viem/assets/3408480/b5a92b0a-4be2-4bca-b077-7595199098b5">

PR #1804, opened 2024-02-13
Fix checksum validation on isAddress

dawsbot:
I doubt I'm doing the proper code-style, but this is a POC for closing #1803

---

### Focus of this PR:
Upgrade the `isAddress` function to return `false` for invalid checksummed addresses.

### Detailed summary:
- Modified the `isAddress` function to include an optional `strict` parameter.
- Updated multiple files to use the new `isAddress` function with `strict: false` parameter.
- Added error messages for invalid addresses, including requirements for valid addresses.
- Added tests to ensure the validity of addresses and error handling.
- Added a new `LruMap` class for caching with a LRU policy.

> The following files were skipped due to too many changes: `src/actions/public/call.test.ts`, `src/utils/address/isAddress.ts`, `src/utils/address/getAddress.test.ts`

### DIFF
--- a/src/utils/address/isAddress.ts
+++ b/src/utils/address/isAddress.ts
@@ -1,18 +1,18 @@
 import type { Address } from 'abitype'
 import type { ErrorType } from '../../errors/utils.js'
-import { checksumAddress} from './getAddress.js'
+import { checksumAddress } from './getAddress.js'
 
 const addressRegex = /^0x[a-fA-F0-9]{40}$/
 
 export type IsAddressErrorType = ErrorType
 
 export function isAddress(address: string): address is Address {
-  if (!addressRegex.test(address) ) {
-    return false;
+  if (!addressRegex.test(address)) {
+    return false
   }
   if (address.toLowerCase() === address) {
-    return true;
+    return true
   }
   // @ts-ignore
-  return checksumAddress(address) === address;
+  return checksumAddress(address) === address
 }
--- a/src/utils/address/isAddress.ts
+++ b/src/utils/address/isAddress.ts
@@ -1,10 +1,18 @@
 import type { Address } from 'abitype'
 import type { ErrorType } from '../../errors/utils.js'
+import { checksumAddress} from './getAddress.js'
 
 const addressRegex = /^0x[a-fA-F0-9]{40}$/
 
 export type IsAddressErrorType = ErrorType
 
 export function isAddress(address: string): address is Address {
-  return addressRegex.test(address)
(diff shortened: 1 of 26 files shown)


src/utils/address/isAddress.ts
```diff
@@ -1,10 +1,18 @@
...
 export function isAddress(address: string): address is Address {
-  return addressRegex.test(address)
+  if (!addressRegex.test(address) ) {
+    return false;
+  }
+  if (address.toLowerCase() === address) {
+    return true;
+  }
+  // @ts-ignore
+  return checksumAddress(address) === address;
```
dawsbot, line 17: I was not sure your standard practice here for going from `string` to `0x${string}`, so I threw in a "`// @ts-ignore`" but would love to learn how you handle this properly

src/utils/lru.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Map with a LRU (Least recently used) policy.
+ *
+ * @link https://en.wikipedia.org/wiki/Cache_replacement_policies#LRU
+ */
+export class LruMap<value = unknown> extends Map<string, value> {
```
dawsbot, line 6: Great optimization!

src/utils/address/isAddress.ts
```diff
@@ -1,18 +1,35 @@
 import type { Address } from 'abitype'
 import type { ErrorType } from '../../errors/utils.js'
+import { LruMap } from '../lru.js'
 import { checksumAddress } from './getAddress.js'
 
 const addressRegex = /^0x[a-fA-F0-9]{40}$/
 
+const isAddressCache = /*#__PURE__*/ new LruMap<boolean>(8192)
```
dawsbot, line 8: Can you teach me what the thinking was for `8192`?

jxom: No reason really, just wanted to have an acceptable enough memory size to store an acceptable amount of addresses!

8192 * 20B = ~163kB

malonehedges:
Just came across this, I disagree with this change. This approach breaks the assumptions I had around how this function works (can I convert this into bytes and put it in my database). This leads to case-sensitive API routes which wasn't my intention when implementing my API. I can see how validating a checksum could be useful to some, I feel like this could be solved with either a new function `isAddressStrict` or an option passed to this function so that not every use requires lowercased or checksummed addresses.

malonehedges:
I moved too quickly, I see it is an option, just is defaulted to true which is not what I expected. I will change my usage to pass this argument, but was still surprised by this default behavior change

PR #1817, opened 2024-02-16
Update SKALE Testnet Chains

TheGreatAxios:
## Notes
1. SKALE Testnets have been migrated from Goerli to Holesky
2. All four core hubs (Calypso, Europa, Nebula, Titan) have updated RPCs, ChainIds, WSS Urls, and Block Explorers
3. Multicall3 re-deployed to new chains: https://github.com/mds1/multicall/pull/214
4. Note that Chaos testnet has been deprecated and so the file was deleted
5. Did not create a changeset as this does not impact core Viem, but happy to if you would like me too.

## Question
Looking for guidance from the Viem Team. SKALE does have some additional custom contracts pre-deployed on every SKALE Chain. I noticed that there are custom folders for some other stacks. Can I create a PR for SKALE to add this additional information to make Viem an even stronger choice for new SKALE developers?

Thanks!

---

## PR-Codex overview
The focus of this PR is to migrate SKALE Testnet chains to new testnet networks.

### Detailed summary
- Migrated SKALE Testnet chains to new testnet networks
- Updated chain IDs and names for SKALE Nebula, Europa, Titan, and Calypso testnets
- Updated RPC URLs and block explorers for the testnet networks


TheGreatAxios:
Hey @jxom 

Wanted to check in here and see if there is anything else that you require to move this forward? 

Thank you!

jxom:
Sorry about the late reply. 

It looks like these changes are breaking as we are modifying chain IDs, which could break some consumers that rely on those existing networks (even though they may be deprecated, it's still a breaking change).

To avoid a breaking change, can you please add new chains for these changes, and mark the existing chains as deprecated by adding a ```/** @deprecated Use `foo` instead */```. Once we prepare a Viem v3 release, we can remove the old ones.

TheGreatAxios:
Hey @jxom 

No worries -- I know you are incredibly busy, this is just urgent on the SKALE end as we have many chains and not a single testnet. To provide you some context on why I don't think what you said last makes sense in our case (I do understnad your thought process):

1. The SKALE Testnets (staging-v3 endpoints and urls) were all linked to Goerli as of the 14th of February
2. New SKALE Testnet Chains were created (testnet endpoints) on Holesky
3. The core hubs (Calypso, Europa, Titan, Nebula) have retained their optical name since they match the SKALE Mainnet setup, however, the endpoints and chain ids have changed.
4. Because the Goerli chains were spun down entirley (i.e deleted) it's actually breaking people testing now because Viem/Wagmi/Rainbowkit is defaulting to chains that don't exist. 
5. As these chains on Goerli no longer exist, no developer is activley trying to use them. They all are using the new chains as per the code update. 
6. Lastly, removal of Chaos + Pscyhe chains are because the community chose to not re-create those on the new testnet system due to design changes with the setup and them no longer being needed.

NOTE** This does not impact the SKALE Mainnet Chains at all OR their Viem config

Please let me know if any other questions. Additionally, if you need proof of the confusion you can go to the [sFUEL Station](https://sfuelstation.com) and when selecting to switch to Calypso Testnet you will see it try to connect to staging-utter-unripe-menkar which no longer exists. 

Thanks again for your great work!

jxom:
Since they are defunkt, happy to merge this.

TheGreatAxios:
Thanks so much!!!
I did leave a note as well related to some questions I have about adding network specific functionality to Viem. 

Do you have any guidance or docs on what is acceptable?

Thanks again!!!

PR #1829, opened 2024-02-21
Fix localhost chain ID to 31_337 from 1_337

twynne20:
Fixed localhost chain ID to match Hardhat and Foundry defaults.

Changed id: 1_337 to id: 31_337 in [viem/src/chains/definitions/localhost.ts](https://github.com/wevm/viem/blob/main/src/chains/definitions/localhost.ts)

---

Identified from seeing the following error on hardhat & anvil:
```
eth_sendRawTransaction

  Trying to send a raw transaction with an invalid chainId. The expected chainId is 31337

```

---

## PR-Codex overview
This PR increases the `id` of the `localhost` chain definition from 1_337 to 31_337.

### Detailed summary
- Updated the `id` of the `localhost` chain definition from 1_337 to 31_337.

### DIFF
--- a/src/chains/definitions/localhost.ts
+++ b/src/chains/definitions/localhost.ts
@@ -1,7 +1,7 @@
 import { defineChain } from '../../utils/chain/defineChain.js'
 
 export const localhost = /*#__PURE__*/ defineChain({
-  id: 1_337,
+  id: 31_337,
   name: 'Localhost',
   nativeCurrency: {
     decimals: 18,


holic:
I think you want to use the foundry chain? https://github.com/wevm/viem/blob/main/src/chains/definitions/foundry.ts

twynne20:
Hey Holic,

Thank you for pointing out the existence of `foundry.ts` - that was an oversight on my part, I looked for `anvil.ts` instead. Is a localhost chainID typically `1_337` (Ganache) more often than `31_337`? I still feel aligning the `localhost.ts` `id` with Hardhat and Foundry's default of `31_337` is beneficial for a couple of reasons:

1. **Consistency Across Tools**: Developers (especially new ones) using Ethereum development environments like Hardhat and Foundry might expect the `chainId` for localhost to be `31_337`, as it's the default setting for these tools. Hardhat logs that the `chainID` is incorrect, while Anvil does not, which can make it quite confusing for new comers who start on Anvil. This means a file should be added for `ganache.ts` if that were to happen. 

2. **Clarity and Documentation**: While Ganache defaults to `1_337`, and there currently isn't a specific configuration (`ganache.ts`) for Ganache, it might be worth considering either adding support for Ganache explicitly or documenting the `id` expectations for different tools within the project. This could help prevent confusion and support developers using any of these tools. NOTE: Aligning the `localhost.ts` to chainID `31_337` largely aligns with my opinion that Hardhat & Foundry are used more often then Ganache and people will think of it being `localhost` chain. 

Would it be possible to discuss this further, maybe consider adding a `ganache.ts` for explicit Ganache support, or at least add a note somewhere considerate to new developers to guide developers working with different local environments?

NOTE: This isn't the biggest deal, but comes as a recommendation from a hackathon where we had a few students confused & not seeing sendTransaction / useWriteContract working when trying to use `localhost` with Anvil / Hardhat.

jxom:
Thanks for the PR! I am happy to merge this. But this is a breaking change. Are you happy if I add a note to this and we can merge it when we eventually release v3?

For the meantime, we could probably deprecate the `foundry` chain, and rename it to `anvil`. Open to a PR for that!

twynne20:
Thanks for the response @jxom, do you all have a standard for marking something as deprecated? I would love to open a new PR for the foundry -> anvil change. 

I am assuming by this you just mean within a new PR: 
- Update [foundry.ts](https://github.com/wevm/viem/blob/main/src/chains/definitions/foundry.ts) to reflect anvil naming 
  - Remove `foundry.ts` and add `anvil.ts`
  - modify to `export const anvil` 
  - modify to `name: 'Anvil' `

Can you help me understand why the modification of `localhost.ts` is a breaking change, but not modifying `foundry.ts` to `anvil.ts`?

jxom:
@twynne20 – By deprecating, I mean marking the `foundry` export in `src/chains/index.ts` with a `/** @deprecated Use `anvil` instead */` JSDoc comment. This means that it will be removed in the next major version.

PR #1868, opened 2024-02-27
feat: add BNB greenfield chain mainnet

hstarorg:
* Add BNB Greenfield chain mainnet

https://greenfield.bnbchain.org/en


src/chains/definitions/bscGreenfield.ts
```diff
@@ -0,0 +1,21 @@
+import { defineChain } from '../../utils/chain/defineChain.js'
+
```
pegahcarter, line 2: Do you need to add a `sourceId` var here?

hstarorg: I don’t think it’s necessary, it’s similar to bsc mainnet

> Do you need to add a `sourceId` var here?

PR #1919, opened 2024-03-08
fix: estimateL1Fee

roninjin10:
fixes https://github.com/wevm/viem/issues/1918


roninjin10:
Still need to update tests so moving to draft

PR #1979, opened 2024-03-19
chore: demo fetch a better way

zhangshichun:
## PR-Codex overview
This PR changes certain types to be imported as `type` in `index.tsx`, removes the `hash` state, adds a `refreshTransactionReceipt` function, and updates the logic for setting transaction receipt.

### Detailed summary
- Updated types to be imported as `type`
- Removed `hash` state
- Added `refreshTransactionReceipt` function
- Updated logic for setting transaction receipt


jxom:
Thanks for the PR! This is done on purpose to avoid a long-running asynchronous function.

PR #1992, opened 2024-03-21
feat: add optional options to `toBlobs` to allow >2 blobs on tx

tunnckoCore:
Allows for making transactions with more than 2 blobs as currently. Resolves #1991 

It's not breaking change, current defaults still apply.

_Note: i tried to run it locally but it cannot even begin with install because it wants Forge which i don't have and don't want to add. So i moved to Codespaces._


jxom:
closing in favor of #1995

tunnckoCore:
@jxom AWESOME :tada: thanks!

PR #1993, opened 2024-03-21
chore: Add name to root package.json

roninjin10:
Add name to root package.json. This makes it so you can npm install viem directly from github which is helpful when trying to hack using an unpublished viem commit because you can install it, build viem, and then add a "viem": "file:path/to/viem/src" to package.json. 

Without this only way to do it is to use github submodules or to use unstable pnpm 9 which supports installing the `viem/src` nested folder

---

## PR-Codex overview
The focus of this PR is to update the project configuration by adding a new name "viem-root" and setting the type to "module".

### Detailed summary
- Added project name "viem-root"
- Set project type to "module"
- Updated `ignore` field in `changeset/config.json` to include "viem-root"

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -13,5 +13,5 @@
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
     "onlyUpdatePeerDependentsWhenOutOfRange": true
   },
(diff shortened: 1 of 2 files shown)


jxom:
Let's leave it for now and make it hard for consumers to do this. Using unreleased software on feature branches is a security concern.

PR #2175, opened 2024-04-23
recoverPublicKey: Does not need to be async

bh2smith:
There was an awkward async import inside this method making it asynchronous (when it doesn't need to be). A global search on this project shows that this import is being made everywhere else as the change proposed here. I am not sure if this was intended or what, but this change makes everything much more consistent.

**Note that** ⚠️  this change might require a version bump (for any dependents on this project) since users will have to remove `await` statements as was done here.

Screenshot 2024-04-23 at 13 25 08

---

## PR-Codex overview
The focus of this PR is to optimize the `recoverPublicKey` function by removing unnecessary async/await operations.

### Detailed summary
- Removed unnecessary async/await from `recoverPublicKey` function
- Updated function signature to remove Promise return type
- Removed import of `secp256k1` from `recoverPublicKey`
- Updated function calls in `recoverPublicKey.md` and test files to reflect changes

### DIFF
--- a/site/pages/docs/utilities/recoverPublicKey.md
+++ b/site/pages/docs/utilities/recoverPublicKey.md
@@ -11,7 +11,7 @@ Recovers the original signing 64-byte public key from a hash & signature.
 ```ts [example.ts]
 import { recoverPublicKey } from 'viem'
  
-const address = await recoverPublicKey({
+const address = recoverPublicKey({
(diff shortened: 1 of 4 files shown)


tmm:
Since this is a breaking change, probably best to add a [TODO comment for v3](https://github.com/wevm/viem/blob/main/src/accounts/utils/sign.ts#L1) instead.

PR #2223, opened 2024-05-07
fix: checksum contract address in logs formatter

banteg:
the `address` field in the returned logs wasn't checksummed, this takes care of it and updates the tests.


jxom:
Closing for now as this is a breaking change. Will add a note to revisit for v3.

PR #2404, opened 2024-06-13
fix: getAction with syncronous action bug

roninjin10:
Fixes corner case bug in getAction where if an action is syncronous it might fail if the action returns undefined or null.

---

## PR-Codex overview
This PR fixes a bug in `getAction` where synchronous actions could fail. 

### Detailed summary
- Fixed bug in `getAction` for synchronous actions
- Improved handling of synchronous actions returning falsy values

### DIFF
--- a/.changeset/twelve-hornets-explain.md
+++ b/.changeset/twelve-hornets-explain.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed bug where getAction could fail on decorated actions that are syncrounous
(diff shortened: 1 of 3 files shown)


roninjin10:
Haven't ran vitest yet struggling with this error `Error: Failed to load url ~viem/constants/abis.js (resolved id: ~viem/constants/abis.js) in /Users/williamcory/viem/test/src/abis.ts. Does the file exist?`

PR #2414, opened 2024-06-15
feat: added missing contracts to mode

turtlemoji:


jxom:
Addresses confirmed via https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/extra/addresses/mainnet/mode.json.

PR #2418, opened 2024-06-17
feat: nonce manager + dedupe in-flight requests

jxom:
## PR-Codex overview
This PR introduces support for a Nonce Manager and in-flight request deduplication. 

### Detailed summary
- Added `nonceManager` support for Local Accounts
- Implemented in-flight request deduplication for Transport JSON-RPC requests
- Updated various test files and utility functions
- Added new types and exports for `nonceManager`
- Updated package.json configurations for `nonce`
- Enhanced error handling in public actions
- Improved promise caching and deduplication functionality

> The following files were skipped due to too many changes: `src/actions/public/getBlock.ts`, `src/accounts/index.ts`, `src/accounts/types.ts`, `site/sidebar.ts`, `src/clients/transports/createTransport.ts`, `src/accounts/hdKeyToAccount.ts`, `src/actions/public/getTransaction.ts`, `src/accounts/privateKeyToAccount.ts`, `src/actions/wallet/prepareTransactionRequest.ts`, `src/actions/wallet/sendTransaction.test.ts`, `src/utils/promise/withDedupe.test.ts`, `site/pages/docs/accounts/createNonceManager.md`, `src/utils/nonceManager.ts`, `src/utils/buildRequest.test.ts`, `src/clients/transports/http.test.ts`, `src/utils/buildRequest.ts`, `src/utils/nonceManager.test.ts`, `src/actions/wallet/prepareTransactionRequest.test.ts`


src/utils/nonceManager.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export type NonceManager = {
+  /** Get and increment a nonce. */
+  consume(parameters: FunctionParameters & { client: Client }): Promise<number>
+  /** Increment a nonce. */
+  increment(chainId: FunctionParameters): Promise<void>
+  /** Get a nonce. */
+  get(chainId: FunctionParameters & { client: Client }): Promise<number>
+  /** Reset a nonce. */
+  reset(chainId: FunctionParameters): Promise<void>
```
tmm, line 25: Trying to [use object property syntax instead of shorthand syntax](https://www.totaltypescript.com/method-shorthand-syntax-considered-harmful) whenever possible these days.

```suggestion
  /** Get and increment a nonce. */
  consume: (parameters: FunctionParameters & { client: Client }) => Promise<number>
  /** Increment a nonce. */
  increment: (chainId: FunctionParameters) => Promise<void>
  /** Get a nonce. */
  get: (chainId: FunctionParameters & { client: Client }) => Promise<number>
  /** Reset a nonce. */
  reset: (chainId: FunctionParameters) => Promise<void>
```

src/utils/nonceManager.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Creates a nonce manager for auto-incrementing transaction nonces.
+ *
+ * @example
```
tmm, line 31:
```suggestion
 * Creates a nonce manager for auto-incrementing transaction nonces.
 *
 * - Docs: https://viem.sh/docs/accounts/createNonceManager
 *
 * @example
```

src/utils/nonceManager.ts
```diff
@@ -0,0 +1,128 @@
...
+
+type NonceManagerSource = {
+  /** Get a nonce. */
+  get(parameters: FunctionParameters & { client: Client }): MaybePromise<number>
+  /** Set a nonce. */
+  set(parameters: FunctionParameters, nonce: number): MaybePromise<void>
+}
+
+/** JSON-RPC source for a nonce manager. */
+export function jsonRpc(): NonceManagerSource {
```
tmm, line 109: Should we make this name sound more nonce manager-related or export from a nonce manager-named entrypoint? Feel like someone might confuse `jsonRpc` with a transport type and try:

```ts
import { createClient, jsonRpc } from 'viem'
const client = createClient({
  transport: jsonRpc(),
})
```

TypeScript will tell them this is wrong, but might be able to avoid some potential confusion with different names and/or moving nonce manager to a separate entrypoint (could be on accounts entrypoint along with `privateKeyToAccount` since they are used together).

src/utils/nonceManager.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+    set() {},
+  }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////
+// Default
+
+/** Default Nonce Manager with a JSON-RPC source. */
+export const nonceManager = /*#__PURE__*/ createNonceManager({
```
tmm, line 126: Either a more descriptive name or removing this and training folks to use `createNonceManager` could be good.

src/utils/nonceManager.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/** JSON-RPC source for a nonce manager. */
+export function jsonRpc(): NonceManagerSource {
+  return {
+    async get(parameters) {
+      const { address, client } = parameters
+      return getTransactionCount(client, {
+        address,
+        blockTag: 'pending',
```
holic, line 119: We've seen this as a point of potential trade-offs.

Ideally you want to use pending so you can queue up lots of transactions and have them execute in order, using the mem pool as your source of truth for how many are in the queue. But if you have one stuck transaction, you'll never be able to move past it with this logic, and instead keep queuing on top of that stuck transaction.

If you were to switch this to `latest`, you'd end up with "transaction replacement underpriced" errors if/when the nonce is reset. But handling these automatically is a little scary, especially when there may be significant consequences in terms of gas costs, etc.

How are ya'll thinking about these trade-offs?

jxom: The purpose of this PR was to implement a nonce manager, not a transaction manager (ie. handle retries, replacements, etc). Consumers will need to handle that themselves. It is also possible to pass a custom source if you don’t want to use the source of truth logic here: https://viem.sh/docs/accounts/createNonceManager#source

PR #2438, opened 2024-06-25
Add: encoding of boolean by string or other equivalent type

dbeal-eth:
This PR expands the functionaly of the `encodeAbiParameter` utility function to accept `string`, `number`, and `bigint` as encodable types for ABI `boolean`. In particular, a string with value '`false'` will be interpreted as ABI `false` as that is the likely intent of such a string input.

Currently, all other basic types of values in an ABI (uint/int, address, bytes, strings themselves) are able to be passed in as a string in Javascript, despite the types not officially allowing for this.

We recently identified a difficulty arising from this issue with our tooling, [Cannon](https://usecannon.com/). Our tool only allows interpolation of strings as a means for value injection, but we noticed that interpolation was not possible for ABI arguments which were of type `boolean`. After exploring various options for resolution, we identified that upstreaming a change seems most appropriate since the type handling for booleans seemed unusually strict compared to other types and any fixes we attempt on our end would simply be a workaround for this restriction.

Notes:
* I considered editing the type for `AbiToParameter`, to account for more broad types to match those accepted by `encodeAbiParameters` after this change, but this type is likely used by other structures (such as decodeAbiParameters) for type resolution and therefore carried a risk for side effects. Additionally, other types which accept strings or others (ex. uint) do not officially allow those types in the encode function, so I chose not to edit the type. Please let me know if it would be better to edit.
* Any non-zero numeric value will be treated as true. This most closely matches the behavior or most programming languages, but should an error be thrown if a number other than 0 or 1 is provided?
* in https://github.com/wevm/viem/issues/1960, it was reported tha-t supplying a string accidentally to boolean values would yield a NaN, so this issue was fixed by (it appears) restricting the supply of non-boolean values and adding a test case. I updated this test case to supply an invalid string (`"truez"`) which would be like a typo and fail in the same way, so the issue should not relapse.

---

## PR-Codex overview
The focus of this PR is to enhance the `boolToHex` function and its usage in encoding boolean values for ABI parameters.

### Detailed summary
- Updated `boolToHex` to handle various input types
- Improved encoding of boolean values in ABI parameters
- Added tests for encoding boolean values from strings, numbers, and BigInts

### DIFF
--- a/src/utils/abi/encodeAbiParameters.ts
+++ b/src/utils/abi/encodeAbiParameters.ts
@@ -315,13 +315,23 @@ type EncodeBoolErrorType = PadHexErrorType | BoolToHexErrorType | ErrorType
 
 function encodeBool(value: any): PreparedParam {
   if (typeof value === 'string')
-    return { dynamic: false, encoded: padHex(boolToHex(value.toLowerCase() !== 'false' && value !== '')) }
-  else if (typeof value === 'number' || typeof value === 'bigint' || typeof value === 'boolean')
+    return {
+      dynamic: false,
+      encoded: padHex(
+        boolToHex(value.toLowerCase() !== 'false' && value !== ''),
+      ),
+    }
+  
+  if (
+    typeof value === 'number' ||
+    typeof value === 'bigint' ||
+    typeof value === 'boolean'
+  )
     return { dynamic: false, encoded: padHex(boolToHex(value)) }
-  else
-    throw new BaseError(
-      `Invalid boolean value: "${value}" (type: ${typeof value}). Expected: \`true\` or \`false\`.`,
-    )
+
+  throw new BaseError(
+    `Invalid boolean value: "${value}" (type: ${typeof value}). Expected: \`true\` or \`false\`.`,
+  )
 }
 
 type EncodeNumberErrorType = NumberToHexErrorType | ErrorType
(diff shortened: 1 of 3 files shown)


src/utils/abi/encodeAbiParameters.ts
```diff
@@ -313,12 +313,25 @@ function encodeBytes<const TParam extends AbiParameter>(
...
-      `Invalid boolean value: "${value}" (type: ${typeof value}). Expected: \`true\` or \`false\`.`,
-    )
-  return { dynamic: false, encoded: padHex(boolToHex(value)) }
+function encodeBool(value: any): PreparedParam {
+  if (typeof value === 'string')
+    return {
+      dynamic: false,
+      encoded: padHex(
+        boolToHex(value.toLowerCase() !== 'false' && value !== ''),
+      ),
```
mjlescano, line 322: On the case for strings, I would only strictly accept `"false" | "" | "true"` values (without taking into acount casing). This is good if for example the user makes a typo and writes `"falsee"` it wouldn't be coerced as `true`.

dbeal-eth: changed to only accept "true" or "false" (or its uppercase variants)

dbeal-eth:
hi! wanted to inquire on this PR @jxom 

has anyone had a chance to look?

tmm:
Appreciate the PR! If you want this behavior, it's best to implement downstream at the moment as it's not the intended behavior for this feature. (Other basic types working is an unintended side effect of their data types and how they are encoded.)

tmm:
The current behavior is how we want to keep things for now.

PR #2557, opened 2024-07-29
refactor: add explicit annotations to SimulateContractReturnType

ssalbdivad:
This PR adds explicit variance annotations to `SimulateContractReturnType` based on some type performance investigation done for Mud as documented [here](https://github.com/latticexyz/mud/pull/2956).

In this case, avoiding the variance checks reduced the check time for a simple repo by over a second locally.

I based the annotations on those that were inferred from the TS compiler as documented in the trace here (from the Mud PR):

image

 Unfortunately, it is impossible to annotate a bivariant relationship directly, and treating `chain` as covariant required a `// @ts-expect-error` comment (I use this strategy somewhat often to cast variance in `arktype`, so no significant cause for concern).
 
 I don't completely understand all the contexts in which this type is used or how common it is to compare multiple instances of `SimulateContractReturnType` that would cause these variance annotations to matter, so if you find you need to make adjustments to the annotations (e.g. by labeling more parameters as `out` only with error comments), please feel free to do so.
 
It is likely if you are interested in optimizing perf, there are more scenarios like this that exist that you could find via tracing where explicit annotations could result in significant downstream perf improvements.
 
I also took the opportunity to update `@arktype/attest` to the latest version under its new primary alias, `@ark/attest`.

Let me know if you have any questions or if there is anything I can chance!

---

## PR-Codex overview
This PR updates dependencies related to testing and documentation tools. It also refactors type definitions for better clarity.

### Detailed summary
- Updated `@ark/attest` to `^0.10.2`
- Refactored type definitions in `simulateContract.ts`
- Updated various dependencies versions

> The following files were skipped due to too many changes: `pnpm-lock.yaml`


src/actions/public/simulateContract.ts
```diff
@@ -84,33 +84,34 @@ export type SimulateContractParameters<
   >
 
 export type SimulateContractReturnType<
-  abi extends Abi | readonly unknown[] = Abi,
-  functionName extends ContractFunctionName<
+  out abi extends Abi | readonly unknown[] = Abi,
```
jxom, line 87: Never knew about this syntax! Where can I find the docs?

jxom: Found: https://github.com/microsoft/TypeScript/pull/48240. nice

ssalbdivad: Yep! Here's the release notes as well:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#optional-variance-annotations-for-type-parameters

Yuripetusko: Sorry, not adding anything meaningful to the PR, but I am also only learning of this now and this is really cool. Cool to see more nuanced use of typescript and the performance comparison is really interesting too. How did you generate that graph for tsc build btw?

ssalbdivad: @Yuripetusko There's a [command](https://github.com/arktypeio/arktype/tree/main/ark/attest#trace) for it in the attest CLI.

It's a thin wrapper around `tsc --generateTrace` + https://github.com/microsoft/typescript-analyze-trace

ssalbdivad:
@tmm @jxom Is there any other information I can provide to get this merged?

It made such a big difference for @latticexyz and I'd be it could have a significant impact more broadly as well.

tmm (approved):

PR #2579, opened 2024-08-03
chore: make websocket can be used in service worker

zouhangwithsweet:
import() is disallowed on ServiceWorkerGlobalScope by the HTML specification. See https://github.com/w3c/ServiceWorker/issues/1356.

---

## PR-Codex overview
The focus of this PR is to update the WebSocket import in `webSocket.ts` file to use `isows`. 

### Detailed summary
- Updated WebSocket import to use `isows`
- Removed dynamic import for WebSocket

### DIFF
--- a/src/utils/rpc/webSocket.ts
+++ b/src/utils/rpc/webSocket.ts
@@ -1,4 +1,4 @@
-import type { MessageEvent } from 'isows'
+import { type MessageEvent, WebSocket } from 'isows'
 
 import {
   SocketClosedError,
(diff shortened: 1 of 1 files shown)


jxom:
This will break some serverless environments 😅

zouhangwithsweet:
> This will break some serverless environments 😅

So is there a better solution here? For now, I just add a patch in my repo.

jxom:
Will think of a better solution, but closing for now.

PR #2587, opened 2024-08-06
EIP-7668: empty logsBloom

SebastienGllmt:
Implement [EIP-7668](https://github.com/ethereum/EIPs/pull/8368) which states the following:

> The logs bloom of an execution block is now required to be empty (ie. 0 bytes long). The logs bloom of a transaction receipt is now required to be empty (ie. 0 bytes long).

# Point of discussion

Two main points of discussion:
- It's unclear from the original EIP how other tools should actually implement this on historical data or if there is any start point to this transition. That is to say, maybe this type change should only specify to specific `TransactionType`s. Given basically nobody uses `logsBloom` though, maybe it's not much of a concern
- It's not clear to me if `0 bytes long` means it should be of type `''` as in totally removed, or `0x0` (in which case, maybe the existing types are sufficient), or if we should make this field optional entirely and fill it to `0x0` behind the scenes

I have a feeling `0x0` is maybe the better option out of these, but part of opening this PR was for myself to get clarity on this point since I couldn't find any other discussion on EIP-7668 here

# TODO

Currently, `npm run typecheck` fails with the following error

```
src/types/window.ts:5:5 - error TS2717: Subsequent property declarations must have the same type.  Property 'ethereum' must be of type '{ on: <event extends keyof EIP1193EventMap>(event: event, listener: EIP1193EventMap[event]) => void; removeListener: <event extends keyof EIP1193EventMap>(event: event, listener: EIP1193EventMap[event]) => void; request: EIP1193RequestFn<...>; } | undefined', but here has type '{ on: <event extends keyof EIP1193EventMap>(event: event, listener: EIP1193EventMap[event]) => void; removeListener: <event extends keyof EIP1193EventMap>(event: event, listener: EIP1193EventMap[event]) => void; request: EIP1193RequestFn<...>; } | undefined'.

5     ethereum?: EIP1193Provider | undefined
      ~~~~~~~~

  src/_types/types/window.d.ts:4:9
    4         ethereum?: EIP1193Provider | undefined;
              ~~~~~~~~
    'ethereum' was also declared here.


Found 1 error in src/types/window.ts
```

That is to say, it's complaining that `src/types/window.ts` has a different type definition from `src/_types/types/window.d.ts` (which is code-generated). I tried re-running `npm run build`, but that doesn't fix the issue. Any help on this would be appreciated

---

## PR-Codex overview
The focus of this PR is to update the `logsBloom` property in `block.ts` and `transaction.ts` to allow for either a `Hex` value or an empty string.

### Detailed summary
- Updated `logsBloom` property in `block.ts` and `transaction.ts` to accept `Hex` or `''`
- Updated type assertions in `getBlock.test-d.ts` for `logsBloom` to reflect the new type flexibility


src/types/block.ts
```diff
@@ -31,7 +31,7 @@ export type Block<
   /** Block hash or `null` if pending */
   hash: blockTag extends 'pending' ? null : Hash
   /** Logs bloom filter or `null` if pending */
-  logsBloom: blockTag extends 'pending' ? null : Hex
+  logsBloom: blockTag extends 'pending' ? null : Hex | ''
```
jxom, line 34: I would assume that RPC Providers / Nodes will coerce zero data as undefined as they do with other fields. 

```suggestion
  logsBloom?: Hex | null | undefined
```

src/types/transaction.ts
```diff
@@ -62,7 +62,7 @@ export type TransactionReceipt<
   /** List of log objects generated by this transaction */
   logs: Log<quantity, index, false>[]
   /** Logs bloom filter */
-  logsBloom: Hex
+  logsBloom: Hex | ''
```
jxom, line 65: per above 

```suggestion
  logsBloom?: Hex | undefined
```

jxom:
EIP is still very early. Wonder if it's worth parking this PR for now.

jxom:
Closing for now. Will reopen in future!

PR #2623, opened 2024-08-16
fix: default nonce manager resets previous nonce

holic:
I am using the default nonce manager to trigger a bunch of transactions in a queue to ensure they land in the mempool in order. If any transaction fails to submit to the mempool, I call `nonceManager.reset()`.

For example:

```
tx:1 -> nonce:1
tx:2 -> nonce:2
tx:3 -> nonce:3
```

If `tx:1` fails, I call `nonceManager.reset()` and retry it. However, the logic that looks up the nonce after reset also takes into account the previously cached nonce in the nonce manager:

https://github.com/wevm/viem/blob/5f350ecf63b2da861c8dcb60e77fbcae3e7192ab/src/utils/nonceManager.ts#L79-L84

Since the queue is blocked on `tx:1`, our nonce at the source should still be `nonce:1`. But because of the logic above, the nonce returned will actually be `nonce:4` (because of our previously queued txs).

IMO `.reset` should reset the state of the nonce manager back to the source, including the cached nonce value.

---

## PR-Codex overview
The focus of this PR is to add tests for the `nonceManager` reset functionality and ensure correct nonce handling during transactions.

### Detailed summary
- Added a test for `nonceManager` reset functionality
- Implemented handling of nonce increment and reset during transactions

### DIFF
--- a/src/utils/nonceManager.test.ts
+++ b/src/utils/nonceManager.test.ts
@@ -258,3 +258,56 @@ test('dropped tx', async () => {
   })
   expect(await nonceManager.get(args)).toBe(10)
 })
+
+test('nonceManager reset', async () => {
+  const nonceManager = createNonceManager({
+    source: jsonRpc(),
+  })
+
+  const args = {
+    address: account.address,
+    chainId: mainnetClient.chain.id,
+    client: mainnetClient,
+  }
+
+  const nonce_1 = await nonceManager.consume(args)
+  expect(nonce_1).toBe(10)
+  const hash_1 = await sendTransaction(mainnetClient, {
+    account,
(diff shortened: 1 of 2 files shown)


holic:
Just noticed this line:
https://github.com/wevm/viem/blob/5f350ecf63b2da861c8dcb60e77fbcae3e7192ab/src/utils/nonceManager.ts#L85-L87

I think this means I need a bit deeper of a change. Will play with this more locally and report back with an alternative.

holic:
Added a test to demonstrate the expected behavior.

jxom:
closing for now, will reopen when we visit in near future

PR #2627, opened 2024-08-17
refactor: use getAction internally

holic:
As an author of custom actions and action overrides, I will use viem actions internally but sometimes forget to call them `getAction` so that the actions can compose on one another.

For example, maybe I'll use `getChainId` internally in my custom action, but I want this to be composable with an action that overrides `getChainId` to cache the RPC's chain ID. If I don't call this with `getAction`, I'll accidentally bypass this cache and viem will do the underlying RPC call.

I've seen this pattern in both in my code, other SDKs, and within viem itself.

I am curious if, instead of requiring all action authors to use `getAction`, that viem itself exports by default the `getAction`-wrapped variant of each of these actions, while also exporting the internal implementation in case it's useful to make an explicit, direct call.

I'm mocking this up here to see if this approach seems reasonable before doing more because its quite a big refactor.

---

## PR-Codex overview
This PR focuses on refactoring functions in various action files to directly call `getBlock`, `getBlockNumber`, and `getChainId` functions instead of using `getAction`.

### Detailed summary
- Replaced `getAction` calls with direct calls to `getBlock`, `getBlockNumber`, and `getChainId` functions in multiple files.
- Updated function signatures and parameters accordingly.
- Removed unnecessary `getAction` imports from specific files.

### DIFF
--- a/src/actions/public/estimateFeesPerGas.ts
+++ b/src/actions/public/estimateFeesPerGas.ts
@@ -25,7 +25,7 @@ import {
   type EstimateMaxPriorityFeePerGasErrorType,
   internal_estimateMaxPriorityFeePerGas,
 } from './estimateMaxPriorityFeePerGas.js'
-import { getBlock } from './getBlock.js'
+import { type GetBlockErrorType, getBlock } from './getBlock.js'
 import { type GetGasPriceErrorType, getGasPrice } from './getGasPrice.js'
 
 export type EstimateFeesPerGasParameters<
@@ -52,6 +52,7 @@ export type EstimateFeesPerGasReturnType<
 
 export type EstimateFeesPerGasErrorType =
   | BaseFeeScalarErrorType
+  | GetBlockErrorType
   | EstimateMaxPriorityFeePerGasErrorType
   | GetGasPriceErrorType
   | Eip1559FeesNotSupportedErrorType
@@ -126,9 +127,7 @@ export async function internal_estimateFeesPerGas<
     (base * BigInt(Math.ceil(baseFeeMultiplier * denominator))) /
     BigInt(denominator)
 
-  const block = block_
-    ? block_
-    : await getAction(client, getBlock, 'getBlock')({})
+  const block = block_ ? block_ : await getBlock(client)
 
   if (typeof chain?.fees?.estimateFeesPerGas === 'function') {
     const fees = (await chain.fees.estimateFeesPerGas({
--- a/src/actions/public/estimateMaxPriorityFeePerGas.ts
+++ b/src/actions/public/estimateMaxPriorityFeePerGas.ts
@@ -30,7 +30,6 @@ export type EstimateMaxPriorityFeePerGasErrorType =
   | GetBlockErrorType
   | HexToBigIntErrorType
   | RequestErrorType
-  | GetBlockErrorType
   | GetGasPriceErrorType
   | Eip1559FeesNotSupportedErrorType
   | ErrorType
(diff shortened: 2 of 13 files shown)


src/actions/public/estimateMaxPriorityFeePerGas.ts
```diff
@@ -30,7 +30,6 @@ export type EstimateMaxPriorityFeePerGasErrorType =
   | GetBlockErrorType
   | HexToBigIntErrorType
   | RequestErrorType
-  | GetBlockErrorType
```
holic, line 33: removed duplicate union

src/actions/public/estimateFeesPerGas.ts
```diff
@@ -52,6 +52,7 @@ export type EstimateFeesPerGasReturnType<
 
 export type EstimateFeesPerGasErrorType =
   | BaseFeeScalarErrorType
+  | GetBlockErrorType
```
holic, line 55: this was missing from the error type even though we use `getBlock` internally

src/actions/public/getBlockNumber.ts
```diff
@@ -52,6 +53,13 @@ export function getBlockNumberCache(id: string) {
  * // 69420n
  */
 export async function getBlockNumber<chain extends Chain | undefined>(
+  client: Client<Transport, chain>,
+  params: GetBlockNumberParameters = {},
+): Promise<GetBlockNumberReturnType> {
+  return getAction(client, _getBlockNumber, 'getBlockNumber')(params)
+}
+
+export async function _getBlockNumber<chain extends Chain | undefined>(
```
roninjin10, line 62: Is it actually necessary to export both?  Shouldn't getBlockNumber always "just work"?

roninjin10 (commented):
My worry here is if this pattern is expected, that means third party actions will also need to remember to implement it. 

Maybe this is a unimportant because a third party action is less likely to be overriden.

Maybe a solution to that would be to encourage folks to build actions using a `defineAction(...)` utility or something?

src/actions/public/getBlock.ts
```diff
@@ -92,6 +93,18 @@ export async function getBlock<
   account extends Account | undefined,
   includeTransactions extends boolean = false,
   blockTag extends BlockTag = 'latest',
+>(
+  client: Client<Transport, chain, account>,
+  params: GetBlockParameters<includeTransactions, blockTag> = {},
+): Promise<GetBlockReturnType<chain, includeTransactions, blockTag>> {
+  return getAction(client, _getBlock, 'getBlock')(params)
+}
```
roninjin10, line 101:
```suggestion
>(
  client: Client<Transport, chain, account>,
  params: GetBlockParameters<includeTransactions, blockTag> = {},
): Promise<GetBlockReturnType<chain, includeTransactions, blockTag>> {
  return getAction(client, _getBlock, 'getBlock')(params)
}
```

Might be able to do this just to save some boilerplate

```typescript
const withGetAction = <TAction>(action: TAction, name: string): TAction => {
  return ((client, params) => {
     return getAction(client, action, name)(params)
  }) as unknown as TAction
}

export const getBlock = withGetAction(client, action, 'getBlock')
```

roninjin10: DOn't do this it's convenient but will make stack traces suck

src/actions/public/getBlockNumber.ts
holic: My understanding is if I override `getBlockNumber` with my own action, then inside that override call into viem's `getBlockNumber` here, it'd end up resolving to my own `getBlockNumber`, not viem's underlying one, because that's the one attached to the client object?

holic:
> means third party actions will also need to remember to implement it

this is already true for `getAction` usage, which isn't correctly used in most cases

holic:
Another thing I am realizing that I don't think I realized before: viem clients only support _one_ instance of a given action, rather than a stack of actions (like middleware).

So if I have two decorators that each override `getChainId`, only the most recently used one (called via `.extend`) will be attached to the client and called by all other actions?

jxom:
closing for now – will revisit later

PR #2651, opened 2024-08-23
🐛 utils: fix strict address check

cruzdanilo:
# issue

`isAddress` strict check is skipped when the input is an address entirely lowercase.

# solution

remove lowercase comparison.

# impact

although this fix doesn't change the _expected_ result, it is actually a breaking change for many projects (and many tests in this project), so i will leave the decision to the core maintainers.

---

## PR-Codex overview
This PR updates the `isAddress` utility function to handle case sensitivity and strict mode for Ethereum addresses.

### Detailed summary
- Fixed case sensitivity check in `isAddress` function
- Added strict mode option to `isAddress` function
- Updated test cases for `isAddress` function to reflect changes


iosh:
As you said, this will be a big change, maybe we can add some comments or document to this function to tell user that lowercase address will not be checksum check and return true

jxom:
The behavior we have now is expected. The strict behavior checks for casing mismatches in the checksum (ie. end user may not encounter a memory bit flip, or type an incorrect character) and not the fully lowercased address. Maybe we should document it better.

cruzdanilo:
how should we check if a `0x{string}` is a checksummed address?

iosh:
> how should we check if a `0x{string}` is a checksummed address?

How about this
```ts
import { checksumAddress } from "viem";
const addr = '0x....'
const isChecksumAddress = checksumAddress(addr) === addr;
```

PR #2680, opened 2024-08-31
test: demonstrate callstack in extended client action

holic:
Continued from https://github.com/wevm/viem/pull/2675

- replaces base actions to avoid RPC calls
- demonstrates callstack of inner action calls

---

## PR-Codex overview
The focus of this PR is to refactor and extend action composition in the `createClient` module.

### Detailed summary
- Added type definition for `Client`
- Refactored action composition logic in `createClient`
- Added new actions `getChainId` and `estimateGas`
- Updated tests to reflect changes in action composition logic

### DIFF
--- a/src/clients/createClient.test.ts
+++ b/src/clients/createClient.test.ts
@@ -5,7 +5,7 @@ import { getChainId } from '../actions/public/getChainId.js'
 import { localhost, mainnet } from '../chains/index.js'
 import type { EIP1193RequestFn, EIP1474Methods } from '../types/eip1193.js'
 import { getAction } from '../utils/getAction.js'
-import { createClient } from './createClient.js'
+import { type Client, createClient } from './createClient.js'
 import { publicActions } from './decorators/public.js'
 import { createTransport } from './transports/createTransport.js'
 import { custom } from './transports/custom.js'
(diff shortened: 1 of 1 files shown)


src/clients/createClient.test.ts
```diff
@@ -577,29 +577,62 @@ describe('extends', () => {
...
+    ])
+
+    calls = []
+    expect(await extended.estimateGas()).toBe(1000n)
+    expect(calls).toEqual([
+      'estimateGas:second',
+      'estimateGas:first',
+      'estimateGas:base',
+      'getChainId:base',
+    ])
```
holic, line 636: This is the functionality I was trying to understand, where I hoped that this would look more like
```ts
[
  'estimateGas:second',
  'estimateGas:first',
  'estimateGas:base',
  'getChainId:second',
  'getChainId:first',
  'getChainId:base',
]
```

jxom: Could consider binding them to behave like this.

holic: I'm not sure of the consequences of that. Would that be a breaking change because it would change the behavior of the entire callstack of extended clients?

I was thinking of writing a separate "compose" decorator wrapper that could handle this.

PR #2693, opened 2024-09-05
Fix LRU cache not updating touched keys

kyscott18:
The LRU algorithm in `utils/lru.ts` does not update touched keys, it instead is acting like a "first recently used" algorithm.

---

## PR-Codex overview
This PR focuses on fixing the LRU algorithm in `LruMap` by updating touched keys.

### Detailed summary
- Updated `LruMap` to properly update touched keys in the LRU algorithm.
- Added an `override get` method to handle key updates.
- Added a test case to ensure touched keys are updated correctly.


src/utils/lru.ts
```diff
@@ -11,6 +11,17 @@ export class LruMap<value = unknown> extends Map<string, value> {
     this.maxSize = size
   }
 
+  override get(key: string) {
+    const value = super.get(key)
+
+    if (value !== undefined) {
```
holic, line 17: curious if this should use `super.has(key)` instead to account for `LruMap<undefined>`?

jxom: Yeah probably better

PR #2714, opened 2024-09-11
Upgrade alfajores to op

aaronmgdr:
Alfajores will be upgraded from an L1 to an OP L2 at end of the month

This PR adds the relevant contract addresses to the chain definition 

(pending review by team)

---

## PR-Codex overview
This PR updates the Celo Alfajores chain configuration to include new contract addresses.

### Detailed summary
- Updated Celo Alfajores chain with new contract addresses
- Added contracts: portal, disputeGameFactory, l2OutputOracle, l1StandardBridge, l2StandardBridge

### DIFF
--- a/src/celo/chainConfig.ts
+++ b/src/celo/chainConfig.ts
@@ -1,7 +1,7 @@
+import { contracts } from '../op-stack/contracts.js'
 import { fees } from './fees.js'
 import { formatters } from './formatters.js'
 import { serializers } from './serializers.js'
-import { contracts } from '../op-stack/contracts.js'
 
 export const chainConfig = {
   contracts,
--- a/src/celo/chainConfig.ts
+++ b/src/celo/chainConfig.ts
@@ -1,8 +1,10 @@
 import { fees } from './fees.js'
 import { formatters } from './formatters.js'
 import { serializers } from './serializers.js'
+import { contracts } from '../op-stack/contracts.js'
 
 export const chainConfig = {
+  contracts,
   formatters,
   serializers,
   fees,
--- a/src/chains/definitions/celoAlfajores.ts
+++ b/src/chains/definitions/celoAlfajores.ts
@@ -1,7 +1,7 @@
 import { chainConfig } from '../../celo/chainConfig.js'
 import { defineChain } from '../../utils/chain/defineChain.js'
 
-const sourceId = 11_155_111 // sepolia 
+const sourceId = 11_155_111 // sepolia
 
 export const celoAlfajores = /*#__PURE__*/ defineChain({
   ...chainConfig,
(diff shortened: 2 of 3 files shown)


src/chains/definitions/celoAlfajores.ts
```diff
@@ -21,12 +23,43 @@ export const celoAlfajores = /*#__PURE__*/ defineChain({
       url: 'https://explorer.celo.org/alfajores',
       apiUrl: 'https://explorer.celo.org/api',
     },
+    celoscan: {
+      name: 'CeloScan',
+      url: 'https://alfajores.celoscan.io/',
+      apiUrl: 'https://api-alfajores.celoscan.io/api',
+    },
```
tmm, line 30: `blockExplorers` isn't meant to be exhaustive and we only support one per chain so can either make this the default or remove it.

aaronmgdr: cool

aaronmgdr: the type definition allows for other arbitrary keys, is the idea that users can add their own in in the app they are building?

src/celo/chainConfig.ts
```diff
@@ -1,8 +1,10 @@
+import { contracts } from '../op-stack/contracts.js'
```
aaronmgdr, line 1: for easier review this is 

```typescript 

const contracts = {
  gasPriceOracle: { address: '0x420000000000000000000000000000000000000F' },
  l1Block: { address: '0x4200000000000000000000000000000000000015' },
  l2CrossDomainMessenger: {
    address: '0x4200000000000000000000000000000000000007',
  },
  l2Erc721Bridge: { address: '0x4200000000000000000000000000000000000014' },
  l2StandardBridge: { address: '0x4200000000000000000000000000000000000010' },
  l2ToL1MessagePasser: {
    address: '0x4200000000000000000000000000000000000016',
  },
} 
```

src/chains/definitions/celoAlfajores.ts
tmm: exactly!

aaronmgdr:
@jcortejoso please review addresses

aaronmgdr:
@tmm I based the contracts on what base / op include in their config. there isn't any reason to have additional contract addresses defined right? 

slightly related but would it be correct to add additional celo core contracts to this config?

tmm:
> there isn't any reason to have additional contract addresses defined right?

Correct, unless they are used internally in Viem. (Can search `chain.contracts` to see if you missed any.)

> slightly related but would it be correct to add additional celo core contracts to this config?

You are welcome to add additional ones that you know will be used downstream, but similar to block explorers/RPC urls, we try to keep things lean when we can!

Issue #2796, opened 2024-09-30
prepareUserOperation fails when Chain is not passed as an arguement


AyushBherwani1998:
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.15

### Current Behavior

When creating `createSmartAccountClient` from permissionless-js, the `chain` is an optional parameter.
```
const smartAccountClient = createSmartAccountClient({
      account: smartAccount,
      bundlerTransport: http(BUNDLER_URL),
      paymaster: bundlerClient,
      // chain: arbitrumSepolia,
      userOperation: {
        estimateFeesPerGas: async () => (await bundlerClient.getUserOperationGasPrice()).fast,
      }
    });
```

If we try to sign the user operation when paymaster is also present, it throws an error that chain id is not present, and the logs are from prepareUserOperation. Checking from the code, chain is a nullable parameter, but we are not doing null checks, and force casting to non nullable. 

### Expected Behavior

Ideally we should check if the chain is nullable, and accordingly get the chain id. Currently chain is nullable, but it won't work if we don't pass the chain. 

### Steps To Reproduce

You can try to create a SmartAccountClient with paymaster, and no chain, it'll throw an error while signing the transaction.

jxom:
Fixed via https://github.com/wevm/viem/commit/705eb21b18f9993abbfd666c48de8c588ab86822. Added you as a co-author!

PR #2797, opened 2024-09-30
Fixes prepareUserOperation

AyushBherwani1998:
Fixes: #2796

---

## PR-Codex overview
This PR addresses the `prepareUserOperation` function by fixing an issue where `chainId` could be null if the `Chain` was not provided. It ensures that the `chainId` is correctly retrieved from `bundlerClient` when necessary.

### Detailed summary
- Introduced a check for `client.chain` to set `chainId`.
- If `client.chain` is undefined, `chainId` is fetched using `bundlerClient.getChainId()`.
- Updated references to `chainId` in the `getPaymasterStubData` and `getPaymasterData` calls.

### DIFF
--- a/.changeset/cuddly-rats-think.md
+++ b/.changeset/cuddly-rats-think.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixes prepareUserOperation issue of chainId being null when Chain is not passed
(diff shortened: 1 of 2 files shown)


AyushBherwani1998:
@jxom looks like test are failing, should I get the client, and use the getActions to get the chain id?

PR #2798, opened 2024-09-30
Alfajores OP contracts config

aaronmgdr:
Celo's Alfajores testnet is now an OP based chain. Add these contracts to the celoAlfajores config and update the blockexplorer to the new and improved blockscout.


src/chains/definitions/optimismGoerli.ts
```diff
@@ -34,11 +35,13 @@ export const optimismGoerli = /*#__PURE__*/ defineChain({
     portal: {
       [sourceId]: {
         address: '0x5b47E1A08Ea6d985D6649300584e6722Ec4B1383',
+        blockCreated: 26384000
```
aaronmgdr, line 38: wait how did this happen

shazarre (approved):

aaronmgdr:
chain tests seem to be because of other chains / servers

Issue #2484, opened 2024-07-07
Public action `verifyHash` does not work with zksync public client


tahos81:
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.16.5

### Current Behavior

When doing anything that uses `verifyHash` internally (like `verifySiweMessage`) if the public client in context is a zksync client verification fails. Reason for that is `verifyHash` uses erc-6492 for signature verification and has a constant called `universalSignatureValidatorByteCode` but regular bytecodes does not work on zksync so verification fails even when the signature is valid. This is not a problem for eoa's (you can just verify with an arbitrary public client) but it is breaking for smart contract accounts

### Expected Behavior

Either a different constant bytecode should be used when the chain in context is zksync or 6492 flow should not be used and verification flow should revert to old flow (1271 + ecrecover).

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/viem-getting-started-shqa8r?file=index.ts

jxom:
Unfortunately, zkSync handles contract deployments a bit differently to Ethereum. :\ 

Can see what I can do.

jxom:
Added a fallback for this scenario here: https://github.com/wevm/viem/commit/ec3b0e08f0a379249224e65092277191f1a8502b.

PR #2807, opened 2024-10-02
add ability to call deployed universal signature validator for chains that do not support createless call

coffeexcoin:
Creates a new property on `chain.contracts` for `universalSignatureVerifier`. Calling the `verifyHash` action will prefer this deployed address over a createless deploy call. This is useful for chains that do not support createless calls (zk stack)

Added deployed version of universal signature verifier to zksync, zksync sepolia, and abstract mainnet.
 

---

## PR-Codex overview
This PR introduces a new `universalSignatureVerifier` feature across multiple files, implementing a signature verification mechanism and related deployment functions. It removes unused code and adds tests to ensure proper functionality.

### Detailed summary
- Added `universalSignatureVerifier` property in `chain.ts` and various chain definitions.
- Implemented `isValidSig` function in `VerifySig` contract for signature validation.
- Created `deployUniversalSignatureVerifier` function for deploying the verifier.
- Updated tests in `verifyHash.test.ts` to include scenarios for the new verifier.
- Removed `isBytesEqual.ts` and associated tests as they were no longer needed.
- Modified `verifyHash.ts` to integrate the new verifier into the verification process.


coffeexcoin:
This is related to #2484 as zksync is one of the chains that does not support createless deployment. This will add full support for ERC-6942 and ERC-1271 validation to zksync

rookmate (approved):

contracts/src/UniversalSigValidator.sol
jxom, line 1: Curious if the deployed contracts should use https://github.com/AmbireTech/signature-validator/blob/main/contracts/EIP6492Full.sol instead of this? If so, can we revert these changes?

coffeexcoin: I think ideally the onchain logic would be identical between the deployed contract and the deployless contract.

These seem to be pretty much functionally identical for the purpose of validation.

An alternative would be to use the `EIP4692Full` contract as the deployed validator and its associated `ValidateSigOffchain` deployless call helper as the deployless version used by viem, effectively replacing the current version in viem.

Let me know if you agree on one of these approaches as preferable, otherwise I can just revert the change to the contracts - the abi for these is the same regardless of approach.

coffeexcoin: Hey @jxom just wanted to follow up on this and see if you had a preferred path forward.

jxom: Sounds okay. Can we change the name from `isValidUniversalSig` to `isValidSig`?

coffeexcoin:
I don't think these HTTP test failures seem to be related to my changes - are you able to confirm that?

src/chains/definitions/zksync.ts
```diff
@@ -33,5 +33,9 @@ export const zksync = /*#__PURE__*/ defineChain({
     multicall3: {
       address: '0xF9cda624FBC7e059355ce98a31693d299FACd963',
     },
+    universalSignatureVerifier: {
+      address: '0x872146211f996755C8729042093ffb8660F8b129',
```
jxom, line 37: Can we get these contracts verified please?

coffeexcoin: Updated deployments for the new contract with the new function signature
Contracts are verified:
[zksync mainnet](https://explorer.zksync.io/address/0xfB688330379976DA81eB64Fe4BF50d7401763B9C#contract)
[zksync sepolia](https://sepolia.explorer.zksync.io/address/0xfB688330379976DA81eB64Fe4BF50d7401763B9C#contract)
[abstract testnet](https://explorer.testnet.abs.xyz/address/0xfB688330379976DA81eB64Fe4BF50d7401763B9C#contract)

PR #2811, opened 2024-10-03
Fix Celo maxFeePerGas calculation for fee currency transactions

shazarre:
Previously the multiply function was not applied to the base fee returned from the node. This PR fixes that.

---

## PR-Codex overview
This PR focuses on fixing the calculation of `maxFeePerGas` for fee currency transactions in the Celo network, ensuring accurate gas fee estimations and adjustments in tests.

### Detailed summary
- Updated the calculation of `maxFeePerGas` in `src/celo/fees.ts` to use `suggestedMaxFeePerGas`.
- Adjusted mock return values for gas price and priority fee in `src/celo/fees.test.ts`.
- Changed expected values in the snapshot test for fees.
- Modified return values for gas price in `src/celo/sendTransaction.test.ts`.

### DIFF
--- a/src/celo/fees.test.ts
+++ b/src/celo/fees.test.ts
@@ -30,8 +30,8 @@ describe('celo/fees', () => {
     const requestMock = vi.spyOn(client, 'request')
     // @ts-ignore
     requestMock.mockImplementation((request) => {
-      if (request.method === 'eth_gasPrice') return '11619349802'
-      if (request.method === 'eth_maxPriorityFeePerGas') return '2323869960'
+      if (request.method === 'eth_gasPrice') return '15057755162'
+      if (request.method === 'eth_maxPriorityFeePerGas') return '602286'
       return
     })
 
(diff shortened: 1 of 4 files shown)


src/celo/fees.test.ts
```diff
@@ -30,24 +30,25 @@ describe('celo/fees', () => {
     const requestMock = vi.spyOn(client, 'request')
     // @ts-ignore
     requestMock.mockImplementation((request) => {
-      if (request.method === 'eth_gasPrice') return '11619349802'
-      if (request.method === 'eth_maxPriorityFeePerGas') return '2323869960'
+      if (request.method === 'eth_gasPrice') return '15057755162'
+      if (request.method === 'eth_maxPriorityFeePerGas') return '602286'
```
jeanregisser, line 34: nit: Real requests to these RPC methods returns Hex strings.

```
❯ curl --silent -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_gasPrice","params":[],"id":1}' https://forno.celo.org | jq
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0x2540be400"
}
```

I guess here it still works because we parse them with `BigInt(value)`. And Hex strings are accepted.

PR #2820, opened 2024-10-03
cache address in toCoinbaseSmartWallet

holic:
The address in `toCoinbaseSmartAccount` is derived from args that can't be changed later, so we may as well cache this value rather than fetching it repeatedly.

Alternatively, we could overwrite `getAddress` in `toSmartAccount` with a function that returns the address since its fetched immediately. But unclear if the intent of this is to be always deterministic. Also unclear if `this.getAddress()` would be scoped to the object from `toCoinbaseSmartWallet` or `toSmartAccount`.

---

## PR-Codex overview
This PR modifies the handling of the `address` property within the `ToCoinbaseSmartAccountParameters` in the `toCoinbaseSmartAccount.ts` file. It changes how `address` is destructured from `parameters`, shifting from destructuring to a separate assignment.

### Detailed summary
- Removed `address` from destructuring in the parameters.
- Introduced a separate assignment for `address` from `parameters`.

### DIFF
--- a/src/account-abstraction/accounts/implementations/toCoinbaseSmartAccount.ts
+++ b/src/account-abstraction/accounts/implementations/toCoinbaseSmartAccount.ts
@@ -70,7 +70,8 @@ export type CoinbaseSmartAccountImplementation = Assign<
 export async function toCoinbaseSmartAccount(
   parameters: ToCoinbaseSmartAccountParameters,
 ): Promise<ToCoinbaseSmartAccountReturnType> {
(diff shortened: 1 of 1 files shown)


holic:
@jxom after your changes, this doesn't seem to be doing anything anymore?

jxom:
whoops

jxom:
fixed: https://github.com/wevm/viem/commit/44ae1a5549e704486f2643c67f222ab3ce87cd47

PR #2860, opened 2024-10-14
fix: update `setTimeout` return type in `socket.ts`

vpontis:
When we install Viem we get a Typescript issue because `Timer` is not defined in some environments. Instead, we can look at the return type for `setTimeout` so that this works well across Node and web environments.

---

## PR-Codex overview
This PR focuses on updating the type of the `keepAliveTimer` variable in the `src/utils/rpc/socket.ts` file to improve type safety and clarity.

### Detailed summary
- Updated the type of `keepAliveTimer` from `Timer | undefined` to `ReturnType<typeof setInterval> | undefined`.
- Added a note in the `.changeset/small-pianos-agree.md` file indicating the patch for `viem` and the update to the internal `keepAliveTimer` type.

### DIFF
--- a/.changeset/small-pianos-agree.md
+++ b/.changeset/small-pianos-agree.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Updated internal `keepAliveTimer` type.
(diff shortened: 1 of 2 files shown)


vpontis:
This project is a bit confused between `pnpm` and `bun`. `Timer` is a bun-only type and in the screenshot below you can see the project says "pnpm" but links to "bun"...

CleanShot 2024-10-14 at 16 06 20@2x

vpontis:
Thank you for merging this in so quickly. Could you publish a version with this fix in?

PR #2928, opened 2024-10-25
feat: Improve return type for toSignature

SebastienGllmt:
Integrates https://github.com/wevm/abitype/pull/255 into Viem to improve the return type of `toSignature`

This helps resolve these two discussions: https://github.com/wevm/viem/discussions/260 and https://github.com/wevm/viem/discussions/2706

A follow-up PR to this will have to be adding the option to pass the full signature to functions like `getContractEvents` and `getAbiItem`

---

## PR-Codex overview
This PR focuses on improving the type safety and return types of the `toSignature` and `normalizeSignature` functions in the codebase, enhancing the handling of ABI signatures.

### Detailed summary
- Updated `toSignature` to use generics for better return types.
- Enhanced `normalizeSignature` with stricter type definitions.
- Added detailed comments addressing limitations and expected behavior of signature normalization.
- Introduced new tests to validate the changes in type handling.

### DIFF
--- a/src/utils/hash/normalizeSignature.ts
+++ b/src/utils/hash/normalizeSignature.ts
@@ -1,13 +1,44 @@
+import type {
+  AbiEvent,
+  AbiFunction,
+  ParseAbiItem,
+  SignatureAbiItem,
+} from 'abitype'
 import { BaseError } from '../../errors/base.js'
 import type { ErrorType } from '../../errors/utils.js'
 
-type NormalizeSignatureParameters = string
-type NormalizeSignatureReturnType = string
+/**
+ * TODO: ParseAbiItem<Signature> is too strict here since
+ *       1. normalizeSignature allows loose formatting (ex: spacing)
+ *       2. normalization doesn't require the string to conform to abitype's human-readable encoding
+ *       ex: 'function  bar( )' should still results in `bar()` as the result
+ *       ex: `event foo(tuple(uint16, uint16))` results in a broken type response
+ *       for now, we return`string` in cases where abitype fails
+ *                but this doesn't handle error cases where abitype is just wrong like
+ * TODO: Some obvious errors get the return type `string`
+ *       this is because we fallback to `string` if `ParseAbiItem<Signature>` fails
+ *       but it could fail because not because it's too strict, but because the string is plain incorrect
+ *       ex: normalizeSignature('bar') falls back to `string` as the response type
+ *           since it can't tell if this is wrong since `ParseAbiItem<Signature>` is too strict or if it really is invalid
+ * TODO: Some normalizations are missing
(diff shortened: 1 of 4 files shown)


src/utils/hash/normalizeSignature.ts
```diff
@@ -1,13 +1,44 @@
...
+  AbiFunction,
+  ParseAbiItem,
+  SignatureAbiItem,
+} from 'abitype'
 import { BaseError } from '../../errors/base.js'
 import type { ErrorType } from '../../errors/utils.js'
 
-type NormalizeSignatureParameters = string
-type NormalizeSignatureReturnType = string
+/**
```
SebastienGllmt, line 10: There are multiple TODOs here and I'm not sure which one of these would have to be fixed to be considered good enough to merge (given that any type is better than no type, but the wrong type is also worse than no type)

I wanted to make the original commit for this PR contain as detailed information before we start truncating out cases to support the subset that makes sense

Here are the 4 options in increase level of work required (each option requires the previous option to be completed):

Option 1: we can make that the `toSignature` function only returns a type for `AbiFunction | AbiEvent` (since those cases are always correct) and then just return `string` otherwise (instead of trying to parse the string)

Option 2: we make `string` inputs work in some cases, but not all. This would require:
a. fixing the normalization of `uint` to `uint256` (this is a bug in `normalizeSignature` and not in the type)
b. arguably, fixing the `tuple(...)` notation handling (but if the JS code is also wrong, no point in making the type match it)

Option 3: we try to make all `string` inputs work. This would require:
a. fixing `tuple(uint)` handling (maybe this is an abitype concern? However, this syntax isn't part of their human-readable string encoding)
b. adding proper trimming logic on whitespace

Option 4: make all `string` inputs work, and properly return `never` on invalid input

src/utils/hash/normalizeSignature.ts
```diff
@@ -1,13 +1,44 @@
+import type {
+  AbiEvent,
+  AbiFunction,
+  ParseAbiItem,
+  SignatureAbiItem,
```
SebastienGllmt, line 5: note: this type (`SignatureAbiItem`) depends on https://github.com/wevm/abitype/pull/255 getting merged and release

SebastienGllmt:
Anything at all I can do to help this move forward?

tmm:
Let's use [`AbiItem.getSignature`](https://oxlib.sh/api/AbiItem/getSignature) from [Ox](https://github.com/wevm/ox).

@jxom and I are going to do a larger refactor in the new year that moves a lot of Viem's internals over to Ox, but can proxy from Ox for now.

SebastienGllmt:
Just to make sure we're clear, `ox` also returns just `string` as the type of `getSignature` so it runs into the same issue this PR is trying to solve

jxom:
We can move this PR to Ox. Feel free to raise a PR there, or we will do it when we get some time.

PR #2954, opened 2024-11-01
Make `BaseError#walk` tolerant to an `undefined` cause

alcuadrado:
I had an issue where I was setting `Error.cause` to `undefined`, and `viem` was giving me some errors because it was trying to access an `undefined`. This PR fixes it.

---

## PR-Codex overview
This PR focuses on improving the `BaseError#walk` method to handle cases where the `cause` property is `undefined`, ensuring it doesn't throw errors in such scenarios.

### Detailed summary
- Modified the logic in `BaseError#walk` to check if `err.cause` is not `undefined` before recursively calling `walk`.
- Added a test case to verify that `walk` correctly handles an error with an `undefined` cause.

### DIFF
--- a/src/errors/base.ts
+++ b/src/errors/base.ts
@@ -87,7 +87,7 @@ function walk(
   fn?: ((err: unknown) => boolean) | undefined,
 ): unknown {
   if (fn?.(err)) return err
-  if (err && typeof err === 'object' && 'cause' in err)
+  if (err && typeof err === 'object' && 'cause' in err && err.cause !== undefined)
     return walk(err.cause, fn)
   return fn ? null : err
 }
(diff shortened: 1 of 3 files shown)


src/errors/base.ts
```diff
@@ -87,7 +87,7 @@ function walk(
   fn?: ((err: unknown) => boolean) | undefined,
 ): unknown {
   if (fn?.(err)) return err
-  if (err && typeof err === 'object' && 'cause' in err)
+  if (err && typeof err === 'object' && 'cause' in err && err.cause !== undefined)
```
alcuadrado, line 90: I'm pretty sure this doesn't pass the linter, but I had some installation issues. Sorry about that.

PR #3103, opened 2024-12-06
fix: export `hashStruct`

iamacook:
## Summary

It is not possible to import `hashStruct`. This method is useful should a contract not follow EIP-712.

This exports `hashStruct` and the associated `HashStructErrorType`.

Note: there is also an [open discussion requesting this change](https://github.com/wevm/viem/discussions/761).

## Changes

- Export `HashStructErrorType`
- Import `hashStruct` and `HashStructErrorType` in both index and `/utils`

### DIFF
--- a/.changeset/cool-tips-lick.md
+++ b/.changeset/cool-tips-lick.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Exported `hashStruct`.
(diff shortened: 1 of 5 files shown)


jxom:
Feel free to check out Ox for lower-level utilities too: https://oxlib.sh/api/TypedData/hashStruct

PR #3138, opened 2024-12-18
Adding in Ink Mainnet Network Information

alexanderbradshaw:
Add the Ink Mainnet network information so devs can build on top of Ink


pnpm-lock.yaml
jxom, line 1: please revert this change

jxom:
chain also needs to be added to `src/chains/index.ts`

jxom:
please reopen when comments are addressed

thINKoor:
Addressed the changes in the following PR: https://github.com/wevm/viem/pull/3141

PR #3139, opened 2024-12-18
chore: Update Flow EVM chain definitions

jribbink:
Update Flow EVM chain definitions to align with latest standards:
 - Branding: "Flow EVM"
 - Added missing multicall3 address
 - Deprecate Previewnet as it [does not exist anymore](https://forum.flow.com/t/shutting-down-migration-testnet-and-previewnet/6468).  Mainnet was previously marked as deprecated, but looks like it was a mistake and was supposed to be for previewnet.


src/chains/definitions/flowPreviewnet.ts
jxom, line 1: We will need to `/** @deprecate */` this in `src/chains/index.ts` instead of deleting. We can remove it entirely in v3.

jxom (changes requested):

PR #3150, opened 2024-12-20
feat: Add Aleph Zero

pxrl:
Add Aleph Zero mainnet. Aleph Zero is a new Orbit chain using AZERO as the gas token.


pxrl:
> ### ⚠️ No Changeset found
> Latest commit: [ddf8e9f](https://github.com/wevm/viem/commit/ddf8e9f45891c6589af032f81407eda7bed20ae6)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**

My bad; seems like I want a changeset to bump the patch level.

PR #3151, opened 2024-12-21
refactor(siwe): ox

### LINKED ISSUE #2531: (simulate|write)Contract payable value not working for overloads
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.17.11

### Current Behavior

If an overloaded function has nonpayable and payable overloads, `simulateContract` and `writeContract`'s `value` parameter is not inferred correctly for the payable overload.

### Expected Behavior

Payable overload should allow `value` property.

### Steps To Reproduce

_No response_

https://www.typescriptlang.org/play/?target=99#code/JYWwDg9gTgLgBAbzgYygUwIYzQdQwG3zRgGF9g0A7GAGjgAsYYw6wMoBnNAQQCNg4AXzgAzKBBBwA5ADcKIKXAD0SuDACeYNBwBccAEwA6AIwB2E8YBQoSLERwA7lGDYSEalAzJacDqACu+Fhobh5e8MJiEtJyaCBK4cDuHIoqapraekZmFtbg0PBIIBjAlJTEQqLikrLySsj0JZQpyqoaWroGJubGVpbIyfDI5FTwALwo6MF4hMRkFNQAFAiWcHBeA-7UelIADAAeUjSrKI2lesWl5bQnMJ7NtjB6jMyLAJTHgm-9g+v8cBM2JwePxFgBtE5SERbbxJSi+DAiNAAFXuHCRUAAYtVFhgACZ49AcDh0fGE7QkuD+UowfQAVgAbG8jpDoZRYe4EUjURhmhjsRJcQSiZSySK6NTqPSGXReOpsBw3nA2OoMLwiCyALrfSxOFwhdx3cKLYYLHwrNZiik7A4sy38Y5rNkcygAOQwIDQOw4iJRaP51Tt6ygAHNOmC9vtDNGjtIDtHDLHjLtdpRNY7WnAAAIwDgAWjQ+y03gLUHEUBOMgI-i9cGMlE+3z1rkNnm8JpG1DoFvWwutccOGbVwAzzpgcPdnu9vp5fLQWMDQ9D4cjCdjq5jdGTqboG8T6cr1dr9cblksfhAgWCoSN7dNo27JytxJtg6fDpOY4nHtrUh93P9ecBQUJcwz0CN403Ac1y3FM0wzNIc3zQtixgUty0PfAaz0E9LC+c8AiCFswjvTtzSfPsXwHINh1HGFx3cSdf3-P1eXRIDFyfZdwL3ddIMTWCd2gmMDzWKssOPBs8O+IA

### Anything else?

_No response_

### LINKED ISSUE #2573: Lisk using zkSync formatters seems wrong
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.18.6

### Current Behavior

zkSync receipts seem to have a `transactionLogIndex` property which doesn't exist on Lisk receipts. Maybe this changed with their recent transition to the OP Stack.

### Expected Behavior

client.getTransactionReceipt does not throw when using Lisk

### Steps To Reproduce

import { list } from 'viem/chains'

```
const client = createPublicClient({ chain: lisk })
client.getTransactionReceipt({ hash })
```

_No response_

### Anything else?

N/A

### LINKED ISSUE #2591: `prepareEncodeFunctionData` drop type safety
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.18.8

### Current Behavior

You can pass incorrect values for `args` in `encodeFunctionData` without a type error

### Expected Behavior

A type error just like if you use the `abi` field directly without using `prepareEncodeFunctionData`

### Steps To Reproduce

You can see if I use `prepareEncodeFunctionData`, there is no type error detected

```ts
const constInputData = prepareEncodeFunctionData({
  abi: [
    {
      inputs: [{ internalType: 'string', name: 'input', type: 'string' }],
      name: 'convertedPaimaData',
      outputs: [],
      stateMutability: 'nonpayable',
      type: 'function',
    },
  ],
});

const foo = encodeFunctionData({
  ...constInputData,
  args: [1], // note: wrong data, but no error!
});
```

However, if I inline the abi directly, it properly detects the error

```ts
const foo = encodeFunctionData({
  abi: [
    {
      inputs: [{ internalType: 'string', name: 'input', type: 'string' }],
      name: 'convertedPaimaData',
      outputs: [],
      stateMutability: 'nonpayable',
      type: 'function',
    },
  ],
  args: [1], // error working properly as expected: Type 'number' is not assignable to type 'string'
});
```

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2669: estimateGas usage with nonceManager-bundled account
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.19.8

### Current Behavior

I am using a wallet account, bundled with nonce manager:

```ts
  const nonceManager = createNonceManager({
    source: jsonRpc()
  });
  const privateKeyAccount: PrivateKeyAccount = privateKeyToAccount(privateKey, {
    nonceManager
  }) as PrivateKeyAccount;
  const account = createWalletClient({
    account: privateKeyAccount,
    chain: defineChain({
      name: chainName,
      id: chainId,
      rpcUrls: {
        default: {
          http: [rpcHttpEndpoint]
        }
      },
      nativeCurrency: chainNativeCurrency
    }),
    transport: http()
  }).extend(publicActions);
```

I also have a contract instance, created from ABI and that account.

I want to manually estimate gas for contract function, so I call it like:

```ts
const estimatedGas = await contract.estimateGas.functionToEstimate(
    [...args],
    { value }
)
```

after that call, `nonceManager's` internal nonce counter is broken, because `estimateGas.ts` uses `prepareTransactionRequest`:

https://github.com/wevm/viem/blob/06d7e190b8968b1575cf8be9b2c2d962f7ec8a55/src/actions/public/estimateGas.ts#L131

which calls `nonceManager` stateful function `consume`:
https://github.com/wevm/viem/blob/06d7e190b8968b1575cf8be9b2c2d962f7ec8a55/src/actions/wallet/prepareTransactionRequest.ts#L308-L316
which increments current nonce if it is greater than zero
https://github.com/wevm/viem/blob/06d7e190b8968b1575cf8be9b2c2d962f7ec8a55/src/utils/nonceManager.ts#L55-L66

### Expected Behavior

readonly functions should be stateless.

### Steps To Reproduce

create an account, make sure it has at least one transaction, then follow the steps I note.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2721: 7702: Failed to decode transaction
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.6

### Current Behavior

When using the 7702 example [here](https://viem.sh/experimental/eip7702/contract-writes) anvil (ran with `anvil --hardfork=prague`) returns the error "Failed to decode transaction"

### Expected Behavior

The example works

### Steps To Reproduce

Run the linked example against `anvil --hardfork=prague`

https://github.com/chris13524/viem/blob/failed-to-decode-transaction/examples/7702/index.ts

### Anything else?

Upon debugging myself it appears to be an RLP encoding issue: leading zero. The error is that the nonce is encoded as a zero when it should be an empty array. When using `anvil_setNonce` to bump the nonce to 1 then the decoding error disappears.

https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/

https://github.com/recmo/uint/blob/293d2962a3a5e54ff1cc583d4165003695e89bb5/src/support/alloy_rlp.rs#L81

### LINKED ISSUE #2731: Typescript error: Missing nonceManager in `mnemonicToAccount` options type
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.8

### Current Behavior

When providing the `nonceManager` option, Typescript gives an error even though it's a valid option:
```ts
import { mnemonicToAccount, nonceManager } from 'viem/accounts';

const account = mnemonicToAccount('some mnemonic...', { nonceManager });
```
```
error TS2353: Object literal may only specify known properties, and 'nonceManager' does not exist in type 'HDOptions'.
```

### Expected Behavior

Typescript should not give an error as `hdKeyToAccount` supports passing in a `nonceManager`:
```ts
export function mnemonicToAccount(
  mnemonic: string,
  opts: HDOptions = {},
): HDAccount {
  const seed = mnemonicToSeedSync(mnemonic)
  return hdKeyToAccount(HDKey.fromMasterSeed(seed), opts)
}
```

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2733: `watchBlock` using a `wss` transport not emitting latest block when `emitOnBegin` is enabled
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.0

### Current Behavior

We're running into a case with `watchBlocks` using a `wss` transport not properly emitting the latest block when `emitOnBegin` is enabled. We're noticing this on an L3 Arbitrum chain with low tx volume, Arbitrum only generates new block as new TX come in. 
Blocks are emitted as soon as a new TX/Block is made.

### Expected Behavior

`emitOnBegin` always emits the latest available block on begin
(similar to how `http` would, or `wss` transport using polling)

### Steps To Reproduce

```typescript
import {
  createPublicClient,
  type Client,
  http,
  type Chain,
  webSocket,
} from "viem";
import { arbitrumSepolia as _arbitrumSepolia } from "viem/chains";

export const testConduit = Object.assign(_arbitrumSepolia as Chain, {
  name: "Conduit Cloudlines",
  id: 573146821555,
  blockExplorers: {
    default: {
      name: "Conduit",
      url: "https://explorer-cloudlines-7gm4rnga08.t.conduit.xyz",
    },
  },
  rpcUrls: {
    default: {
      http: ["https://rpc-cloudlines-7gm4rnga08.t.conduit.xyz"],
      webSocket: ["wss://rpc-cloudlines-7gm4rnga08.t.conduit.xyz"],
    },
  },
});

async function main() {
  const publicClient = await createPublicClient({
    transport: webSocket(testConduit.rpcUrls.default.webSocket[0]),
  });

  const publicClient2 = await createPublicClient({
    transport: http(testConduit.rpcUrls.default.http[0]),
  });

  const lastWssBlock = await publicClient.getBlock();
  const lastHttpBlock = await publicClient2.getBlock();

  console.log("lastWssBlock", lastWssBlock);
  console.log("lastHttpBlock", lastHttpBlock);

  publicClient.watchBlocks({
    emitOnBegin: true,
    onBlock: async (block) => {
      console.log("publicClient1: ", block);
    },
    pollingInterval: 1000,
  });

  publicClient2.watchBlocks({
    emitOnBegin: true,
    onBlock: async (block) => {
      console.log("publicClient2:", block);
    },
    pollingInterval: 1000,
  });
}

main().catch((err) => {
  console.error("mainLoop Err", err);
});
```

https://gist.github.com/ya7ya/322444102d4b926c289148bc98973010

### Anything else?

_No response_

### LINKED ISSUE #2767: `waitForTransactionReceipt` timeout exception when `confirmations` argument is provided
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.2.14

### Current Behavior

When calling `waitForTransactionReceipt` and providing a `confirmations` argument >1, the request throws a `WaitForTransactionReceiptTimeoutError`. Same behaviour is present if passing a `timeout` value. From the block explorer, you can see that the transactions are in fact confirmed within the duration of the timeout. Strangely, when no `confirmations` argument is provided, the function returns properly.

Noticing this on Polygon (mainnet) and using alchemy as an RPC for what its worth. 

There is a related issue but is closed. See https://github.com/wevm/viem/issues/2650.

### Expected Behavior

When calling `waitForTransactionReceipt` and providing a `confirmations` argument >1, if the promise resolves within the `timeout` period, the result should be returned and no `WaitForTransactionReceiptTimeoutError` should be thrown.  

### Steps To Reproduce

1. Visit the reproduction link
2. Enter an alchemy key
3. Enter a private key
4. Enter an address to transfer to
5. Run `pnpm install && pnpm start`

or alternatively:
1. Send a transaction on Polygon (mainnet)
2. Replace lines 32-38 (`sendTransaction`) with `const hash = "0x..."`
3. Run `pnpm install && pnpm start`

https://stackblitz.com/edit/viem-getting-started-j89ho7?file=index.ts

### Anything else?

_No response_

### LINKED ISSUE #2775: Improve handling of JSON RPC calls which do not return valid json
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.14

### Current Behavior

We have a scenario where a node could be offline and our nginx would return a 404 error.
When this happens the json rpc throws an error on parsing the json making it very hard to do proper error handling.

The response of the json rpc request would be as followed:

```
HTTP/1.1 404 Not Found
Date: Wed, 25 Sep 2024 11:40:13 GMT
Content-Type: text/plain; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
CF-Cache-Status: DYNAMIC
Strict-Transport-Security: max-age=2592000; includeSubDomains; preload
X-Content-Type-Options: nosniff
Server: cloudflare
CF-RAY: 8c8abcf7f80c74fa-BRU
Content-Encoding: gzip
alt-svc: h3=":443"; ma=86400

default backend - 404
```

When it is parsed it will throw an 'invalid JSON' error.
`Uncaught SyntaxError: Unexpected token 'd', "default ba"... is not valid JSON`

The error is coming from [here](
https://github.com/wevm/viem/blob/a878c56bbd963c1c882306c8ecb78d5441967461/src/utils/rpc/http.ts#L124)

### Expected Behavior

Return a proper error message with the status code.

As a fix I would do the following change in [http.ts](https://github.com/wevm/viem/blob/a878c56bbd963c1c882306c8ecb78d5441967461/src/utils/rpc/http.ts#L124)

```ts
if (
  response.headers.get('Content-Type')?.startsWith('application/json')
)
  data = await response.json()
else {
  data = await response.text()
  try {
    data = JSON.parse(data || '{}')
  } catch {
    data = { data }
  }
}

```

### Steps To Reproduce

Execute a json rpc call on a node which is offline that does not return a valid json response.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2783: Overriding signature field when calling prepareUserOperation() doesn't work (signature ignored)
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.20.0

### Current Behavior

image

Dummy signature provided from the outside ("0x0000...") is ignored and the default siganture is being used.

### Expected Behavior

Expected my signature provided from the outside to be used for a fully prepared userOp.

Think the signature [at the LoC here](https://github.com/wevm/viem/blob/112c4be27eb2cf38017e2db4dff4960b87a6dacb/src/account-abstraction/actions/bundler/prepareUserOperation.ts#L467) should be returned as
```
return parameters.signature ??  account.getStubSignature()
```

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2810: Recent `parseEventLog` breaking change
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.16

### Current Behavior

The recent change to `parseEventLogs` (2fa3d004467945efc548ac23a659c32a1e206cc4) made in version `2.21.8` has a (maybe intentional?) side effect of allowing `undefined` as a valid value of an argument to filter events by. This was listed as a patch update and so I paid it little mind, but it broke my codebase :(

Passing `undefined` to an arg to filter by in strict mode will most likely mean your query will never return any events now. If this is intentional, I'd ask to include some notice of that change either in the functions' in-code docs or the [external docs](https://viem.sh/docs/actions/public/getLogs#arguments). If not, adding `|| value === undefined` to the same lines changed in that commit would bring back the old functionality.

For example, a function like this no longer works:
```ts
const getTokenTransferEvents = async (
  publicClient: PublicClient,
  tokenAddress: Address,
  options?: { from?: Address[]; to?: Address[] }
) => {
  return await publicClient.getLogs({
    address: tokenAddress,
    event: {
      inputs: [
        { indexed: true, internalType: 'address', name: 'from', type: 'address' },
        { indexed: true, internalType: 'address', name: 'to', type: 'address' },
        { indexed: false, internalType: 'uint256', name: 'value', type: 'uint256' }
      ],
      name: 'Transfer',
      type: 'event'
    },
    args: {
      from: options?.from,
      to: options?.to
      // Existing codebases like this would have to use the following:
      // from: options?.from ?? null,
      // to: options?.to ?? null
    },
    strict: true
  })
}
```

### Expected Behavior

The `parseEventLogs` function should ignore `undefined` as it does `null`, or this new behavior should be clarified in the docs if intended, since it's quite hard to find and debug as of now.

### Steps To Reproduce

Pass `undefined` to any `getLogs` function's indexed args to filter by. No events will be returned.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2815: `TestClient.extend` does not expose overridden `client: TestClient` in extension callback
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.17

### Current Behavior

In the following expression: `createTestClient({ ... }).extend(extensionCallback)`, `extensionCallback` expects a `client: Client<transport, chain, account, rpcSchema, extended>` from [this type definition](https://github.com/wevm/viem/blob/03df7fbe0da3332a3ceee645d7ec6ede2a832010/src/clients/createClient.ts#L137)

Whereas the type could be more precise in the case of a `TestClient` which has an additional `mode` property: `TestClient<mode, transport, chain, account, includeActions, rpcSchema>`

This limits the composability of actions as I cannot extend the client with an action that requires a test client (including the mode property) without force casting

### Expected Behavior

In the previous example, I expect `extensionCallback` to have access to a `client: TestClient<mode, transport, chain, account, includeActions, rpcSchema>`

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2888: Inconsistent validation of typed data
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.29

### Current Behavior

In `hashTypedData`, including a struct with name `bytes32` among `types` confuses viem and it:
- validates values of type `'bytes32'` as if they were the native type `bytes32`, and
- encodes values of type `'bytes32'` as if they were the struct type.

```ts
hashTypedData({
  domain,
  types: {
    Message: [{ name: 'contents', type: 'bytes32' }],
    bytes32: [],
  },
  primaryType: 'Message',
  message: {
    contents:
      '0x0000000000000000000000000000000000000000000000000000000000000000',
  },
});
```

In this case the struct is empty, causing the value of `contents` to be ignored. If the struct is not empty the function fails.

### Expected Behavior

The types at validation phase are interpreted exactly as the types at encoding phase.

Struct names that match native types should probably be rejected.

### Steps To Reproduce

_No response_

https://gist.github.com/frangio/5bdc313cea9c6005b9f0a36a78747817

### Anything else?

_No response_

### LINKED ISSUE #2906: OP stack buildInitiateWithdrawal checks L1 balance
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.18.1

### Current Behavior

When calling `publicClientL1.buildInitiateWithdrawal`, it calls [eth_estimateGas](https://github.com/wevm/viem/blob/5525b186c4ddd051c5ee78a03428d67ede944a8c/src/actions/wallet/prepareTransactionRequest.ts#L406) on the L1  using the `value` passed in if the `gas` parameter is not passed in.

### Expected Behavior

Initiating a withdrawal on L2 should not be affected by the funds the user has on the L1. Especially if passing in a large value for `value`, the call will usually fail. This bug can be bypassed by manually specifying a value for `gas`, but this should not be the default behaviour.

### Steps To Reproduce

Deposit to any OP testnet (https://docs.optimism.io/chain/networks#op-sepolia). Then try using the `publicClientL1.buildInitiateWithdrawal` to bridge funds back to L1, ensuring that the amount you are bridging is larger than the balance on the L1. For example, try calling `publicClientL1.buildInitiateWithdrawal` with `value` = 0.5 ETH, when the balance on the L1 is 0.1 ETH. This call should fail on an `eth_estimateGas` error.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2916: Leaks Detected in waitForTransactionReceipt
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.33

### Current Behavior

Using deno, deno detects that waitForTransaction starts a timeout but never clears it resulting in memory leaks.
checking the [implementation](https://github.com/wevm/viem/blob/main/src/actions/public/waitForTransactionReceipt.ts) for `actions/public/waitForTransactionReceipt.ts` we can see that on line 154 we do start a timeout but it is never cleared.
Can this actually be a leak? or is deno just hallucinating?

Here are logs : 
```bash
error: Leaks detected:
  - A timer was started, but never completed. This is often caused by not calling `clearTimeout`. The operation was started here:
    at Object.queueUserTimer (ext:core/01_core.js:773:9)
    at setTimeout (ext:deno_web/02_timers.js:74:15)
    at Timeout.<computed> (ext:deno_node/internal/timers.mjs:68:7)
    at new Timeout (ext:deno_node/internal/timers.mjs:55:37)
    at setTimeout (node:timers:13:10)
    at file:///home/marsian/dev/pumpfaxt/node_modules/.deno/viem@2.21.33_2/node_modules/viem/_esm/actions/public/waitForTransactionReceipt.js:58:13
    at new Promise (<anonymous>)
    at waitForTransactionReceipt (file:///home/marsian/dev/pumpfaxt/node_modules/.deno/viem@2.21.33_2/node_modules/viem/_esm/actions/public/waitForTransactionReceipt.js:56:12)
    at Object.waitForTransactionReceipt (file:///home/marsian/dev/pumpfaxt/node_modules/.deno/viem@2.21.33_2/node_modules/viem/_esm/clients/decorators/public.js:96:46)
    at tx (file:///home/marsian/dev/pumpfaxt/contracts/runtime.local.ts:84:29)
  - A timer was started in this test, but never completed. This is often caused by not calling `clearTimeout`. The operation was started here:
    at Object.queueUserTimer (ext:core/01_core.js:773:9)
    at setTimeout (ext:deno_web/02_timers.js:74:15)
    at Timeout.<computed> (ext:deno_node/internal/timers.mjs:68:7)
    at new Timeout (ext:deno_node/internal/timers.mjs:55:37)
    at setTimeout (node:timers:13:10)
    at file:///home/marsian/dev/pumpfaxt/node_modules/.deno/viem@2.21.33_2/node_modules/viem/_esm/utils/wait.js:2:33
    at new Promise (<anonymous>)
    at wait (file:///home/marsian/dev/pumpfaxt/node_modules/.deno/viem@2.21.33_2/node_modules/viem/_esm/utils/wait.js:2:12)
    at watch (file:///home/marsian/dev/pumpfaxt/node_modules/.deno/viem@2.21.33_2/node_modules/viem/_esm/utils/poll.js:13:15)
    at eventLoopTick (ext:core/01_core.js:175:7)
```

### Expected Behavior

It should not leak.
Perhaps by using `clearTimeout`

### Steps To Reproduce

here is the function `tx`
```ts
async function tx(txn: Promise<Hash>) {
  return await publicClient.waitForTransactionReceipt({
    hash: await txn,
  });
}
```

intended to be used as such : 
```ts
 const txnRcpt = await tx(deployer.deployContract(parameters));
```

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2979: Overloaded contract methods return type not correctly inferred
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.42

### Current Behavior

Using `readContract`, the return type is not correctly inferred if the functionName have overload in ABI.

### Expected Behavior

Using `readContract`, the return type is correctly inferred if the functionName have overload in ABI.

### Steps To Reproduce

```ts
  // data -> never
  const data = await publicClient.readContract({
    address: '0x...',
    abi: abi,
    functionName: 'getSomething',
    args: ['0x...'],
  });

  // data -> bigint or whatever
  const data2 = readContract(publicClient, {
    address: '0x...',
    abi: abi,
    functionName: 'getSomething',
    args: ['0x...'],
  })
```

https://stackblitz.com/edit/viem-getting-started-tvece3?file=index.ts

### Anything else?

_No response_

### LINKED ISSUE #2987: Unable to resolve ./Bytes.js
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.43

### Current Behavior

Simulator Screenshot - iPhone 16 - 2024-11-08 at 09 44 03

### Expected Behavior

Rolling back to 2.21.42 resolved the issue

### Steps To Reproduce

1. Install viem 2.21.43

_No response_

### Anything else?

Using React-native v0.75.4

### LINKED ISSUE #2995: `newHeads` block does not include array of transaction hashes
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.44

### Current Behavior

`watchBlocks` emits a `Block` type, which expects an array of transaction hashes when not using `includeTransactions: true`:
https://github.com/wevm/viem/blob/82633bda23780791fd5b70aa0a6d3bf7e250955a/src/utils/formatters/block.ts#L36

viem uses `newHeads` when subscribing to new blocks via websocket transport:
https://github.com/wevm/viem/blob/82633bda23780791fd5b70aa0a6d3bf7e250955a/src/actions/public/watchBlocks.ts#L238

However, these `newHeads` objects do not contain the list of transaction hashes, so I get runtime errors when attempting to use `block.transactions`:
https://docs.alchemy.com/reference/newheads

I can also see this behavior when using this against an OP Stack chain:

image

### Expected Behavior

I would expect the `Block` type to not guarantee a list of transaction hashes or Viem to populate these transaction hashes for me.

### Steps To Reproduce

```ts
const unwatch = publicClient.watchBlocks({ 
  onBlock: (block) => console.log("block", block.number, block.transactions),
  pollingInterval: 1_000, 
});
```

https://stackblitz.com/edit/viem-getting-started-wngn4c?file=index.ts

### Anything else?

_No response_

### LINKED ISSUE #2999: `hashTypedData` does not check `domain` type at runtime
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.44

### Current Behavior

Good afternoon!

I can pass any value as `domain` to the `hashTypedData` function and it will be hashed regardless without checking its type at runtime.

```
    hashTypedData({
      ...typedData.complex,
      domain: 'wrong' as unknown as Record<string, unknown>,
      primaryType: 'Mail',
    }),
```

The equivalent in `ethers` does check the value at runtime.

### Expected Behavior

I get an error since I am trying to fool the function and the hash that I get as a result does not take the domain into account (might be just an error on my side).

### Steps To Reproduce

Run `pnpm vitest run src/utils/signature/hashTypedData.test.ts -t 'wrong domain value'` on this forked branch: https://github.com/freemanzMrojo/viem/tree/test-hash-typed-data (it uses the code above).

_No response_

### Anything else?

_No response_

### LINKED ISSUE #3028: Improve security of signature generation
### Problem

Transaction signatures use "nonce" / "k" during their construction. The nonce should never be equal between two different messages. Reusing them would allow attacker to recover private key.

Many years ago, `nonce` was generated using system randomness. On some systems with bad quality of randomness, that lead to breakages.

Today, the nonce is generated from private key and message hash using RFC 6979. Basically `hash(private_key, message)`. However, if some issue would be found in serialization / parsing of those, and during generation of nonce, it would still be possible to recover private keys. The technique is described here: https://github.com/pcaversaccio/ecdsa-nonce-reuse-attack.

### Impact

Private key leakage, hackers stealing money from users.

This is not some theoretical issue. This happened in the past. Soon there would be announcement of a new hack related to this.

### Solution

Use [RFC6979 3.6](https://datatracker.ietf.org/doc/html/rfc6979#section-3.6): additional k' `extraData` to mix-in 32 byte of random data on every signature. It is standard way of doing this. It has been extensively [used by Bitcoin](https://bitcoinops.org/en/topics/low-r-grinding/) for non-taproot transactions, to decrease signature size.

https://github.com/wevm/viem/blob/0c98d991b5ec6990251486d860349718f8e7ea04/src/accounts/utils/sign.ts#L36-L41

`secp256k1.sign(msgHash, privateKey)` becomes `secp256k1.sign(msgHash, privateKey, { extraData: true })`

### Disadvantages

- Signatures (r, s) would become non-deterministic and "new" after every signature.
    - They would still be verifiable. This is not a problem for tests because we can specify our own `random` extraData in tests.

There is no risk for security. If passed-through random is bad, the signature security would be just like today, not worse

### LINKED ISSUE #3079: `readContract` does not error when numbers wider than ABI types are passed
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.53

### Current Behavior

Viem allows numbers wider than the types specified by the ABI to be passed to `readContract` without erroring. Depending on the contract, the `readContract` call may not error at all (!) -- (e.g. on a solidity/vyper contract with a fallback, an assembly contract, or non-contract address as I specified in the repro below)

### Expected Behavior

Viem should validate the arguments passed into `readContract` (and other functions that involve ABI encoding arguments) to ensure they fit within the type widths and error prior to sending a request with malformed calldata. 

### Steps To Reproduce

```
import { createPublicClient, http } from 'viem';

  chain: mainnet,
  transport: http("https://eth.public-rpc.com"),
});

const data = await client.readContract({
  address: '0x0000000000000000000000000000000000000000', // contract here is irrelevant, viem should not send the request
  abi:   [{
    type: 'function',
    name: 'repro',
    inputs: [
      {
        name: '128bitInteger',
        type: 'uint128',
        internalType: 'uint128',
      },
    ],
    outputs: [],
    stateMutability: 'view',
  }],
  functionName: 'repro',
  args: [BigInt(2**128)] // 1 more than max for this type
})

console.log(data)
```

See code snippet in steps to reproduce, had some issues with new.viem.sh

### Anything else?

_No response_

tmm:
Drop bunch of code by using `ox/Siwe` internally


tmm:
Closing for now. Viem v3 on the horizon and will take care of all this.

Issue #3142, opened 2024-12-18
nonceManager still incrementing if tx was not sent (estimateGas failure)


nmalzieu:
### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.21.53

### Current Behavior

I'm using `nonceManager` when instantiating my `walletClient` because my code can send multiple tx in parallel
I noticed that when calling `walletClient.writeContract` (or equivalent), if this method throws (for instance not enough gas and the tx is going to fail, hence the transaction is not triggered because `estimateGas` probably fails), the nonce has been incremented, and all the following transactions will just wait forever because there will be a nonce gap

### Expected Behavior

If the call to writeContract or equivalent fails, the nonce should not be incremented (or be re-decremented)?

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/viem-getting-started-9lqybgzb?file=index.ts

nmalzieu:
I don't think the stackblitz will work actually because the account is empty and the nounce will be 0 even after the failure, but using a private key that has transactions will fail

PR #3153, opened 2024-12-21
fix: avoid nonce gap in nonceManager (#3142)

0xrouss:
Fixes [#3142](https://github.com/wevm/viem/issues/3142) by deferring the nonce consumption in `prepareTransactionRequest` until after gas estimation succeeds, avoiding a nonce gap when gas estimation fails.

## Changes
- Reordered logic so `nonceManager.consume(...)` is called after gas estimation
- Added a changeset for a patch release

## Details
- Old behavior: We called `nonceManager.consume(...)` before gas estimation, so if gas estimation failed, we never broadcast a transaction but had already incremented the nonce.
- New behavior: We’ve reordered the logic in `prepareTransactionRequest` so that all fees and gas estimation happen first. Only if those steps succeed do we call `consume(...)` from the nonce manager. This way, no extra nonce is burned on failure.
- Alternative considered: Rolling back the nonce in a try/catch: We tested that approach but decided to defer `consume` entirely until the end, which is simpler and avoids the need for a rollback

### DIFF
--- a/.changeset/brave-cooks-carry.md
+++ b/.changeset/brave-cooks-carry.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed an issue that caused the nonce to increment prematurely in prepareTransactionRequest, resulting in possible nonce gaps. Now the nonce manager logic is arranged so that gas estimation is performed first, ensuring no wasted nonce if gas estimation fails
--- a/.changeset/brave-cooks-carry.md
+++ b/.changeset/brave-cooks-carry.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Fixed an issue that caused the nonce to increment prematurely in prepareTransactionRequest, resulting in possible nonce gaps. Now the nonce manager logic is arranged so that gas estimation is performed first, ensuring no wasted nonce if gas estimation fails
+Fixed premature nonce increment by rearranging gas estimation logic.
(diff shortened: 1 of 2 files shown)


jxom:
I'm not entirely sure this fixes the issue as offchain errors can still occur upon sending a transaction (and the nonce will not be implemented). I think we will need to account for that as well, and probably add a mechanism to reverse the nonce.

davidenwang:
From my understanding of the nonce manager and gas estimation I'm still in favor of this change.

Though @jxom mentions that off-chain errors can still occur, this change eliminates a nonce gap during gas estimation at the very least so in my eyes it seems like a strict improvement.

jxom:
Merging for now. We can fix the next issue after.

jxom:
How is this a nonce manager issue? If a transaction gets stuck, then no transactions can pass through until you reprice or replace the stuck transaction. The nonce manager doesn’t handle that because it is a “nonce manager”, not a “transaction manager”.

Your backend should handle the logic to reprice or replace the stuck transaction.

saqlain1020:
Yeah you are right, but here when viem throws a transaction fail error, then we will have to manually resend all of the other transactions as well.

jxom:
Hi. I had to revert this change as it encountered nonce gaps if the transactions were not being executed on the same JavaScript event loop. Since we are performing asynchronous work before we consume the nonce, this is likely to occur. 

Instead, I would recommend to prepare the transactions manually to check for any failures, and then send them off.

```ts
const requests = await Promise.all([
  client.prepareTransactionRequest(a),
  client.prepareTransactionRequest(b),
  client.prepareTransactionRequest(c),
])

const hashes = await Promise.all(requests.map(client.sendTransaction))
```

PR #3157, opened 2024-12-24
fixed initVerseGenesis rpc

NORAVemp:
The RPC of the initVerseGenesisi chain in the current library is the http protocol, which cannot be accessed online. Now it is changed to the https protocol.


jxom (changes requested):
Multicall contract needs to be verified: https://genesis-testnet.iniscan.com/address/0x0cF32CBDd6c437331EA4f85ed2d881A5379B5a6F

MaryBecky:
@jxom  hi,We have completed the verification of Multicall contract. Can we continue to merge this submission?

PR #3164, opened 2024-12-29
feat(zksync): add `requestExecute` wallet action and `getL2HashFromPriorityOp` util function

danijelTxFusion:
Introduce the `requestExecute` action, enabling the execution of L2 transactions from L1. Additionally, add the utility function `getL2HashFromPriorityOp`, which retrieves the L2 transaction hash from the L1 transaction hash. This allows tracking the status of an L2 transaction triggered from L1.


src/zksync/actions/requestExecute.ts
```diff
@@ -0,0 +1,282 @@
+import type { Address } from 'abitype'
+import type { Hex } from 'viem'
```
jxom, line 2:
```suggestion
import type { Hex } from '../../types/misc.js'
```

PR #3182, opened 2025-01-03
Fixed InitVerseGenesis PRC

MaryBecky:
Modify the RPC link of the initVerseGenesis test chain to the address of the https protocol


jxom:
Reopen the PR with only one changeset please.

PR #3184, opened 2025-01-04
fix: cronosZKEVMTestnet chain id

aalimsahin:
# WHAT

Updated wrong cronosZKEVMTestnet chain id.


jxom (changes requested):
Please create a new chain and deprecate this one instead. [Example](https://github.com/wevm/viem/blob/main/src/chains/index.ts#L147-L148).

PR #3194, opened 2025-01-07
fix: correct logic for identifying repriced transactions

davidenwang:
This PR corrects the logic when identifying the reason associated with a replaced transaction.

Before, `repriced` was used to classify any replacement transaction with the same value and recipient as the original, but I believe it should also check the transaction `data` to ensure that if the transaction was a contract call, the function data has also not changed.

Therefore, I added a check to ensure the original and replacement transaction share the same contract call data before classifying a replacement as `repriced`

Added relevant tests, but had a bit of trouble running the test suite.

### DIFF
--- a/.changeset/sixty-mice-build.md
+++ b/.changeset/sixty-mice-build.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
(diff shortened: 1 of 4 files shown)


jxom (approved):

seongyun-ko (commented):
wow this is really a critical finding...

PR #3197, opened 2025-01-07
Fix LineaTestnet definition

Gwen-M:
Fix LineaTestnet definition

No documentation update needed.

(Gwen, DevRel @ Linea)

### DIFF
--- a/src/chains/definitions/lineaTestnet.ts
+++ b/src/chains/definitions/lineaTestnet.ts
@@ -1,26 +1,28 @@
+import { chainConfig } from '../../linea/chainConfig.js'
 import { defineChain } from '../../utils/chain/defineChain.js'
 
 export const lineaTestnet = /*#__PURE__*/ defineChain({
-  id: 59_140,
-  name: 'Linea Goerli Testnet',
+  ...chainConfig,
+  id: 59_141,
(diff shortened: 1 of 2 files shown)


src/chains/definitions/lineaTestnet.ts
jxom, line 1: Can we please not change this chain and deprecate it instead? I see you have already done that, so just reverting these changes is fine.

jxom:
Please review comment and then reopen PR.

src/chains/definitions/lineaTestnet.ts
Gwen-M: The aim of this change is to avoid creating confusion among developers, as the wagmi.sh documentation is automatically generated using viem chains definitions (https://wagmi.sh/core/api/chains), which results in something a bit weird when you look at the Linea chains.

But I understand perfectly, if it's not allowed to modify the chains I can revert the modifications and just integrate the deprecate.

Maybe the fix needs to be made in the wagmi doc, by no longer displaying the deprecate chains, or by displaying a message to that effect, but if other documentations do the same, the problem is likely to spread (not a big problemn though).

image

jxom: Yes, please raise a discussion on Wagmi repo.

PR #3200, opened 2025-01-07
Delete src/chains/definitions/metisGoerli.ts

nidhinakranii:
We need to delete this file and we are adding Sepolia becuase metis no longer support goreli network.


jxom:
Please do not delete chains. Instead, we need to deprecate them (mark them to be removed in the next major version. [See example.](https://github.com/wevm/viem/blob/497fa72642f39725bc1364e2164ccc8b16e8131d/src/chains/index.ts#L158-L159)

nidhinakranii:
I need to deprecate metisgoerli from chains and add metisSepolia instead.

Do I still need to do a changeset?

PR #3207, opened 2025-01-08
chore: update abstract block explorer to use etherscan

metarsit:
# Changes
- Update to use etherscan instead of old explorer

# Note
- This is an official partnership
- https://x.com/0xcygaar/status/1876312502977667523?s=46


metarsit:
FYI @coffee

PR #3222, opened 2025-01-13
fix(EIP712): do not allow extraneous attributes

freemanzMrojo:
Added a test with what I mean, title is self-explanatory.


src/utils/typedData.ts
```diff
@@ -124,6 +124,23 @@ export function validateTypedData<
...
+      'salt',
+    ]
+    const domainKeys = Object.keys(domain)
+
+    // Check for unexpected keys
+    for (const key of domainKeys) {
+      if (!expectedDomainKeys.includes(key)) {
+        throw new InvalidDomainError({ domain })
+      }
+    }
```
jxom, line 142: Think a better approach would be to check properties against the type (within `validateData`) as a 712 domain can have custom properties.

jxom (changes requested):

src/utils/typedData.ts
freemanzMrojo: Will do, thanks!

freemanzMrojo: Done, thanks for the pointer, now it looks much better.

freemanzMrojo:
Hi @jxom , hope you had a good weekend!

Please, let me know if I should do something else, thanks!

freemanzMrojo:
Hey @jxom , hope you are doing well! If you think I should close this PR, please let me know 🙏 . Thanks!

jxom:
I might close this PR for now as the type-system should pick up these issues, and extraneous attributes don't lead to incorrect hashes.

PR #3227, opened 2025-01-14
Makes types tsconfig use nodenext like everything else.

MicahZoltu:
Currently, the `_types` is built using `node` module resolution, which results in this project generating invalid types.  Details about the TypeScript bug can be found at https://github.com/microsoft/TypeScript/issues/60930#issue-2773343774 and there are no plans to fix it since `node` (which is equivalent to `node10`) is being deprecated in TypeScript soon.

The rest of this project uses `nodenext`, so this changes the types `tsconfig` file to also use `nodenext`, and fix the bug that currently prevents the library from being used correctly in native ESM projects.

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -10,7 +10,7 @@
     "src/node/trustedSetups_esm.ts"
   ],
   "compilerOptions": {
-    "moduleResolution": "node",
+    "moduleResolution": "nodenext",
     "sourceMap": true,
     "rootDir": "./src"
   }


tsconfig.build.json
```diff
@@ -10,7 +10,7 @@
     "src/node/trustedSetups_esm.ts"
   ],
   "compilerOptions": {
-    "moduleResolution": "node",
+    "moduleResolution": "nodenext",
```
jxom, line 13: I think we can just remove this line entirely as this file extends `tsconfig.base.json`.

jxom (changes requested):

jxom:
After comparing the diff of the build output of this change, it seems there is no change between the old output and the new output, so I am unsure how this fixes the issue. 

However, this does clean the config up a bit, so I ended up making this change here: https://github.com/wevm/viem/commit/6f2bd59cead2c73cde9a9739f525a562b6ebb2db, I have added you as a co-author to retain credit.

MicahZoltu:
Ah, because the problem I fixed was in `ox` not `viem`, I somehow got mixed up when submitting a PR.  😖

PR #3229, opened 2025-01-15
Removes webauthn-p256 dependency.

MicahZoltu:
*This PR is mostly meant to be illustrative of how little needs to be changed to drop webauthn-p256.  I don't fully understand how this libraries internal import/export system is designed and I just made whatever changes got things working in order to show what is needed to get everything *compiling* with the removed webauthn dependency.  I'm certain that the way I'm solving some of the problems here are not idiomatic for this repository, and I recommend someone make a separate change or edit this one to have it match the style found in this repository.  Someone should also do a more thorough validation that none of these changes broke anything, though the most likely failure would be a hex string sneaking past the type checks when a bigint is desired or vice versa.*

The webauthn-p256 dependency doesn't play well with pure ESM projects.  You get compiler errors like this if you don't disable `libcheck` in `tsconfig.json` (apps shouldn't need to disable with properly built dependencies).
```
node_modules/webauthn-p256/_types/types.d.ts:27:67 - error TS2536: Type '_K' cannot be used to index type 'fallback'.

27     [_K in Exclude<keys, keyof Item>]?: fallback extends object ? fallback[_K] : undefined;
```

While one could submit a PR to `webauthn-p256` to try to fix the underlying issue, in the case of viem I believe it is easier to just remove the `webauthn-p256` dependency entirely since everything it provides is available in `ox`, which is already a dependency, and dropping a dependency is generally a good thing.

I also added `.pnpm-store` to `.gitignore` because when I install with `pnpm` git wants to commit thousands of cache files.

I'm not familiar with pnpm, so I don't know why the pnpm-lock file changed in some ways, but at least the dependency changes are necessary.


jxom:
Ah yes, this is much needed. I will take a closer look tomorrow.

jxom:
Closing in favor of https://github.com/wevm/viem/pull/3232. Added you as co-author.

PR #3231, opened 2025-01-15
Add Aleph Zero Testnet chain

ggawryal:


jxom:
Needs to be exported from `src/chains/index.ts`. Please reopen the PR when done.

PR #3255, opened 2025-01-22
Add Reddio Sepolia

neilydhan:
Add Reddio Sepolia network for other developers to use.


jxom:
Removing multicall address as it is unverified.

PR #3312, opened 2025-02-06
fix: bundler client type

holic:
Not sure if there was any intent to this being a `Client` rather than a `BundlerClient`, but probably worth being more specific here?


jxom:
Think this is okay as a bundler client can also technically be constructed with `createClient`.

PR #3313, opened 2025-02-06
fix: ensure socket is closed

holic:
attempting to fix https://github.com/wevm/viem/issues/2563

specifically using the approach suggested in https://github.com/wevm/viem/issues/2563#issuecomment-2429379516


jxom:
Let's run with it.

PR #3315, opened 2025-02-06
Adds multicall to berachain mainnet

michelebruno:
This PR adds the multicall3 address to the brand new Berachain Mainnet. It was included in the genesis file so it's available from block 0. 

Thank you Bears! 🐻


lethaale:
@michelebruno FTW

PR #3328, opened 2025-02-10
fix(docs): custom

sambacha:
complete 'eth_wagmi' custom rpc. pretty sure this is an error


jxom:
Think it’s correct. The `^|` notation is TwoSlash autocomplete (as seen in screenshot).

IMG_1880

PR #3344, opened 2025-02-12
feat: add Story Mainnet

timothyshen:


allenchuang:
📖 ❤️

src/chains/definitions/storyOdyssey.ts
jxom, line 1: please deprecate instead of removing

src/chains/definitions/storyTestnet.ts
jxom, line 1: please deprecate instead of removing

src/chains/definitions/story.ts
jxom, line 1: needs format

src/chains/definitions/storyAeneid.ts
jxom, line 1: needs format

jxom (changes requested):
please review comments and then reopen the PR

PR #3354, opened 2025-02-16
feat: Add Story Aeneid chain

rasplarry:
Added story-aeneid

### DIFF
--- a/src/chains/index.ts
+++ b/src/chains/index.ts
@@ -442,7 +442,10 @@ export { sophonTestnet } from './definitions/sophonTestnet.js'
 export { spicy } from './definitions/spicy.js'
 export { step } from './definitions/step.js'
 export { story } from './definitions/story.js'
+export { storyAeneid } from './definitions/storyAeneid.js'
+/** @deprecated Use `storyAeneid` instead. */
 export { storyOdyssey } from './definitions/storyOdyssey.js'
+/** @deprecated Use `storyAeneid` instead. */
 export { storyTestnet } from './definitions/storyTestnet.js'
 export { stratis } from './definitions/stratis.js'
 export { superlumio } from './definitions/superlumio.js'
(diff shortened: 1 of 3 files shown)


src/chains/index.ts
```diff
@@ -442,6 +442,7 @@ export { sophonTestnet } from './definitions/sophonTestnet.js'
 export { spicy } from './definitions/spicy.js'
 export { step } from './definitions/step.js'
 export { story } from './definitions/story.js'
+export { storyAeneid } from './definitions/storyAeneid.js'
```
ppedziwiatr, line 445: Shouldn't the `storyOdyssey` and `storyTestnet` be removed? I believe this generates some confusion. The only testnet described currently in the docs is Aeneid - https://docs.story.foundation/docs/aeneid

rasplarry: From what I understood storyOdyssey and storyTestnet are deprecated.
I added Aeneid because my app needs it, which means I’m not from Story.
I just saw that a previously closed (not merged) PR marked them as deprecated..
I’m not sure if I can remove them, so should I wait for the foundation to add the chain?

ppedziwiatr: I was planning to do the same (i.e. add a PR with the Aeneid testnet), but I've noticed your PR ;)

I just belive that "by the way" the old testnets could be removed (to reduce the confusion among the users) - the Odyssey scanner does not even work anymore - https://odyssey.storyscan.xyz/

I can ask someone from the Story team for confrimation

rasplarry: Could you ask someone then?! 
I'll remove it if it's confirmed. Thank you!

rasplarry: For now I just added the @deprecated mark, as the PR I mentioned above did before

jxom: Removing chains is a breaking change. They will be removed in the next major version.

PR #3372, opened 2025-02-21
chore: update sonic testnet

zeroxdeaf:
Updated Sonic Testnet with latest information from https://chainlist.org/chain/57054


jxom:
Please mark the chain as `/** deprecated */` (see existing ones in `src/chains/index.ts`) and add a new chain.

PR #3389, opened 2025-02-27
Handle Non-JSON Error Responses from Providers

yogaajs:
Some providers, return error responses as strings instead of JSON objects. For example, getBlock.io returns "rate limit reached" as a plain string, causing a SyntaxError when attempting to parse it as JSON.

### DIFF
--- a/src/utils/rpc/webSocket.ts
+++ b/src/utils/rpc/webSocket.ts
@@ -40,7 +40,7 @@ export async function getWebSocketRpcClient(
           const _data = JSON.parse(data);
           onResponse(_data);
         } catch (error) {
-          onResponse(data);
+          onError(error);
         }
       }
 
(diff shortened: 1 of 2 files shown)


src/utils/rpc/webSocket.ts
```diff
@@ -36,7 +36,12 @@ export async function getWebSocketRpcClient(
         onClose()
       }
       function onMessage({ data }: MessageEvent) {
-        onResponse(JSON.parse(data))
+        try {
+          const _data = JSON.parse(data);
+          onResponse(_data);
+        } catch (error) {
+          onResponse(data);
```
shrpne, line 43: probably here should be `onError(error)`

yogaajs: This is not an error, if the provider returns a string instead of an object the message is still valid

shrpne: but according to `onResponse` type https://github.com/wevm/viem/blob/main/src/utils/rpc/socket.ts#L22 it only accept `RpcResponse` object and can't accept string as argument

yogaajs: You're right on that point, I hadn't seen it that way, but a rate limit is sort of an error. In reality anything that is not an rpc response is an error.

PR #3390, opened 2025-02-27
chore: add hyperevm

Quertyy:
Add hyperevm


src/chains/definitions/hyperevm.ts
```diff
@@ -0,0 +1,20 @@
+import { defineChain } from '../../utils/chain/defineChain.js'
+export const hyperevm = /*#__PURE__*/ defineChain({
+  id: 999,
```
jxom, line 4: Chain ID already exists: https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-999.json

jxom:
please review comment and then reopen the PR.

Quertyy:
> please review comment and then reopen the PR.

a PR is currently open which replace Wanchain Testnet by HyperEVM 
https://github.com/ethereum-lists/chains/pull/6840/files

How does it works when there is a conflict like this?

Ulydev:
+1 also need HyperEVM added on viem, what's the next step? @jxom

jxom:
next step is for hyperevm to not use an occupied chain id :)

PR #3394, opened 2025-02-28
add berachain bepolia

bearpong:
Adding new Berachain testnet here, rpc is going to be turned on soon but for now it won't work.


jxom:
please add this chain to `src/chains/index.ts` and reopen the PR.

PR #3398, opened 2025-03-03
Add Somnia Testnet

emmaodia:
@jxom Kindly review. Thank you.


contracts/lib/solady-6c2d0da
jxom, line 1: Please revert this change.

PR #3414, opened 2025-03-05
feat(transport): Add shouldThrow as an option to fallback transport

spalladino:
Adds a `shouldThrow` option to the fallback transport, to customize which errors trigger a fallback vs a throw. We are using this so a contract error (eg a failed `require` in contract code) does not trigger the fallback and fails immediately.

Happy to add some tests if this contribution makes sense.

### DIFF
--- a/.changeset/two-bears-warn.md
+++ b/.changeset/two-bears-warn.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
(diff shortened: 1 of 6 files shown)


jxom:
Looks good. Needs some tests.

spalladino:
@jxom thanks for reviewing! Added a test and changeset entry.

PR #3425, opened 2025-03-09
feat: add HyperEVM Mainnet and Testnet

pricegoupdev:
This PR adds HyperEVM (Hyperliquid) networks to viem. Currently multicall3 is only on HyperEVM mainnet.


jxom:
We don’t accept PRs from private accounts.

pricegoupdev:
Gotcha, just un-privated, sorry. Also fixed the currency name which was an oversight. I will re-open this.

PR #3438, opened 2025-03-12
Add Pyrope Chain

0xxlegolas:
- add pyrope chain
- chain has been added to ethereum [list](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-695569.json)


src/chains/definitions/pyrope.ts
```diff
@@ -0,0 +1,33 @@
...
+      url: 'https://pyrope.blockscout.com',
+    },
+  },
+  contracts: {
+    ...chainConfig.contracts,
+    l1StandardBridge: {
+      [sourceId]: {
+        address: '0xC24932c31D9621aE9e792576152B7ef010cFC2F8',
+      },
+    },
```
jxom, line 31: Is there an official reference for this contract via docs or GH?

0xxlegolas: here is a reference : https://pyropechain.com/docs/contract-addresses

PR #3440, opened 2025-03-13
feat: Local CCIP-Read Batch Gateway

adraffy:
* ENS `UniversalResolver` uses a gateway to circumvent CCIP-Read HTTP failures and parallelize multiple CCIP-Read requests that otherwise would be sequential when triggered from Solidity.
* This PR introduces a local batch gateway handler using a [special-purpose URL](https://github.com/adraffy/viem/blob/batched-gateway/src/utils/ens/localBatchGatewayRequest.ts)

---

* reduces all ENS resolution latency
* makes resolutions private (not leaked to centralized server)

---

- [x] Finalized [ENSIP-21](https://github.com/ensdomains/ensips/pull/23)


jxom:
Nice! What is left for this PR?

adraffy:
All good, let me know what needs changed

adraffy:
okay, [ENSIP-21](https://github.com/ensdomains/ensips/blob/master/ensips/21.md) Batch Gateway Offchain Lookup Protocol is now standardized.  I think this PR is ready to go.

PR #3448, opened 2025-03-15
Add Somnia Testnet

emmaodia:
Kindly review @jxom


contracts/lib/solady-6c2d0da
jxom, line 1: again, please remove this from the diff and i will merge this PR.

PR #3450, opened 2025-03-16
Add Somnia Testnet

emmaodia:
Kindly review @jxom


src/chains/index.ts
jxom, line 1: this file needs to be formatted, please run `pnpm check`

emmaodia: @jxom Done. Thank you.

PR #3458, opened 2025-03-18
Adds missing re-exports.

MicahZoltu:
Without these, something as simple as the following results in a compiler error in projects that depend on viem:
```ts export const createReadClient = () => {
	return createWalletClient({ chain: mainnet, transport: custom(ethereum), cacheTime }).extend(publicActions)
}
```

This is because `publicActions`'s return type depends on `CreateAccessList` and `Simulate*`.

Note: The current mechanism of manually adding these things to `index.ts` seems ripe for this class of failure.  It feels like either an alternative re-export system should be used (e.g., `export * from '...'`) or some automation should be built to ensure that everything is properly re-exported.

I also added `.pnpm-store` to `.gitignore` because with out it this commit was going to be a bajillion files large as `pnpn install` drops a ton of stuff in that folder as a cache.

### DIFF
--- a/.changeset/plenty-mayflies-leave.md
+++ b/.changeset/plenty-mayflies-leave.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added missing re-exports.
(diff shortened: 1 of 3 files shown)


jxom:
We likely need to tweak our [`knip`](https://github.com/webpro-nl/knip) config for this.

PR #3479, opened 2025-03-20
feat: Add hoodi testnet

Volchunovich:
Added [Hoodi](https://github.com/eth-clients/hoodi?tab=readme-ov-file) testnet. 

Problem: due to the recent fork of Pectra into holesky, holesky no longer works properly. As EF decided to migrate to Hoodi, we need Hoodi as testnet for testing our staking products. 

PS: Pectra will be in Hoodi around the 26th March, we need this network to prepare the migration from Holesky to Hoodi


jbrower95:
hi, getting this merged would fix some issues in my app. Thanks!

PR #3487, opened 2025-03-21
feat: add more Monad Block Explorers

therealharpaljadeja:
The PR adds more Monad Block Explorers


jxom:
We don't extend block explorers beyond the default anymore.

PR #3496, opened 2025-03-25
feat: support async rpc headers

sakulstra:
Currently it's not easily possible to use flashbots / mev-share with viem.

While generally i guess it would be nice to support `eth_sendPrivateTransaction`,  `eth_sendBundle` etc. as a `actions` extension, the "root issue" making userland solutions to the problem currently very difficult is the lack of dynamic headers.
On flashbots/mev-share you need to [sign your transaction body](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication), which is currently not possible.

This pr tackles the issue by adding a `asyncHeaders` property which receives the request body and returns asyncHeaders.

https://github.com/wevm/viem/discussions/1746#discussioncomment-12608465

### DIFF
--- a/src/utils/rpc.ts
+++ b/src/utils/rpc.ts
@@ -1,4 +1,4 @@
-import type { MessageEvent } from 'isows'
+import type { MessageEvent } from "isows";
 
 import {
   HttpRequestError,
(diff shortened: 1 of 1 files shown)


sakulstra:
realized i'm on quite outdated viem

PR #3502, opened 2025-03-26
Update writeContract.md

SyedAsadKazmi:
Added a note to the [account](https://viem.sh/docs/contract/writeContract.html#account) section of `writeContract`, as referenced in #3477.

### DIFF
--- a/site/pages/docs/contract/writeContract.md
+++ b/site/pages/docs/contract/writeContract.md
@@ -252,7 +252,7 @@ await walletClient.writeContract({
 
 The Account to write to the contract from.
 
-Accepts a [JSON-RPC Account](/docs/clients/wallet#json-rpc-accounts) or [Local Account (Private Key, etc)](/docs/clients/wallet#local-accounts-private-key-mnemonic-etc). If set to `null`, it is assumed that the transport will handle the filling the sender of the transaction.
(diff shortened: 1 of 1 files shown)


site/pages/docs/contract/writeContract.md
```diff
@@ -264,6 +264,13 @@ await walletClient.writeContract({
 })
 ```
 
+:::note 
+**NOTE** 
+`account` can be an of type `Address` when calling through the wallet directly from a UI (using the browser's injected provider). 
+
+In contrast, `account` can be an type `Account` in any case, whether the signer comes from the browser's injected provider or JSON-RPC.
```
jxom, line 271: reasoning sounds duplicated to https://github.com/wevm/viem/pull/3502/files#diff-9b19daf2ba53e43c72c260b853a10e65b1de1467c5e22d7980ea33ea660456d6R255

SyedAsadKazmi: @jxom, to be honest, it's not super clear—specifically the fact that the `account` can **only** be of type `Address` when coming from the UI (via the browser's injected provider).

A safer approach would be to use `account` as `Account` instead of `Address`, ensuring it works in both cases: a private key signer as well as the injected provider's signer.

jxom: we are not stopping consumers from only using `Account`.

SyedAsadKazmi: @jxom, the scenario I mentioned above specifically applies when `writeContract` is used in an API that relies on the `account` field of type `Address`.

This means a third party cannot call the API using a `client` created via JSON-RPC with a private key. Instead, they must use a `client` created with the browser’s injected provider and the corresponding signer through a UI.

PR #3510, opened 2025-03-27
fix(celo): remove L2 check in fee estimation following the successful hardfork

jeanregisser:
This removes the L2 chain detection check in Celo's fee estimation logic following the successful hardfork. 

Previously, the code would check if the chain was L2 by looking for code at the proxy admin address to determine which fee calculation method to use. This distinction is no longer needed and eliminates an extra `eth_getCode` request during fee estimation.

Changes:
- Removed `isCel2` function and L2 detection logic
- Simplified fee calculation to use a single formula
- Updated tests to remove L1/L2 specific cases

### DIFF
--- a/.changeset/short-scissors-matter.md
+++ b/.changeset/short-scissors-matter.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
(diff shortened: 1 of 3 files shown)


aaronmgdr (approved):
launch it

PR #3531, opened 2025-04-01
chore: add the 1155 abi, expose abi in src/index

RobbyUitbeijerse:
Hello! 

We would love to include an 1155 abi in viem as we're currently able to import the ERC-20 and 721 abis, but still usually still ship a 1155 abi in most of our codebases.

The 1155 abi I added here is the default one from that you can find here https://app.unpkg.com/@openzeppelin/contracts@5.2.0/files/build/contracts/ERC1155.json


RobbyUitbeijerse:
Hey @jxom, in order to get the bundle size workflows checks to ✅ I had to up the `import * from` bundle size with 0.5kb, not sure if you are willing to go there but I would think it's a relatively safe assumption that everyone who's concerned about bundle size would already be using named imports. Let me know if this is suitable 👍

PR #3532, opened 2025-04-01
Mention ZeroDev for using bundler RPC as paymaster

derekchiang:
We recently added support for using ZeroDev bundler RPC as paymaster so I wanted to add a mention here.


jxom:
We only mention sponsors here. [See more](https://github.com/sponsors/wevm)

PR #3535, opened 2025-04-01
feat(4337): entrypoint 0.8 + eip7702

plusminushalf:
This PR will add entrypoint 0.8 & EIP-7702 support for 4337 accounts.


src/account-abstraction/accounts/toSmartAccount.ts
```diff
@@ -115,7 +115,7 @@ export async function toSmartAccount<
         this.getFactoryArgs(),
         implementation.signMessage(parameters),
       ])
-      if (factory && factoryData)
+      if (factory && factoryData && factory !== '0x7702')
```
jxom, line 118: 7702 accounts can still use ERC-6492, right?

jxom, line 118:
```suggestion
      if (factory && factoryData)
```

src/account-abstraction/accounts/toSmartAccount.ts
```diff
@@ -128,7 +128,7 @@ export async function toSmartAccount<
         this.getFactoryArgs(),
         implementation.signTypedData(parameters),
       ])
-      if (factory && factoryData)
+      if (factory && factoryData && factory !== '0x7702')
```
jxom, line 131:
```suggestion
      if (factory && factoryData)
```

src/account-abstraction/accounts/toSmartAccount.ts
plusminushalf: @jxom if the account is not deployed state then ERC-6492 will require a valid factory & factory data. 7702 accounts do not have those but rather will need a valid authorisation to point the EOA to a valid contract code that can be used to validate the signature.

jxom:
Moving to #3580 to unlock CI properly.

PR #3540, opened 2025-04-03
feat: `ensTlds` chain config

jxom:
Resolves unintended behavior that could arise from chain-specific ENS resolvers.

### DIFF
--- a/src/actions/ens/getEnsResolver.ts
+++ b/src/actions/ens/getEnsResolver.ts
@@ -67,26 +67,30 @@ export type GetEnsResolverErrorType =
  */
 export async function getEnsResolver<chain extends Chain | undefined>(
   client: Client<Transport, chain>,
-  {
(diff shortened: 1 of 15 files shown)


src/actions/ens/getEnsResolver.ts
```diff
@@ -67,26 +67,30 @@ export type GetEnsResolverErrorType =
...
+      chain,
       contract: 'ensUniversalResolver',
     })
-  }
+  })()
+
+  const tlds = chain?.ensTlds
+  if (tlds && !tlds.some((tld) => name.endsWith(tld)))
+    throw new Error(
+      `${name} is not a valid ENS TLD (${tlds?.join(', ')}) for this chain.`,
```
tmm, line 92: Instead of `this chain`, maybe `chain ${chain.id}` or name?

PR #3573, opened 2025-04-16
added hyperliquid mainnet and testnet.  Removed Wanchain testnet

conradjohnson:
This is adding HyperEVM Mainnet (chain id: 999) and Testnet (chain id: 998) support and removing Wanchain Testnet support for chain id 999.


conradjohnson:
closing.

PR #3601, opened 2025-04-29
Add RISE Testnet

awesamarth:
Added RISE Testnet (chain ID: 11155931)


jxom:
RPC URLs need to be added here: https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-11155931.json

PR #3602, opened 2025-04-30
cccccbulcrhrckuvdnlllvblbgkinrfkbeblcfeeellcfix: ENS batch gateway function signatures

dgca:


jxom:
think there is an open PR for UR v3 here: https://github.com/wevm/viem/pull/3512/files#diff-75dbee1a05e5b8710a64fd8da1e94e9544bcf870299d83bea3554a7cf85ec992

Viem only supports UR v2 rn.

PR #3603, opened 2025-04-30
fix: ENS batch gateway function signatures

dgca:
#3440 changed the arguments passed to ENS resolver methods "resolve" and "reverse" to support [ENSIP-21](https://github.com/ensdomains/ensips/blob/master/ensips/21.md). Unfortunately, these functions are overloaded, and they don't always take all three arguments.

Base's ENS resolver, [for example](https://basescan.org/address/0xc6d566a56a1aff6508b41f6c90ff131615583bcd#code), extends [ExtendedResolver.sol](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/resolvers/profiles/ExtendedResolver.sol) and does not support the `gateways` argument.

This PR updates the calls to "resolve" and "reverse" to determine the right signature to call based on the existence of `gatewayUrls`.

Note: Currently, if `gatewayUrls` is not found, it defaults to `["x-batch-gateway:true"]`. This PR changes that behavior, as the defaulting is what's ultimately causing the issue.

Also, I noticed that the [UniversalResolver.sol](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/universalResolver/UniversalResolver.sol) contract in the ENS repo currently doesn't have the overloaded resolve/reverse functions, but uses resolve/resolveWithGateways and reverse/reverseWithGateways to support either the binary or ternary signatures.

In Viem's current chain definitions, `mainnet.contracts.ensUniversalResolver.address` points to the [older UniversalResolver](https://etherscan.io/address/0xce01f8eee7E479C928F8919abD53E553a36CeF67#code) which has the overloaded functions, but the latest resolver [referenced by ENS](https://docs.ens.domains/learn/deployments/#mainnet) uses the "WithGateways" suffix. Just wanted to call that out.

Last thing, I could not get tests to run. I was running an Anvil node, provided an RPC URL, but kept getting 404 errors from the HTTP RPC client. Wanted to open this to get a gut check that everything I've said is more or less right before continuing to work on getting those to run.


jxom:
It looks good! One question: why would apps be using Base's ENS resolver for resolving ENS addresses/names? Can we not always use the L1 universal resolver since v3 now has support for L2 reverse resolution? I am worried that if apps (reverse) resolve exclusively on a specific L2, then it's harder to make compatible across chains. There is an open PR for it here: https://github.com/wevm/viem/pull/3512

jxom:
Either way, if there is a use case where using the L1 UR isn't ideal, happy to look into merging this and making local batch gateway opt-in (the privacy + perf gains were nice however).

jxom:
> I am worried that if apps (reverse) resolve exclusively on a specific L2, then it's harder to make compatible across chains.

Thinking about this more: perhaps this isn’t a big issue as apps use tooling (Wagmi) to handle resolution between chains anyway. 

I think it’s okay. I’ll look into merging this tomorrow.

jxom:
Tomorrow is now.

PR #3512, opened 2025-03-27
feat: universalresolver v3 support

TateB:
This PR uses the [Universal Resolution ENSIP](https://github.com/ensdomains/ensips/pull/11) for a stable `UniversalResolver` API, which will become the canonical entrypoint to ENS. The API, when used with the UniversalResolver proxy (TBC) will allow clients to migrate automatically to ENSv2 on launch. For reverse resolution, the new API also enables [ENSIP-19](#)'s chain-specific reverse resolution process by providing a coinType.

Changes:
- Added `evmChainIdToCoinType` helper func
- Added `chainId` and `coinType` param to `getEnsName()` for L2 reverse lookups
- Added `chainId` param to `getEnsAddress()` for parity
- Removed legacy `UniversalResolver` error handling 

Remaining tasks:
- [ ] Add UniversalResolver proxy address (currently using hardcoded)
- [ ] Add tests for chainId-specific `getEnsName()` requests (not yet deployed on mainnet)

PR #3606, opened 2025-04-30
[DO NOT MERGE] fix: ENS batch gateway function signatures (#3605)

jxom:
Reopens https://github.com/wevm/viem/pull/3603. Pending further discussion.

Ref: https://discord.com/channels/1156791276818157609/1367046484880527461

### DIFF
--- a/.changeset/tiny-rabbits-watch.md
+++ b/.changeset/tiny-rabbits-watch.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed ENS batch gateway signatures, and made Local Batch Gateway opt-in.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6232,6 +6232,14 @@ packages:
       typescript:
         optional: true
 
+  viem@2.28.1:
+    resolution: {integrity: sha512-7eqGfxAPlMW9u9aE3SMEFPzNYqqU7uFLKUQyd/GwccyW4OAdq7VqJkPIpdULUePN9m3XmfBunA9mswYFp9sUuQ==}
+    peerDependencies:
+      typescript: '>=5.0.4'
+    peerDependenciesMeta:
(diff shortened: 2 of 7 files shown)


adraffy:
@dgca I believe this is a confusion between the `UniversalResolver` "interface" (which isn't standard but an ENSIP and [interface ](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/universalResolver/IUniversalResolver.sol)are coming) and `IExtendedResolver` interface.  Additionally, the fact that the `UniversalResolver` itself is a misnomer (since it itself isn't an resolver, but a "resolver helper").

It was only by luck that an `IExtendedResolver` could be passed as `UniversalResolver` for resolution because they shared the same signature `resolve(bytes, bytes)` when no gateway was supplied.

The most recently deployed `UniversalResolver` (the one in github staging) currently hasn't been integrated into clients yet (like viem, see https://github.com/wevm/viem/pull/3512).   This design introduces the new `IUniversalResolver` interface and main purpose is enable [ENSIP-19](https://docs.ens.domains/ensip/19/).

If you want to bypass mainnet for resolution, I would just call the resolver directly.  Unfortunately, the default viem helpers only work with a custom UR, not a custom resolver.  Instead, you could use something like [ensjs](https://github.com/ensdomains/ensjs).

[ENSIP-21](https://docs.ens.domains/ensip/21) only applies to `OffchainLookup()`'s that contain a `x-batch-gateway:true` URL.  And that URL is only passed to the `UniversalResolver` in viem when it's called without a custom gateway.  The actual resolver is called by the UR according to ENSIP-10 and`IExtendedResolver` (if supported).  The most recently deployed `UniversalResolver` will likely get `x-batch-gateway:true` set onchain so the special-purpose URL doesn't even to be set in the client.

jxom:
We will close this in favor of #3512, and using the L1 UR with a `chainId`.

PR #3613, opened 2025-05-02
fix: add mnemonic validate before toAccount

izayl:


jxom:
Thanks for the PR. We would need to add support for more mnemonic languages (not just english), and the side-effects of including this PR would have an impact on bundle size (validating against every langset) and performance (performing the validation). I would advise consumers to perform validation themself.

izayl:
what about re-export the `validateMnemonic`?

PR #3614, opened 2025-05-02
Added arenaz chain

alien-max:


package.json
jxom, line 1: Revert this change.

pnpm-lock.yaml
jxom, line 1: Revert this change.

jxom (changes requested):
Please review comments and then reopen PR.

PR #3631, opened 2025-05-07
feat: support block overrides on simulate/call

cruzdanilo:
this pr introduces support for block context overrides (e.g., `block.timestamp`, `block.number`, `baseFee`, etc.) to the `call` action.

since the `simulateContract` action utilizes `call` internally, it also benefits from this enhancement, allowing for more precise simulation of contract execution under specific block conditions.

geth introduced this for `eth_call` in its [v1.12.0 release](https://github.com/ethereum/go-ethereum/releases/tag/v1.12.0) (may 2023).

geth `eth_call` documentation: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-call

**testing:**

a test case for this functionality has been added. however, it is currently marked as `.skip` because anvil does not yet support block overrides for `eth_call`. this test can be enabled once the feature is implemented: https://github.com/foundry-rs/foundry/issues/10458


src/actions/public/call.ts
```diff
@@ -1,4 +1,5 @@
 import { type Address, parseAbi } from 'abitype'
+import * as BlockOverrides from 'ox/BlockOverrides'
```
jxom, line 2: nice :D

PR #3635, opened 2025-05-08
Add Flame mainnet chain

joroshiba:
PR add's the Flame mainnet chain


Genebond (changes requested):
Add ownership @Genebond

PR #3636, opened 2025-05-09
Remove Apexlayer (Broken Link)

h3ku:
Apexlayer is dead, the domain do no longer exist and the its available for anyone to register.


jxom:
Please `/** @deprecate */` chains instead of removing them. We will remove deprecated chains in v3.

PR #3615, opened 2025-05-04
feat: typed eip5792 capabilities

jxom:

PR #3642, opened 2025-05-13
Update sendCalls docs to match eips

okwme:
PR #3615 added the method `formatRequestCapabilities` to ensure `paymasterService` matched the format as described in the official EIP at https://eips.ethereum.org/EIPS/eip-7677#paymasterservice-capability. The docs did not reflect this change so this PR updates them.

NOTE: Coinbase's paymaster does not honor the correct format of paymaster capabilities and still expects `url` to be the key directly under `paymasterService`. I found this topic trying to resolve the issue in my app which is still not resolved. There is an open ticket in Coinbase's discord and I've brought the issue up in Viem's discord as well. I'm not sure if other paymaster services are honoring this format or not. Further discussion and documentation of the errors I encountered are in #3615.

### DIFF
--- a/site/pages/docs/actions/wallet/sendCalls.mdx
+++ b/site/pages/docs/actions/wallet/sendCalls.mdx
@@ -352,9 +352,12 @@ const { id } = await walletClient.sendCalls({
   ],
   capabilities: { // [!code focus]
     paymasterService: { // [!code focus]
-      [chainId]: { // [!code focus]
+      1: { // [!code focus]
         url: 'https://...' // [!code focus]
-      }
+      }, // [!code focus]
+      8453: { // [!code focus]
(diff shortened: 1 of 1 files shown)


jxom:
Thanks, I will forward this to Coinbase.

jxom:
@okwme – I have been talking with Coinbase and it seems that the ERC-7677 spec was actually wrong. I made a [PR here](https://github.com/ethereum/ERCs/pull/1039) to fix this, and have consequently [upstreamed](https://github.com/wevm/viem/commit/ba718e23e3e19a064abc35e31234b7759b0d1b6e) this into Viem.

okwme:
wow thanks for following up on it!!

PR #3648, opened 2025-05-14
feat: fall back to `eth_sendTransaction` on `sendCalls`

jxom:
Adds `experimental_fallback` property to `sendCalls` to indicate for Viem to fall back to executing the calls sequentially
via `eth_sendTransaction` if the wallet does not support EIP-5792.

The `experimental_fallback` property enables [this behavior](https://eips.ethereum.org/EIPS/eip-5792#specification) on the spec:
> Apps may begin using these first three methods immediately, falling back to eth_sendTransaction and eth_getTransactionReceipt when they are not available.

### DIFF
--- a/site/pages/docs/actions/wallet/sendCalls.mdx
+++ b/site/pages/docs/actions/wallet/sendCalls.mdx
@@ -173,7 +173,7 @@ export const walletClient = createWalletClient({
 ### Compatibility Fallback
 
 If the Wallet does not support EIP-5792 and `wallet_sendCalls`, passing the `experimental_fallback` 
-flag to `sendCalls` will indicate for Viem to fall back to executing the calls sequentially
+flag to `sendCalls` will allow Viem to fall back to executing the calls sequentially
 via `eth_sendTransaction`. 
 
 :::code-group
--- a/site/pages/docs/actions/wallet/sendCalls.mdx
+++ b/site/pages/docs/actions/wallet/sendCalls.mdx
@@ -170,6 +170,50 @@ export const walletClient = createWalletClient({
 
 :::
 
+### Compatibility Fallback
+
+If the Wallet does not support EIP-5792 and `wallet_sendCalls`, passing the `experimental_fallback` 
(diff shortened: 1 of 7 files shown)


johanneskares:
Amazing to see the progress on this! 🙏

site/pages/docs/actions/wallet/sendCalls.mdx
```diff
@@ -170,6 +170,50 @@ export const walletClient = createWalletClient({
 
 :::
 
+### Compatibility Fallback
+
+If the Wallet does not support EIP-5792 and `wallet_sendCalls`, passing the `experimental_fallback` 
+flag to `sendCalls` will indicate for Viem to fall back to executing the calls sequentially
+via `eth_sendTransaction`.
```
tmm, line 177:
```suggestion
If the Wallet does not support EIP-5792 and `wallet_sendCalls`, passing the `experimental_fallback` 
flag to `sendCalls` will allow Viem to fall back to executing the calls sequentially
via `eth_sendTransaction`. 
```

src/actions/wallet/sendCalls.ts
```diff
@@ -134,6 +147,68 @@ export async function sendCalls<
...
+      for (const call of calls) {
+        const promise = sendTransaction(client, {
+          account,
+          chain,
+          data: call.data,
+          to: call.to,
+          value: call.value ? hexToBigInt(call.value) : undefined,
+        })
+        promises.push(promise)
+        await new Promise((resolve) => setTimeout(resolve, 32))
```
tmm, line 200: what's this sleeping for?

src/actions/wallet/sendCalls.ts
```diff
@@ -134,6 +147,68 @@ export async function sendCalls<
...
+      }
+      const hashes = await Promise.all(promises)
+      return {
+        id: concat([
```
tmm, line 204: what happens if someone tries to directly use this with `wallet_getCallsStatus`?

PR #3653, opened 2025-05-15
Make WebSocket transport correctly serialize bigints

cdrappi:
First of all thank you to all the maintainers for doing such a tremendous job. viem is a beautiful project

**Problem**
When sending data with `BigInt`s over WebSocket Transport, viem will throw this error:
```
TypeError: Do not know how to serialize a BigInt
```

This behaves correctly using HTTP Transport, because HTTP transport uses `stringify` instead of the built-in `JSON.stringify`

**Proposed fix**
Use `stringify` in the WebSocket client, just like in the HTTP client

**Notes**
- I understand that this behaviour might be intentional; if so I won't be sad if you close my PR
- I added one test that mimics the use case that led me to find this
    - It should fail now but succeed with this patch
- After more struggling than I'd like to admit, I was unable to get the test suite running locally

### DIFF
--- a/src/utils/rpc/webSocket.test.ts
+++ b/src/utils/rpc/webSocket.test.ts
@@ -1,7 +1,9 @@
 import { WebSocket } from 'isows'
 import { describe, expect, test, vi } from 'vitest'
+import { privateKeyToAccount } from '~viem/accounts/privateKeyToAccount.js'
 
 import { anvilMainnet } from '../../../test/src/anvil.js'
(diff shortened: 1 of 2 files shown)


jxom:
I don't believe this is the correct usage of `eth_signTypedData_v4` (passing the object itself). You should be serializing it within the parameters with `serializeTypedData`. [Example](https://github.com/wevm/viem/blob/main/src/actions/wallet/signTypedData.ts#L187-L192)

PR #3665, opened 2025-05-20
Add Tea Sepolia

Yutaro-Mori-eng:


jxom:
Chain conflict: https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-842.json

PR #3668, opened 2025-05-21
Depracate Zircuit Testnet 48899

gil7788:
Zircuit testnet 48899 is no longer available, instead one should use Zircuit Garfield testnet 48898 (which is already supported)


jxom:
Please `/** deprecate */` chains instead of removing them.

PR #3671, opened 2025-05-23
Fix: waitForTransactionReceipt() leaks on timeout

frankudoags:
closes https://github.com/wevm/viem/issues/3515

### DIFF
--- a/src/actions/public/waitForTransactionReceipt.ts
+++ b/src/actions/public/waitForTransactionReceipt.ts
@@ -152,26 +152,29 @@ export async function waitForTransactionReceipt<
   let receipt: GetTransactionReceiptReturnType<chain>
   let retrying = false
 
+  // biome-ignore lint/style/useConst: <we need to reassign these variables to the actual unobserve and unwatch functions later>
+  let _unobserve: () => void
+  let _unwatch: () => void
+
   const { promise, resolve, reject } =
     withResolvers<WaitForTransactionReceiptReturnType<chain>>()
 
   const timer = timeout
     ? setTimeout(
-        () => reject(new WaitForTransactionReceiptTimeoutError({ hash })),
+      () => {
+        _unwatch()
+        _unobserve()
+        reject(new WaitForTransactionReceiptTimeoutError({ hash }))
+      },
         timeout,
       )
     : undefined
 
-  // keep track of when we started "waiting" for the transaction receipt
-  // so that we can stop the internal polling of the block number if the timeout is reached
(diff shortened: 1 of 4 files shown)


src/actions/public/waitForTransactionReceipt.ts
```diff
@@ -176,6 +181,12 @@ export async function waitForTransactionReceipt<
         poll: true,
         pollingInterval,
         async onBlockNumber(blockNumber_) {
+          // If the timeout is reached, stop polling.
+          if (timeout && Date.now() - startTime >= timeout) {
+            _unwatch()
+            _unobserve()
```
jxom, line 187: Would it work if we invoked this on L160, and we declared the `_unobserve` + `_unwatch` variables before L158? Means we don't have to manage `startTime`, and it will teardown exactly when the timeout hits.

E.g.

```diff
+ let _unobserve
+ let _unwatch

const timer = timeout
  ? setTimeout(
      () => {
+       _unwatch()
+       _unobserve()
        reject(new WaitForTransactionReceiptTimeoutError({ hash }))
      },
      timeout,
    )
  : undefined
  
- const _unobserve = observe(
+ _unobserve = observe(
...
-   const _unwatch = getAction(
+   _unwatch = getAction(
```

frankudoags: yes, exactly, I see, did this because they were declared after. so I declare them before, invoke them inside the timeout, knowing fully well that they'd become the actual functions they should be. 

Do you reckon I do that?

jxom: Go for it.

frankudoags: all done now @jxom PTAL

src/actions/public/waitForTransactionReceipt.ts
```diff
@@ -152,195 +152,196 @@ export async function waitForTransactionReceipt<
...
   const { promise, resolve, reject } =
     withResolvers<WaitForTransactionReceiptReturnType<chain>>()
 
   const timer = timeout
-    ? setTimeout(
-        () => reject(new WaitForTransactionReceiptTimeoutError({ hash })),
-        timeout,
-      )
+    ? setTimeout(() => {
+        _unwatch()
```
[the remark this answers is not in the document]
duykhanh-pendle: @frankudoags I just created a demo

https://stackblitz.com/edit/github-p8vofkjc?file=index.ts
image

PR #3674, opened 2025-05-24
Hyperevm chain

simeonkerkola:
Added HyperEVM chain


D2758695161:
This async/cache issue seems like a good fit. I can investigate and fix it.

Open to paid work. TypeScript, Node.js, Web3, Rust.
DM or reply here.

PR #3698, opened 2025-06-04
feat: add `waitAsRateLimit` option on `http` transport

iyarsius:
This PR introduces an option to enable a batch queue system by setting `batch.waitAsRateLimit` to `true`. I was searching for a solution until I discovered #1305, which motivated me to explore ways to update the HTTP client.

The core idea is to modify the `batchScheduler` by changing the `shouldSplitBatch` parameter to a `getBatchSize` parameter. Once the scheduler can determine the batch size, it becomes capable of queuing requests.
I also updated multicall to utilize `getBatchSize` with behavior similar to the previous version.

The main drawback is that if too many requests are queued, the queue could continuously grow, potentially causing delays or triggering cache limits. I have added a warning about this in the documentation.

Therefore, for users interacting with rate-limited endpoints, this option could be extremely beneficial for managing such interactions.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6240,6 +6240,14 @@ packages:
       typescript:
         optional: true
 
+  viem@2.30.6:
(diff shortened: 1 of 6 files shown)


jxom:
Thanks, will review!

jxom:
Closing this out since the PR is stale. Feel free to reopen if you want to pick it back up.

PR #3702, opened 2025-06-05
feat: add `chain.blockTime`

jxom:


src/celo/chainConfig.ts
```diff
@@ -4,6 +4,7 @@ import { formatters } from './formatters.js'
 import { serializers } from './serializers.js'
 
 export const chainConfig = {
+  blockTime: 2_000,
```
aaronmgdr, line 7:
```suggestion
  blockTime: 1_000,
```

celo's block time is 1 second  https://specs.celo.org/deployments.html?highlight=second#mainnet

PR #3717, opened 2025-06-10
Add additional known errors to trigger the sendCalls experimental_fallback

azf20:
Mentioned in #3643 ([comment](https://github.com/wevm/viem/discussions/3643#discussioncomment-13368511)), there are certain wallets / chains where the `experimental_fallback` is not triggered because the error returned is not expected (either situations where sendCalls is supported, but a given chain can't use it, or an unexpected error). 

This PR adds additional matches to the catch statement so that these errors trigger the experimental fallback. Currently this has the same "belt and braces" approach (matching the `name` and the `details`), but that might be overkill.

Tested and verified the fix locally with the affected wallets, but didn't update sendcalls.test.ts (could add mocks with these error types)

### DIFF
--- a/src/actions/wallet/sendCalls.ts
+++ b/src/actions/wallet/sendCalls.ts
@@ -155,8 +155,6 @@ export async function sendCalls<
       experimental_fallback &&
       (error.name === 'MethodNotFoundRpcError' ||
         error.name === 'MethodNotSupportedRpcError' ||
-        error.name === 'InternalRpcError' ||
-        error.name === 'UnsupportedChainIdError' ||
         error.name === 'UnknownRpcError' ||
         error.details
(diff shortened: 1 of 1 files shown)


src/actions/wallet/sendCalls.ts
```diff
@@ -155,13 +155,21 @@ export async function sendCalls<
       experimental_fallback &&
       (error.name === 'MethodNotFoundRpcError' ||
         error.name === 'MethodNotSupportedRpcError' ||
+        error.name === 'InternalRpcError' ||
+        error.name === 'UnsupportedChainIdError' ||
```
jxom, line 159: wouldn't we not want to fall back here?
```suggestion
```

PR #3757, opened 2025-06-22
Update: Updated new Lumia Testnet Details for VIEM

dnzdlklc:
Updates details for new Lumia Beam Testnet with Multicall3 in place.

### DIFF
--- a/src/chains/definitions/lumiaTestnet.ts
+++ b/src/chains/definitions/lumiaTestnet.ts
@@ -1,9 +1,9 @@
 import { defineChain } from '../../utils/chain/defineChain.js'
 
 export const lumiaTestnet = /*#__PURE__*/ defineChain({
-  id: 1952959480,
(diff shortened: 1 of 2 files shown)


src/chains/definitions/lumiaTestnet.ts
```diff
@@ -1,29 +1,29 @@
 import { defineChain } from '../../utils/chain/defineChain.js'
 
 export const lumiaTestnet = /*#__PURE__*/ defineChain({
-  id: 1952959480,
-  name: 'Lumia Testnet',
-  network: 'LumiaTestnet',
+  id: 2030232745,
```
jxom, line 4: If you're adding a new chain ID, please deprecate this chain (as seen in `src/chains/index.ts`) and add a new chain.

jxom:
See comment and then re-open PR.

PR #3760, opened 2025-06-25
Add plume slug edits

kristinatub:


src/chains/definitions/plumeTestnet.ts
jxom, line 1: if new chain ID, please deprecate this one, and add a new chain. see `src/chains/index.ts` for examples

jxom:
review comments and then re-open the PR

PR #3762, opened 2025-06-26
fix: Local Batch Gateway Recursion

adraffy:
- allow recursion in local batch gateway to support nested batching


tmm (commented):
can we add a test for this?

adraffy:
@tmm added

I have a much more complicated test [here](https://github.com/ensdomains/namechain/blob/feat/bet-379-dns-resolver/contracts/test/universalResolver/UniversalResolver.test.ts).

PR #3772, opened 2025-07-02
Replay subscriptions when reconnecting in websocket

### LINKED ISSUE #3771: WebSocket subscriptions are lost after reconnection and not automatically restored
### Check existing issues

### Viem Version

2.31.6

## Description

When using viem's WebSocket transport with `reconnect: true`, the client successfully reconnects after a connection drop, but any active subscriptions (e.g., `watchBlockNumber`, `watchContractEvent`) are lost and not automatically restored. This causes clients to silently stop receiving events after a reconnection.

## Current Behavior

1. Client creates a subscription using `watchBlockNumber` or similar methods
2. WebSocket connection drops (network issue, server restart, etc.)
3. Viem automatically reconnects to the WebSocket endpoint
4. The subscription is lost - no more events are received
5. No error is thrown to indicate the subscription failure

## Impact

This bug affects any production application using viem with WebSocket subscriptions:
- DeFi applications monitoring contract events
- Block explorers tracking new blocks
- Trading bots watching for specific events
- Any real-time blockchain monitoring system

The silent failure makes this particularly problematic as applications may appear to be running normally while missing critical events.

## Suggested Solution

The WebSocket transport should:
1. Maintain a registry of active subscriptions with their parameters
2. On successful reconnection, iterate through active subscriptions and re-subscribe
3. Map the new subscription IDs to the existing callbacks
4. Handle edge cases like subscriptions that fail to re-establish

## Environment

- viem version: 2.31.6
- Node.js version: (via Bun)
- TypeScript version: 5.x
- Happens with all WebSocket RPC providers (Alchemy, Infura, local nodes)

## Related Issues

This may be related to WebSocket transport reconnection handling in general. Similar issues have been reported in other Ethereum libraries (ethers.js, web3.js) and were fixed by implementing subscription replay on reconnection.

### Expected Behavior

When the WebSocket transport is configured with `reconnect: true`, viem should:
- Track active subscriptions
- After successful reconnection, automatically re-establish all active subscriptions
- Continue delivering events to the original callbacks without interruption

### Steps To Reproduce

I've created a complete reproduction repository: **https://github.com/hazelnutcloud/viem-ws-replay-bug**

### Quick test:
```bash
git clone https://github.com/hazelnutcloud/viem-ws-replay-bug
cd viem-ws-replay-bug
bun install
./run-test.sh
```

### Manual reproduction:

```typescript
import { createPublicClient, webSocket } from 'viem';

  chain: mainnet,
  transport: webSocket('wss://eth-mainnet.g.alchemy.com/v2/YOUR_API_KEY', {
    reconnect: true,
  }),
});

// This subscription will stop working after any reconnection
const unwatch = client.watchBlockNumber({
  onBlockNumber: (blockNumber) => {
    console.log('New block:', blockNumber);
  },
  onError: (error) => {
    console.error('Error:', error);
  },
});
```

To test:
1. Start the subscription
2. Disconnect network or restart the RPC server
3. Observe that viem reconnects but no more block events are received

## Test Case

The reproduction repository includes:

- **server.ts**: WebSocket server that simulates eth_subscribe API and disconnects after 10 seconds
- **client.ts**: Viem client that demonstrates subscription loss after reconnection
- **run-test.sh**: Script to run the complete test

Running this test shows:
- Events arrive every 2 seconds initially
- After 10 seconds, server disconnects
- Client reconnects successfully
- No more events are received (subscription lost)

https://github.com/hazelnutcloud/viem-ws-replay-bug

### Anything else?

_No response_

hazelnutcloud:
Fixes #3771

This PR fixes the issue where WebSocket subscriptions are lost after a reconnection and not automatically restored. When using the WebSocket transport with `reconnect: true`, subscriptions created via `watchBlockNumber`, `watchContractEvent`, etc. would silently stop working after the connection dropped and reconnected.

### DIFF
--- a/.changeset/curly-glasses-fix.md
+++ b/.changeset/curly-glasses-fix.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed issue where WebSocket subscriptions did not replay on reconnect.
--- a/src/clients/transports/webSocket.test.ts
+++ b/src/clients/transports/webSocket.test.ts
@@ -212,13 +212,7 @@ test('throws on socket closure', async () => {
   socket.close()
   await wait(100)
 
-  expect(error).toMatchInlineSnapshot(`
-    [SocketClosedError: The socket has been closed.
-
-    URL: http://localhost
-
-    Version: viem@x.y.z]
-  `)
+  expect(error).toMatchInlineSnapshot(`undefined`)
 })
 
 test('throws on bogus subscription', async () => {
--- a/src/clients/transports/webSocket.test.ts
+++ b/src/clients/transports/webSocket.test.ts
@@ -212,7 +212,7 @@ test('throws on socket closure', async () => {
   socket.close()
   await wait(100)
 
-  expect(error).toMatchInlineSnapshot(`undefined`)
+  expect(error).toMatchInlineSnapshot('undefined')
 })
 
 test('throws on bogus subscription', async () => {
--- a/src/utils/rpc/socket.ts
+++ b/src/utils/rpc/socket.ts
@@ -197,13 +197,15 @@ export async function getSocketRpcClient<socket extends {}>(
           keepAliveTimer = setInterval(() => socket.ping?.(), keepAliveInterval)
         }
 
-        if (reconnect && reconnectCount > 0) {
-          subscriptions.clear()
-          const requestEntries = requests.entries()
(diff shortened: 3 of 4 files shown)


Amnat-auyana1998 (approved):

Amnat-auyana1998:
@Amnat-auyana1998

PR #3774, opened 2025-07-03
Update API URLs for various chains to use Etherscan v2 endpoints

Jds-23:
This PR replaces the soon-to-be-deprecated Etherscan v1 API URLs with new v2 multichain API URLs. Read about it here: https://docs.etherscan.io/etherscan-v2/v2-quickstart

Did not add tests, as this is just a data update.


jxom:
Hmm, this may be a breaking change as folks would likely be building up their search param already (so they could append another `?`). I would say we don't have to worry about this as folks using the V2 APIs would not use the `apiUrl` here and just do something like `https://api.etherscan.io/v2/api?chainid=${chain.id}` instead of `chain.blockExplorers.default.apiUrl`

PR #3779, opened 2025-07-07
Fixed plume mainnet to plume

kristinatub:


rachitag112:
Hey @jxom, this is Rachit from Plume team. Can you please review this.
We have updated the Plume domain and there's also a new testnet in correspondence to the mainnet for which the chainID and details have been updated.
Ref: https://docs.plume.org/plume/developers/network-information

jxom:
If chain IDs have changed, you will need to deprecate them and add new ones.

PR #3803, opened 2025-07-16
feat: add hyperEVM chain

Kang-heeju:
Added hyperEVM chain


jxom:
Chain ID already occupied in ethereum-lists/chains. See https://github.com/ethereum-lists/chains?tab=readme-ov-file#collision-management

PR #3806, opened 2025-07-16
feat(4337): add `toUserOperation` utility to convert `PackedUserOperation` to `UserOperation`

jeanregisser:
### Why this is needed
**Real user need**: When analyzing account abstraction transactions, or working with existing `PackedUserOperation` data, you often need the unpacked format to use with functions that expect the standard `UserOperation` format (e.g. `getUserOperationHash`).

**The gap**: Bundlers and newer EntryPoints use packed format for efficiency, but there's no built-in way to convert back to the unpacked format when needed for analysis or compatibility.

### What it does
Converts packed user operations to standard unpacked format by splitting combined fields:
- `accountGasLimits` → separate `verificationGasLimit` + `callGasLimit`
- `gasFees` → separate `maxPriorityFeePerGas` + `maxFeePerGas`  
- `paymasterAndData` → separate paymaster fields
- `initCode` → separate `factory` + `factoryData`

Works across all EntryPoint versions and safe to call on already unpacked data. This completes the conversion utilities - now you can go both ways between packed and unpacked formats.

### DIFF
--- a/.changeset/selfish-goats-appear.md
+++ b/.changeset/selfish-goats-appear.md
@@ -1,5 +1,5 @@
 ---
-"viem": minor
+"viem": patch
 ---
 
 **Account Abstraction**: Added `toUserOperation` utility to convert `PackedUserOperation` to `UserOperation`.


jxom:
Let's use `UserOperation.from` from Ox for this.

PR #3823, opened 2025-07-24
fix: Update getWithdrawalStatus.ts to include new custom errors

nicholaspai:
The recent OptimismPortal2 upgrade has broken this part of the code because the error messages, read by the SDK here to determine the withdrawal status, can no longer decipher the error reasons.

This PR is incomplete because the custom Error's also need to be added to the ABI files. I'll let someone else take them over, I just wanted to open this PR to highlight this urgent issue.


nicholaspai:
addressed by https://github.com/wevm/viem/pull/3824

PR #3828, opened 2025-07-29
fix: update story explorer

jdubpark:
- Deprecate old Story testnet and unsupported testnet configuration
- Update explorer links for Story mainnet and Story Aeneid testnet


jxom:
Please `/** @deprecate */` chains instead of removing them.

PR #3845, opened 2025-08-06
sendCalls: Support magic.link with experimental_fallback

johanneskares:
magic.link RPC returns `InternalRpcError` when trying to use with `wallet_sendCalls`.


jxom:
Hmm, I'm not confident that `InternalRpcError` is deterministic enough to add as a fall back.

PR #3848, opened 2025-08-08
feat: UniversalResolver v3 support

TateB:
This PR uses [ENSIP-23](https://docs.ens.domains/ensip/23) for a stable `UniversalResolver` API, which is now the canonical entrypoint to ENS. The API, when used with the UniversalResolver proxy (`0xeeeeeeee14d718c2b47d9923deab1335e144eeee`) will allow clients to migrate automatically to ENSv2 on launch. For reverse resolution, the new API also enables [ENSIP-19](#)'s chain-specific reverse resolution process by providing a coinType.

Changes:
- Added `evmChainIdToCoinType` helper func
- Added `chainId` and `coinType` param to `getEnsName()` for L2 reverse lookups
- Added `chainId` param to `getEnsAddress()` for parity
- Removed unused legacy `UniversalResolver` error handling
- Removed unused `ensRegistry` on mainnet/holesky/sepolia
- `getEnsName` now uses the local batch gateway by default

### DIFF
--- a/.changeset/nervous-kangaroos-fetch.md
+++ b/.changeset/nervous-kangaroos-fetch.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
+Added support for chain-specific ENS resolution and ENS UniversalResolver v3.


shrugs:
There might also be an issue with defaulting to eth mainnet's coinType if no coinType is provided — this 'eth mainnet as default' is explicitly deprecated in ENSIP-19 https://docs.ens.domains/ensip/19/#deprecating-mainnet-as-default which tells me that the api surface for reverse resolution _requires_ a `coinType` (or a `chainId` from which to derive one). i.e. it's the consumer's responsibility to provide a specific `coinType` context within which to perform reverse resolution

TateB:
> which tells me that the api surface for reverse resolution _requires_ a `coinType` (or a `chainId` from which to derive one). i.e. it's the consumer's responsibility to provide a specific `coinType` context within which to perform reverse resolution

that’s correct, but we’re also trying to get the latest universal resolver out to as many consumers as possible to allow for a smooth transition to ENSv2. 

requiring `coinType` then turns these changes into breaking changes for the viem API, and it’d probably be better to rip the bandaid off in a breaking change after this migration, to prevent dissuading consumers from updating viem.

jxom:
Just need to deploy contracts to these chains and update addresses:
- [`linea`](https://github.com/wevm/viem/blob/9c73d150606dc72f82b3533e7807f0d86fd481a8/src/chains/definitions/linea.ts#L31)
- [`berachain`](https://github.com/wevm/viem/blob/9c73d150606dc72f82b3533e7807f0d86fd481a8/src/chains/definitions/berachain.ts#L19)
- [`story`](https://github.com/wevm/viem/blob/9c73d150606dc72f82b3533e7807f0d86fd481a8/src/chains/definitions/story.ts#L21)

PR #3879, opened 2025-08-21
Sonic Testnet and Sonic Blaze Testnet updated

Mejuck:
Updated sonicTestnet 64165 with its new details, 14601. Since Sonic Labs used same rpc and scanner links, updated deprecated chain file instead of creating a new one. 

Updated sonicBlazeTestnet blockscanner link. Current link is now used by 14601, new updated blokcscanner link is for 57054.

Updated deprecated info on index.ts. Previously sonicTestnet was deprecated but now main testnet will be sonicTestnet and sonicBlazeTestnet will deprecate.

Team announcement for this change
https://x.com/0xseg/status/1958495888273248681

Opened a pull request for new 14601 chain on ethereum-lists
https://github.com/ethereum-lists/chains/pull/7633

### DIFF
--- a/src/chains/definitions/sonicTestnet.ts
+++ b/src/chains/definitions/sonicTestnet.ts
@@ -1,7 +1,11 @@
 import { defineChain } from '../../utils/chain/defineChain.js'
 
+// 64165 deprecated for 57054 and with Sonic Labs newest Pectra upgrade
+// 57054 is going to deprecate for this new 14601 chain.
+// Since rpc is same on this one with 64165, override the old infos here.
+
 export const sonicTestnet = /*#__PURE__*/ defineChain({
-  id: 64_165,
+  id: 14_601,
   name: 'Sonic Testnet',
   nativeCurrency: {
     decimals: 18,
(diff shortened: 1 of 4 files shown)


src/chains/definitions/sonicTestnet.ts
```diff
@@ -1,7 +1,11 @@
 import { defineChain } from '../../utils/chain/defineChain.js'
 
+// 64165 deprecated for 57054 and with Sonic Labs newest Pectra upgrade
+// 57054 is going to deprecate for this new 14601 chain.
+// Since rpc is same on this one with 64165, override the old infos here.
+
 export const sonicTestnet = /*#__PURE__*/ defineChain({
-  id: 64_165,
+  id: 14_601,
```
jxom, line 8: please do not change chain ids

Mejuck: 64165 chain is not active anymore, there is not a single rpc pointing to it. and currently the rpc used in 64165 is now used by 14601 so i thought updating this would make sense, to prevent sending wrong txs. if chain id update is not desired, i can create a new file.

PR #3891, opened 2025-08-27
feat: add support for HyperEVM

Kittikornl:
A PR to add chain support for the HyperEVM chain

### DIFF
--- a/.changeset/popular-hounds-jog.md
+++ b/.changeset/popular-hounds-jog.md
@@ -1,5 +1,5 @@
 ---
-"viem": minor
+"viem": patch
 ---
 
 Added HyperEVM chain.


iamshvetsov:
Unfortunately, `chainId: 999` is already taken by Wanchain Testnet, and therefore HYPE is not listed in [ethereum-lists/chains](https://github.com/ethereum-lists/chains)

src/chains/definitions/hyperEVM.ts
```diff
@@ -0,0 +1,12 @@
+import { defineChain } from '../../utils/chain/defineChain.js'
+
+export const hyperEVM = /*#__PURE__*/ defineChain({
+  id: 999,
```
jxom, line 4: [Chain mismatch in registry](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-999.json)

jxom (changes requested):

PR #3895, opened 2025-08-27
feat: add optional `fetchFn` override to http transport to allow full fetch interception

coffeexcoin:
adds optional `fetchFn` parameter to http transport config to allow full interception/decorator/mocking of fetch

this is useful for a number of scenarios including:
- applying a decorator to node fetch for capturing metrics
- easier testing by passing a mock fetch instead of overwriting global fetch
- any scenario where the client needs to be able to process hooks before and after a fetch, while maintaining the same context. This can't be done with the current onRequest/onResponse as there is no way to link the two calls to the same request/response cycle

### DIFF
--- a/src/utils/rpc/http.ts
+++ b/src/utils/rpc/http.ts
@@ -29,7 +29,9 @@ export type HttpRpcClientOptions = {
   /** A callback to handle the response. */
   onResponse?: ((response: Response) => Promise<void> | void) | undefined
   /** Override for the fetch function used to make requests. */
-  fetchOverride?: typeof fetch | undefined
+  fetchOverride?:
+    | ((input: string | URL | Request, init?: RequestInit) => Promise<Response>)
+    | undefined
   /** The timeout (in ms) for the request. */
   timeout?: number | undefined
 }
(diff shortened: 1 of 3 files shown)


src/utils/rpc/http.ts
```diff
@@ -28,6 +28,10 @@ export type HttpRpcClientOptions = {
     | undefined
   /** A callback to handle the response. */
   onResponse?: ((response: Response) => Promise<void> | void) | undefined
+  /** Override for the fetch function used to make requests. */
+  fetchOverride?:
+    | ((input: string | URL | Request, init?: RequestInit) => Promise<Response>)
+    | undefined
```
jxom, line 34:
```suggestion
  fetchFn?:
    | ((input: string | URL | Request, init?: RequestInit) => Promise<Response>)
    | undefined
```

src/utils/rpc/http.ts
```diff
@@ -117,7 +124,7 @@ export function getHttpRpcClient(
             }
             const request = new Request(url, init)
             const args = (await onRequest?.(request, init)) ?? { ...init, url }
-            const response = await fetch(args.url ?? url, args)
+            const response = await fetch_(args.url ?? url, args)
```
jxom, line 127:
```suggestion
            const response = await fetchFn(args.url ?? url, args)
```

src/utils/rpc/http.ts
jxom: please also place in alphabetical order too in this object.

PR #3899, opened 2025-08-29
BXN testnet and mainnet chain update

BlackFortGroupOffice:
This PR updates definition of BXN testnet and mainnet chains

Source: https://chainlist.org/?search=bxn&testnets=true


jxom:
Please do not override chain IDs, they need to be new chains.

PR #3902, opened 2025-08-29
Update sonicTestnet.ts id

matterattetatte:
https://chainlist.org/?search=sonic&testnets=true

Apparently the sonic test network has id 14601.

I encountered this when I was developing an app using this viem npm package.


jxom:
Please do not override chain IDs, they need to be new chains.

PR #3906, opened 2025-09-01
Add Somnia Mainnet

emmaodia:


src/chains/index.ts
jxom, line 1: this should be a one line addition

PR #3936, opened 2025-09-08
Areum changes

emrahsky:
It covers the revision of change information during the Areon rebranding process. It covers the removal of the Areon testnet since it is not active.


jxom:
Please don't remove chains. `/** @deprecate */` them instead.

PR #3960, opened 2025-09-18
breaking (celo) removes alfajores testnet

aaronmgdr:
The Celo Alfajores testnet will shutdown when its L1 Holsky reaches its end of life. Therefore we are removing alfajores from the chain definitions.


aaronmgdr:
recommended to merge when the holeśky definition itself is removed.

src/chains/definitions/celoAlfajores.ts
jxom, line 1: Please add `/** @deprecated */` instead of removing it, and we will remove it in the next major version.

jxom:
Please fix outstanding comment, and then reopen the PR.

PR #3980, opened 2025-09-26
feat: add plasma multicall3 data

xzilja:
Add's `multical3` for plasma chains


xzilja:
@jxom having trouble running changelog command for this, any chance you could add one in 🙏 and could we get this in afterwards. Many protocols asking for multicall on plasma for viem / wagmi.

PR #3981, opened 2025-09-26
Fix handling of empty WebSocket messages

maxandron:
I noticed that we are getting intermittent errors on some networks, specifically observed on Gnosis and Polygon when connecting to an Erigon client using a websocket transport (so I'm assuming this is specific to Erigon). This error happens in about 1% of requests sent - randomly.
After debugging, I realized the problem was that the server was sending an empty string as a message on the connection, which viem then attempts to parse into JSON and fails with `Unexpected end of JSON input`

The problem is that Viem attributes this failure to one of the requests and makes it fail.
I then tested whether ignoring the message makes the original request succeed, and it does.

So I wrote a test that reproduces the issue and then added handling to simply ignore empty messages or messages that just contain a new line.
This works because the node proceeds to send the correct response immediately following the empty message.

Screenshot 2025-09-26 at 18 39 47

credit to The.Lows on Discord - I found his message from 1/27/25, and that helped me pinpoint and validate the issue

### DIFF
--- a/.changeset/empty-ws-messages.md
+++ b/.changeset/empty-ws-messages.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed handling of empty websocket messages from Erigon clients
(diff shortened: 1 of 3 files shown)


dvir1994:
champ! 💪

PR #3995, opened 2025-10-02
add hyperliquid evm mainnet

T3KRAJ:
Added Hyperliquid EVM Mainnet to Viem


usame-algan:
@jxom any timeline on when this will be available?

PR #4010, opened 2025-10-08
Update 0g test net chain id (16601->16602)

hejtao:


jxom:
Please deprecate the chain, and add a new chain instead of replacing chain IDs.

PR #4013, opened 2025-10-09
Add network property to Base chain object

gitmazya:
Adds `network` to base chain object so `base.network` imported from `viem/chains` will return 'base' value.


jxom:
`network` isn't a part of the `Chain` type.

PR #4022, opened 2025-10-14
feat: add EIP-7594 (PeerDAS) blob support

### LINKED ISSUE #4021: EIP-4844 Blob Transactions Failing on Sepolia After Fusaka Upgrade
### Check existing issues

### Viem Version

2.38.2

### Current Behavior

After the Fusaka upgrade on Sepolia, sending EIP-4844 transactions fails with the error:

`legacy blob tx is not supported`

After investigation, it seems that the way blob proofs are computed has changed. Previously, we used compute-blob-kzg-proof, but now it seems we should use compute_cells_and_kzg_proofs.

However, the output of compute_cells_and_kzg_proofs is no longer a single hex string—it returns an array of arrays. This raises a question: do the blob-related fields in the transaction format need to be updated to reflect this new structure?

Requesting clarification or guidance on how EIP-4844 transactions should be constructed post-Fusaka on Sepolia.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

spalladino:
Implements EIP-7594 blob transaction support alongside existing EIP-4844 blobs.

Fixes #4021

BREAKING CHANGES:
- `blobsToProofs` now returns `Hex[][]` or `ByteArray[][]` instead of flat arrays
  - EIP-4844: Returns `[[proof]]` (one proof per blob, wrapped in array)
  - EIP-7594: Returns `[[proof1, ...proof128], ...]` (128 cell proofs per blob)
- `BlobSidecar.proof` type now accepts `type | type[]` to support both formats

Features:
- Add optional `blobVersion` parameter ('4844' | '7594') to blob-related methods
- Automatic blob version detection based on chain ID (Sepolia uses EIP-7594)
- Add `getBlobVersion` helper for chain-based version selection
- Support for `computeCellsAndKzgProofs` in KZG interface
- EIP-7594 uses wrapper version byte `0x01` in transaction serialization
- Transaction parsing handles both 4-element (EIP-4844) and 5-element (EIP-7594) wrappers

Implementation:
- `blobsToProofs`: Returns array of proof arrays, uses `computeCellsAndKzgProofs` for EIP-7594
- `toBlobSidecars`: Handles proof arrays, unwraps single proofs for EIP-4844
- `serializeTransaction`: Flattens proof arrays, adds version byte for EIP-7594
- `parseTransaction`: Groups proofs by blob dynamically based on count
- `getBlobVersion`: Auto-detects version from chain ID with override support

Documentation:
- Update blob transaction guide with EIP-7594 information
- Add `blobVersion` parameter docs to `blobsToProofs` and `toBlobSidecars`
- Document proof structure differences between EIP-4844 and EIP-7594

Dependencies:
- Update `@paulmillr/trusted-setups` to 0.3.0 (PeerDAS support)
- Update `micro-eth-signer` to 0.17.3 (includes `computeCellsAndProofs`)


spalladino:
I see a bunch of tests are failing, my guess is because the PR does not have access to env vars for executing actual requests? Let me know if there's anything I should adjust to make the CI less angry.

spalladino:
I've tested this manually using the `sendRawTransaction.test.ts` test against both [Sepolia](https://sepolia.etherscan.io/tx/0x195b669fff85b8c4583dfa59bbf189d6c5020bd768a0f417d519b138f2a11c5c) and [Hoodi](https://hoodi.etherscan.io/tx/0x5c555ed2a1f37542511726b524cf02eb5595737679631a77a44678093f1b9923) (since it has not yet activated Fusaka), and both worked fine.

src/utils/blob/getBlobVersion.ts
```diff
@@ -0,0 +1,54 @@
...
+export function getBlobVersion(
+  parameters: GetBlobVersionParameters = {},
+): GetBlobVersionReturnType {
+  const { chainId, blobVersion } = parameters
+
+  // If explicit version provided, use it
+  if (blobVersion) return blobVersion
+
+  // Sepolia uses EIP-7594
+  if (chainId === 11_155_111) return '7594'
```
spalladino: Not sure if there's a cleaner version for doing this. We could add a new setting to `src/chains/definitions/sepolia.ts` and all other chains that have implemented this fork. Still, this would require updating viem again on the time the fork kicks in. Is there a precedent for handling breaking JSON RPC API changes at a fork?

spalladino:
Hey @jxom, any news on this one? Fusaka is closing in on mainnet, and without this any blob tx will fail (unless it targets an execution client that backfills pre-fusaka blob proofs).

StackOverflowExcept1on:
@jxom can you please review this and merge?

jxom:
Resolved on current v3 prerelease: https://v3.viem.sh/

spalladino:
@jxom given v3 is still not stable, and it's a breaking API change for clients, would you consider accepting this PR for v2? I think it could be simplified by dropping support for sending old 4844 txs at this point.

jxom:
v3 is pretty stable now. I don't expect much major changes from now until launch.

spalladino:
Thanks! Given it was tagged as pre release, I thought it was still under heavy dev.

PR #4023, opened 2025-10-15
Deprecate the old and add new 0g Galileo test net (chaiId 16602)

hejtao:


jxom:
https://github.com/wevm/viem/pull/4010#issuecomment-3393809897

PR #4024, opened 2025-10-15
Add blockTime property to optimism chain definition

boredland:
According to the [docs](https://docs.optimism.io/connect/resources/glossary#block-time), the block time for optimism should be 2 seconds.


jxom:
already covered in op stack chain config

PR #4025, opened 2025-10-15
Add blockTime property to base chain definition

boredland:
The regular blockTime is 2s, the flashblocks are emitted at 200ms, but are only available at certain flashblock-aware RPCs.

Docs: https://docs.base.org/base-chain/flashblocks/docs#is-there-any-difference-in-tx-inclusion-for-flashblocks-vs-2s-block%3F


jxom:
already covered in op stack chain config

PR #4026, opened 2025-10-15
Add blockTime property to celo chain definition

boredland:
Block time is 1 second according to the docs: https://docs.celo.org/legacy/faq#whats-the-block-time.


jxom:
already covered https://github.com/wevm/viem/blob/main/src/celo/chainConfig.ts#L7

PR #4037, opened 2025-10-16
chore: update Botanix chain spec

jeffgeudens:
Added Multicall3 contract to Botanix chain


jxom:
Chain not listed on https://www.multicall3.com/deployments

PR #4058, opened 2025-10-28
fix: clearTimeout in waitForUserOperationReceipt

ghost:
This PR fixes a bug where the timeout in `waitForUserOperationReceipt` was not properly cleared after the receipt was returned or an error occurred. This could cause unnecessary delayed executions or memory leaks in long-running applications.

**What this PR does:**

- Calls clearTimeout(timer) once the promise resolves or rejects.
- Ensures the timer is properly cleaned up to prevent leaks and race conditions.

**Why it matters:**
Without clearing the timeout, the process may retain references to the pending timer callback. For example, when `waitForUserOperationReceipt` is used in a `vitest` test, the test may hang for up to 120,000 ms (the default timeout) because the testing framework detects ongoing I/O operations that were never cleared.

### DIFF
--- a/.changeset/clever-buttons-cut.md
+++ b/.changeset/clever-buttons-cut.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Fix a bug where the timeout in `waitForUserOperationReceipt` was not properly cleared after the receipt was returned.
+Fixed a bug where the timeout in `waitForUserOperationReceipt` was not properly cleared after the receipt was returned.


jxom:
Please add a changeset.

```
pnpm changeset
```

ghost:
> Please add a changeset.
> 
> ```
> pnpm changeset
> ```

@jxom  Changeset was added

ghost:
@jxom any update mate?

PR #4060, opened 2025-10-28
fix: update jovay default explorer url and default rpc endpoint

zoowii:


zoowii:
There is updates of network config (jovay network config in https://docs.jovay.io/developer/network-information)

PR #4069, opened 2025-10-31
feat(errors): add url to request errors

bearpong:
In order to help better error handling, url is added as property on request errors. 

This helps project understand which rpc failed. This is important as clients might be using `fallback` as transport without knowing which rpc is actually being used and which one is throwing timeout or http errors. 

This makes all errors consistent with `HttpRequestError` that exposes the url as class property.


bearpong:
@jxom any chance to review this? It's an easy one, it just bubbles up more info in errors.

PR #4073, opened 2025-11-02
Add POTOS Mainnet

akitothemoon:


kyonRay:
Thank you for helping add the POTOS mainnet information to viem. It means a lot :)

PR #4076, opened 2025-11-03
chore: Update Xone mainnet name issue

mangosago99:


mangosago99:
@jxom In the current task, I saw the message `npm i https://pkg.pr.new/viem@4076`, and I'm waiting for a solution.

PR #4077, opened 2025-11-03
feat: deprecate old Taiko testnets and add Taiko Hoodi

bearni95:
- Deprecated `taikoHekla`, `taikoJolnir`, `taikoKatla`, and `taikoTestnetSepolia` chains
- Added new `taikoHoodi` chain (Taiko Hoodi L2 testnet)
- All deprecated chains remain functional for backward compatibility
- Users will see deprecation warnings guiding them to use `taikoHoodi`

Resolves feedback to deprecate chains instead of removing them.

### DIFF
--- a/.changeset/taiko-hoodi-deprecate-old-testnets.md
+++ b/.changeset/taiko-hoodi-deprecate-old-testnets.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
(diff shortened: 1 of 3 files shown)


jxom:
Please reopen once chain is on https://github.com/ethereum-lists/chains

PR #4091, opened 2025-11-14
Add multicall3 to Hemi and HemiSepolia

jcvernaleo:
Adds multicall3 to Hemi and HemiSepolia chain definitions.


jxom:
Chains are not listed on https://www.multicall3.com/deployments.

PR #4093, opened 2025-11-16
Update `megaethTestnet` for MegaETH Testnet v2

davidperk:
Chain ID `6342` is now deprecated in favor of `6343` per https://t.me/c/2399063148/3898.


jxom:
Please `/** @deprecate */` this chain, and create a new one.

PR #4105, opened 2025-11-20
fix: use new domain for explorer

QEDK:
Uses the correct domain for explorer.


keone:
please close this in favor of #4109

QEDK:
See #4109

PR #4111, opened 2025-11-22
chore: updated megaeth testnet config to "timothy"

CodingTux:
MegaETH recently switched their testnet to timothy, this update is to support that since it is breaking lot of stuff during development


jxom:
Please `/** @deprecate */` the chain and add a new one, then reopen the PR.

PR #4113, opened 2025-11-24
feat: add HyperEVM mainnet definition and update testnet one

gabrielstoica:
Adds HyperEVM Mainnet definition and renames the existing `hyperliquidEvmTestnet` to `hyperevmTestnet` to standardise.


jxom:
Resolve comment and then reopen PR.

Issue #4117, opened 2025-11-25
attemptFill treats undefined support status as “supported”, causing unwanted eth_fillTransaction calls


dbarczewski-circle:
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

> 2.40.0

### Current Behavior

```
const attemptFill =
  // Do not attempt if `eth_fillTransaction` is not supported.
  supportsFillTransaction.get(client.uid) !== false &&
  ['fees', 'gas'].some((parameter) =>
    args_.parameters?.includes(
      parameter as PrepareTransactionRequestParameterType,
    ),
  )
```

When supportsFillTransaction.get(client.uid) returns undefined (meaning not supported), the check !== false still passes (true).

Consequently, attemptFill becomes true, and eth_fillTransaction is executed against clients that don’t implement it.

### Expected Behavior

If a client does not support eth_fillTransaction (i.e. the map entry is undefined), we should not attempt to use it.
attemptFill should only be true if support is defined (either true or explicitly set).

Corrected logic
```
const attemptFill =
  supportsFillTransaction.get(client.uid) !== undefined &&
  ['fees', 'gas'].some((parameter) =>
    args_.parameters?.includes(
      parameter as PrepareTransactionRequestParameterType,
    ),
  )
```

### Steps To Reproduce

1. Use a client where eth_fillTransaction is not supported.
2. Attempt to send a transaction
3. See that the code still tries to call `eth_fillTransaction`

### Anything else?

prepareTransactionRequest was attempting to call eth_fillTransaction even when the client didn’t support it.
This happened because supportsFillTransaction.get(client.uid) returns undefined for unsupported clients, but the check only excluded false.

I can put up a PR to fix the condition to check for `undefined`.

jxom:
The logic is correct. There will be an attempt to use `eth_fillTransaction`, but if the node returns a “Method not supported” error (or it fails with some other reason), then Viem will fall back to the old route and not use `eth_fillTransaction` again.

The `prepareTransactionRequest` call should never fail, and the flow should still work as it did pre-2.40. The error [is not rethrown](https://github.com/wevm/viem/blob/main/src/actions/wallet/prepareTransactionRequest.ts#L311-L319) and it should **always fall back**.

jxom:
Closing as this is not a bug.

If you don't want the client to make the request to `eth_fillTransaction`, please amend your transport to include `methods`:

```ts
[...]

PR #4118, opened 2025-11-25
fix: Fix fill transaction pre check

dbarczewski-circle:
Related to #4117 

`supportsFillTransaction.get(client.uid)` returns `undefined` if `fillTransaction` is not supported. 
Since `undefined !== false` is true, the whole expression `attemptFill` becomes true if `gas` or `fees` are required. 

This PR adjusts the logic and correctly uses `undefined` now. 

Note: Also the TSDoc for `LruMap.get` states that: 
>@returns — Returns the element associated with the specified key. If no element is associated with the specified key, undefined is returned.

### DIFF
--- a/.changeset/stale-planets-tease.md
+++ b/.changeset/stale-planets-tease.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fix checkup for supporting fillTransaction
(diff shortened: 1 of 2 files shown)


jxom:
See https://github.com/wevm/viem/issues/4117#issuecomment-3576692985

PR #4123, opened 2025-11-26
docs: fix malformed request example in public client docs

Alyaesub:
This PR fixes a small typo in the Public Client documentation.

The current request example contains an incomplete method string:

  method: 'eth_wa

This breaks the code snippet and makes the example invalid TypeScript.

It has been corrected to:

  method: 'eth_wagmi'

This makes the snippet syntactically valid and copy-paste ready.

Let me know if you'd like any adjustments!


jxom:
intentional.

PR #4188, opened 2025-12-19
fix: tempo e2e

jenpaff:
we made recent changes in the [gas schedule for access keys](https://github.com/tempoxyz/tempo/pull/1413/files#diff-15963b1f3be40468c188dd9550dffff7b46e28e1b4d5269a903ad0bc44e39d94), and therefore[ this test fails ](https://github.com/wevm/viem/actions/runs/20369938570/job/58533861186?pr=4187#step:5:2383)

this PR adds 2 new fields for gas estimate

running tests tempo e2e tests locally works

```
VITE_TEMPO_ENV=devnet pnpm test --project tempo e2e
 
 Test Files  1 passed (1)
      Tests  30 passed (30)
   Start at  18:52:31
   Duration  35.08s
 ```

### DIFF
--- a/src/tempo/Formatters.ts
+++ b/src/tempo/Formatters.ts
@@ -125,21 +125,9 @@ export function formatTransactionRequest(
     account && 'accessKeyAddress' in account
       ? account.accessKeyAddress
       : undefined
-  const keyAuthorization = rpc.keyAuthorization
-    ? {
-        ...rpc.keyAuthorization,
-        // FIXME(ox): remove once https://github.com/tempoxyz/tempo/pull/1683 is deployed.
-        chainId:
-          rpc.keyAuthorization.chainId === '0x' ||
-          rpc.keyAuthorization.chainId === '0x0'
-            ? 0
-            : parseInt(rpc.keyAuthorization.chainId || '0x0', 16),
-      }
-    : undefined
 
   return {
     ...rpc,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/tempo/Formatters.ts [hunk 1]
```diff
@@ -121,10 +121,44 @@ export function formatTransactionRequest(
     return [undefined, undefined]
   })()
 
+  const keyId =
+    account && 'accessKeyAddress' in account
+      ? account.accessKeyAddress
+      : undefined
+  const keyAuthorization = rpc.keyAuthorization
+    ? (() => {
+        const keyAuthorization = {
```
jxom: I don’t think we need to coerce this at all as `TransactionRequest.toRpc` already handles key auths. My question is that does the RPC expect a different format for chain ids and `prehash` naming? Guess we can update the prehash naming in Ox though

jxom:
Just updated Ox for the `preHash` change. Also added temporary [`prepareTransactionRequest` hook for 2D nonce gas](https://github.com/wevm/viem/pull/4188/commits/0401df5fce0ddc87b8d7b794b7f2bbffc504ef94#diff-50730ad1ad52026854ef0ce03803a043a24bae4139fcbd689846405fc59decf5R43-R48) until `eth_fillTransaction`/`eth_estimateGas` is fixed for 2D nonces. Though, Tempo devnet tests are failing because the RPC is currently down.

PR #4195, opened 2025-12-26
fix: validate bytes length in getAbiItem

0xRenji:
Fixed getAbiItem for different overloaded bytes lengths. Modified existing test to include this case.


mkenshin1688-collab (approved):
Tank

PR #4198, opened 2025-12-29
fix: getUserOperationHash calculation for EIP-7702 UserOperations (#4197)fix: correct EIP-7702 UserOperation hash calculation

### LINKED ISSUE #4197: Incorrect getUserOperationHash calculation causes signature errors.
### Check existing issues

### Viem Version

2.43.1

### Current Behavior

`getUserOperationHash()` doesn't handle properly different EIP-7702 "factory" settings.
It returns different values than the `EntryPoint.getUserOperationHash()`, and thus cause validation to fail.

The toPackedUserOperation should behave differently when "just packing" (for the purpose of making an ABI call) and "for signing".
- When packing it should pack the `factory` as-is, even if it is only 2 bytes (`0x7702`), not padding it to 20 bytes
- When signing it should replace factory (which is a `0x7702` with possible trailing zeros) with the current delegate (this is a mechanism that ensures a UserOp is never executed against a delegate that it wasn't explicitly signed for)

### Expected Behavior

_No response_

### Steps To Reproduce

I suggest testing the implementation of `getUserOperationHash()` against the values of the on-chain `EntryPoint.getUserOperationHash()`, to make sure they are aligned.
Unfortunately, I couldn't find a test against a contract.
e.g the tests [here](https://github.com/eth-infinitism/account-abstraction/blob/develop/test/entrypoint-7702.test.ts#L133)

_No response_

### Anything else?

_No response_

srivtx:
fixes #4197

## Description
This PR fixes the [getUserOperationHash](cci:1://file:///viem/src/account-abstraction/utils/userOperation/getUserOperationHash.ts:23:0-129:1) calculation for EIP-7702 UserOperations.

Previously, `viem` was substituting the factory address with the delegate address (`authorization.address`) during both packing (bundling) and hashing (signing).

This caused a mismatch because:
1. **Bundling**: Requires the raw factory address `0x7702` (2 bytes) to be present in `initCode` so the chain recognizes it as an EIP-7702 transaction.
2. **Signing/Hashing**: Requires the specific delegate address to be hashed in `initCode` for security.

## Changes
- Modified [getInitCode](cci:1://file://viem/src/account-abstraction/utils/userOperation/getInitCode.ts:3:0-20:1) to accept an `options.forHash` parameter.
- Updated [toPackedUserOperation](cci:1://file:///viem/src/account-abstraction/utils/userOperation/toPackedUserOperation.ts:9:0-67:1) to pass this option through.
- Updated [getUserOperationHash](cci:1://file://open/viem/src/account-abstraction/utils/userOperation/getUserOperationHash.ts:23:0-129:1) (v0.6, v0.7) and [getUserOperationTypedData](cci:1://file:////viem/src/account-abstraction/utils/userOperation/getUserOperationTypedData.ts:30:0-48:1) (v0.8, v0.9) to pass `{ forHash: true }`.

## Verification
- Ran existing tests successfully: 
  `pnpm test src/account-abstraction/utils/userOperation/getUserOperationHash.test.ts`
- Verified that v0.6, v0.7, v0.8, and v0.9 EntryPoints all calculate the correct hash while enabling correct packing for Bundlers.

### DIFF
--- a/.changeset/fuzzy-cats-jump.md
+++ b/.changeset/fuzzy-cats-jump.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `getUserOperationHash` calculation for EIP-7702 UserOperations.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6864,6 +6864,14 @@ packages:
       typescript:
         optional: true
 
+  viem@2.43.3:
+    resolution: {integrity: sha512-zM251fspfSjENCtfmT7cauuD+AA/YAlkFU7cksdEQJxj7wDuO0XFRWRH+RMvfmTFza88B9kug5cKU+Wk2nAjJg==}
+    peerDependencies:
+      typescript: '>=5.0.4'
+    peerDependenciesMeta:
+      typescript:
+        optional: true
+
   viem@file:src:
     resolution: {directory: src, type: directory}
     peerDependencies:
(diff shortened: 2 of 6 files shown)


jxom:
Superseded by 892ebea16241fd4403c691022297463e2fad2d5b (your commit is still preserved).

PR #4205, opened 2026-01-03
chore: use Tempo testnet tag for changesets verification

struong:
### motivation

Tempo tests in the changesets CI workflow were failing because they were using the `latest` Docker image tag for prool, which currently has breaking changes with what's in testnet. This blocks version cuts. 

Following a similar pattern from [tempo-apps#363](https://github.com/tempoxyz/tempo-apps/pull/363), we can fetch the current testnet Tempo version and use that for testing to ensure compatibility.

### change

- Added `get-tempo-testnet-tag` job to the changesets workflow that fetches the current testnet Tempo image SHA
- Updated the `verify` workflow to accept an optional `tempo-tag` input parameter
- Test job in `verify` consumes testnet tag when available, otherwise defaults to `latest`

### what's next?

If non-changeset tests should also run against current testnet we can apply the same strategy in the verify workflow and/or local tests as well.

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -9,12 +9,31 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  get-tempo-testnet-tag:
+    name: Get Testnet Tempo Version
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.fetch-tag.outputs.tag }}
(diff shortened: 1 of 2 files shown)


jxom:
Superseded by https://github.com/wevm/viem/commit/052d669205f7bec1bc4bcbafd64fc14ad768073c

PR #4224, opened 2026-01-09
fix(tempo): include keyData for access key gas estimation

struong:
### motivation

Fee-sponsored transactions using **access keys** (P256 keys granted via `grantAccessKey: true`) are reverted with "out of gas" errors. The same payment flow works correctly when using WebAuthn directly without access keys.

**Repro:** https://github.com/tempoxyz/tempo-ts/compare/struong/repro-fp-gas-error

> **Note:** Used AI to try and figure out what the root cause was, this is what it found. Thought process was
> 1. Why is there not enough gas?
> 2. How does viem set gas amount on transaction before submission?
> 3. Should fix be in viem? Or in estimateGas / "helper" RPCs that determine transaction gas amount?
> 
> AI findings below
> Original repro gist: https://gist.github.com/struong/81a31584b6f280ce95c14ba8949fcfba

The root cause is that gas estimation does not account for the **keychain signature envelope** overhead when using access keys. Access key signatures are wrapped in a keychain envelope that adds 21 bytes (type byte + 20-byte userAddress), requiring ~3,336 additional gas for validation.

### change

- Modified `formatTransactionRequest` in `src/tempo/Formatters.ts` to include `keyData` when an access key is detected (`keyId` present)
- The `keyData` is set to 151 bytes to represent the full keychain envelope: type(1) + userAddress(20) + inner P256 signature(130)
- This ensures `eth_estimateGas` correctly accounts for the keychain signature size and validation overhead
- Added two unit tests to verify the behavior

**Before:** P256 access keys returned `keyType` and `keyId` but no `keyData`, causing underestimated gas.

**After:** P256 access keys return `keyType`, `keyId`, and a 151-byte `keyData` dummy for proper gas estimation.

### testing

- Added unit test: `behavior: access key account populates keyType, keyId, and keyData` ✓
- Added unit test: `behavior: p256 account without access key does not populate keyData` ✓
- End-to-end verification with the payments example using `grantAccessKey: true`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -210,7 +210,7 @@ importers:
     dependencies:
       viem:
         specifier: latest
-        version: 2.43.5(typescript@5.6.2)(zod@3.25.76)
+        version: 2.43.4(typescript@5.6.2)(zod@3.25.76)
     devDependencies:
       typescript:
         specifier: ^5.0.3
@@ -223,7 +223,7 @@ importers:
     dependencies:
       viem:
         specifier: latest
-        version: 2.43.5(typescript@5.6.2)(zod@3.25.76)
+        version: 2.43.4(typescript@5.6.2)(zod@3.25.76)
     devDependencies:
       typescript:
         specifier: ^5.0.3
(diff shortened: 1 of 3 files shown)


struong:
Consumed https://pkg.pr.new/viem@4224 in the tempo-ts repro branch and still getting the same problem. Closing for now

PR #4228, opened 2026-01-11
perf: decode logs

sakulstra:
We have an application that indexes a huge amount of logs and when benchmarking where time is spent, i noticed that the decoding alone takes roughly 10% of the time. Therefore, i started looking into how to improve performance here.

Running against this version locally, seems to improve performance by roughly 25% for us.
Still probably there is more things to optimize.

Opening this more for discussion than to necessarily merge.
Our usecase essentially is: `decoding a huge amounts of logs with a limited set of abiEvents (perhaps 20)`.

As i understand the bench result it's around x3 as fast?
In our realworld example (which does more things than just decoding), these are the results:

Before:
```
✅ Run 1 completed:
   Decoding time: 11.95s
   Total logs: 364674
```

After:
```
✅ Run 1 completed:
   Decoding time: 7.00s 
   Total logs: 364674
 ```

### DIFF
--- a/.changeset/loud-mails-greet.md
+++ b/.changeset/loud-mails-greet.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Slightly improved performance in `parseEventLogs` trough memoization.
+Improved performance in `parseEventLogs` by memoizing.


src/utils/abi/parseEventLogs.ts
jxom: I would be happy to merge this in if it makes things faster.

sakulstra:
@jxom anything i can do to move this forward?

PR #4324, opened 2026-02-06
Fix gasPrice leaking into EIP-1559 transactions

### LINKED ISSUE #4323: `prepareTransactionRequest` leaks `gasPrice: 0n` into EIP-1559 transactions when RPC supports eth_fillTransaction
### Check existing issues

### Viem Version

2.43.3

### Current Behavior

`prepareTransactionRequest` returns a transaction request with `gasPrice: 0n` alongside `type: "eip1559"`, `maxFeePerGas`, and `maxPriorityFeePerGas`.

When this request is passed to a MetaMask, it rejects it with:

`invalid transaction envelope type: specified type "0x2" but included a gasPrice instead of maxFeePerGas and maxPriorityFeePerGas`.

### Expected Behavior

`prepareTransactionRequest` should not include `gasPrice` in the returned request when type is `eip1559`.

### Steps To Reproduce

1. Use an Arbitrum RPC that supports `eth_fillTransaction` (e.g., arbitrum-one.publicnode.com)
2. Call `prepareTransactionRequest` with a basic EIP-1559 transaction
3. Observe `gasPrice: 0n` in the returned object
4. Pass the result to sendTransaction — MetaMask rejects it

  The RPC returns gasPrice: "0x0" for EIP-1559 transactions. This can be verified directly:

```bash
  curl -s -X POST https://arbitrum-one.publicnode.com \
    -H "Content-Type: application/json" \
    -d '{
      "jsonrpc": "2.0",
      "id": 1,
      "method": "eth_fillTransaction",
      "params": [{
        "from": "0x0000000000000000000000000000000000000001",
        "to": "0x0000000000000000000000000000000000000001",
        "value": "0x0",
        "type": "0x2"
      }]
    }'
```

  Response:

```
  {
    "result": {
      "tx": {
        "type": "0x2",
        "gasPrice": "0x0",
        "maxFeePerGas": "0x2e28680",
        "maxPriorityFeePerGas": "0x0",
        "gas": "0x5f12",
        "chainId": "0xa4b1"
      }
    }
  }
```

_No response_

### Anything else?

  - viem: 2.43.3
  - wagmi: 3.1.3
  - Chain: Arbitrum One (42161)
  - RPC: arbitrum-one.publicnode.com (supports eth_fillTransaction)
  - Wallet: MetaMask

t0rbik:
## Summary
- Fixes #4323
- `formatTransaction` now resets `gasPrice` to `undefined` for non-legacy/non-eip2930 transaction types, preventing `gasPrice: 0n` from leaking into EIP-1559 transactions when the RPC supports `eth_fillTransaction`

## Test plan
- [x] Added inline snapshot test verifying `gasPrice` is reset to `undefined` when an EIP-1559 transaction includes a leaked `gasPrice` from RPC
- [x] All existing `formatTransaction` tests pass

### DIFF
--- a/.changeset/tidy-snakes-type.md
+++ b/.changeset/tidy-snakes-type.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Fixed gasPrice leaking into EIP-1559 transactions when formatted by RPC
+Fixed `gasPrice` leaking into EIP-1559 transactions when formatted by RPC.


jxom:
Hmm, I am worried this will cause more harm than good. This is a breaking change and may affect some assumptions from consumers. I'd consider this a bug in MM because they should gracefully prefer maxFeePer*.

t0rbik:
Alright, sounds good. I will open an issue in metamask org. We can always workaround on client as well. 
Thank you for review!

PR #4329, opened 2026-02-09
feat: add getBalance batching

sakulstra:
- closes https://github.com/wevm/viem/issues/4328

The idea is that:
when using a batch client, and having access to multicall3 it's basically always preferable to get a balance via eth_call so it can be batched with other operations.

An alternative approach could have been to have a `batch` flag or similar. Lmk what you think.

---

Another alternative is obviously to not do it via viem, but to force app developers to just do rpc call to multicall3 manually, but it feels like it could/should be automatic.

### DIFF
--- a/src/actions/public/getBalance.ts
+++ b/src/actions/public/getBalance.ts
@@ -1,47 +1,43 @@
-import type { Address } from "abitype";
+import type { Address } from 'abitype'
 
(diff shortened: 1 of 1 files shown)


meny3437-blip (commented):
npm i viem @meny3437-blip

PR #4334, opened 2026-02-10
feat(chains): update seismic testnet RPC url for chain ID 5124

cdrappi:
We launched our testnet. In light of this, we deprecated Seismic devnet 2 (node-2.seismicdev.net)

This commit adds the `seismicTestnet` network. To avoid breaking people's imports, we aliased `seismicDevnet` to this new Seismic testnet

The chain ID is the same (5124) as before, and the RPC paths are the same: /rpc (https) and /ws (wss)


jxom:
Chain needs to be exported from `src/chains/index.ts`

PR #4336, opened 2026-02-11
feat(chains): Add Gensyn Mainnet

abs3ntdev:


jxom:
Thanks for the contribution! Unfortunately we can't merge this at this time — the chain is listed as `status: "incubating"` on [ethereum-lists/chains](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-685689.json) with no RPCs or explorers listed, and the block explorer URL (`ddrg456.explorer.alchemy.com`) appears to be an auto-generated/temporary Alchemy slug rather than a permanent canonical URL.

We'd be happy to revisit once:
1. The chain status is updated on ethereum-lists (no longer "incubating")
2. A permanent/branded block explorer URL is available
3. The `sourceId: 1` claim can be confirmed (L2 settling to Ethereum mainnet with bridge contracts)

Feel free to reopen when ready!

PR #4337, opened 2026-02-12
feat: batch matching stateOverrides in call

haydenshively:
When `client.batch.multicall` is enabled, `call()` skips batching entirely for any call that includes a `stateOverride`. However, the downstream multicall action already supports forwarding `stateOverride` to `eth_call`, so there's no fundamental reason calls sharing the same `stateOverride` can't be batched together.

This PR removes the blanket `!rpcStateOverride` guard and instead groups batched calls by their serialized `stateOverride` value. Calls with the same override batch together; calls with different overrides (or no override) go to separate batches. The `stateOverride` is then forwarded as the third param in the batched `eth_call`.

  Changes:
  - Removed the `!rpcStateOverride` check that prevented all state-overridden calls from batching
  - Forward `rpcStateOverride` through `scheduleMulticall` and include it in the batched `eth_call` params
  - Include `JSON.stringify(rpcStateOverride)` in the batch scheduler ID so calls are grouped by identical overrides
  - Added test covering same-override batching and different-override separation


haydenshively:
@jxom is there anything I can do to help get this merged?

haydenshively:
bump, how can I help?

spennyp:
+1 would be great for this to be supported

Issue #3875, opened 2025-08-20
waitForTransactionReceipt does not work as expected


Pasha8914:
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.34.0

### Current Behavior

The waitForTransactionReceipt function does not handle the scenario where a transaction is no longer found on the network (e.g., because it was replaced by a newer one with the same nonce and a higher gas price, or it was simply dropped from the mempool). Instead of rejecting the promise with a clear error

**First Issue**: The function cannot find transactions that have been replaced or dropped
**Second Issue**: Regardless of the configuration (e.g., timeout, retryCount), the function's promise never settles (neither resolves nor rejects) if it cannot locate the transaction hash on the blockchain

### Expected Behavior

waitForTransactionReceipt should detect replaced transactions or throw an error

### Steps To Reproduce

This app has two buttons:

- **Send 1 wei to yourself** — submits a self-transfer with intentionally tiny EIP-1559 fees so it stays pending.
- **Speed Up** — resubmits the **same nonce** with higher fees, replacing the first transaction.

## How to reproduce

1. Click **Send 1 wei to yourself**.  
   - The tx uses `maxFeePerGas = 2` and `maxPriorityFeePerGas = 1` (wei), so it will **not get mined** on Sepolia. It remains pending indefinitely.
   - `waitForTransactionReceipt` for this tx never resolves with a receipt (by design), because the transaction won’t be included in a block.

2. After a short wait, click **Speed Up**.  
   - The app sends a **replacement tx** with the **same nonce** and **higher fees** (roughly `2×` the current `estimateFeesPerGas` values).
   - This replacement **supersedes** the first tx. You’ll see a **receipt in the console only for the second (speed-up) transaction**.
   - The original tx will **never** produce a receipt, because it was replaced before inclusion.

3. If you **don’t** click **Speed Up**, nothing will happen: the first tx stays pending forever and no receipt is printed.

https://github.com/user-attachments/assets/3452c1fc-abac-4226-b928-b18f9c7985df

### Link to Minimal Reproducible Example

https://github.com/Pasha8914/waitForTransactionReceipt-bug

jxom:
[...]

PR #4342, opened 2026-02-15
fix(waitForTransactionReceipt): detect dropped transactions instead of polling indefinitely

Kropiunig:
## Problem

`waitForTransactionReceipt` polls indefinitely when a transaction has been dropped from the mempool. This happens in a common real-world scenario during gas spikes:

1. User sends a transaction with low gas fees
2. Network congestion causes the transaction to be evicted from the mempool
3. `getTransaction(hash)` starts returning `TransactionNotFoundError` (the node no longer knows about the pending tx)
4. `getTransactionReceipt(hash)` also fails (tx was never mined)
5. The catch block at the `!transaction` check silently `return`s, continuing the poll loop
6. **Result**: Every block triggers a full `withRetry` cycle (default 6 retries with exponential backoff ≈ 13s), producing ~6 wasted RPC calls per cycle, until the 180s timeout fires with a generic `WaitForTransactionReceiptTimeoutError`

If `timeout` is set to `0` / `undefined`, the promise **never settles**.

This is #3875. The issue also affects the "speed-up" flow: when a user replaces a dropped tx with a same-nonce tx that gets mined under a different hash, the original `waitForTransactionReceipt` can't detect the replacement because `transaction` was never populated (the evicted tx was gone before the first `getTransaction` call succeeded).

## Solution

Track consecutive blocks where neither `getTransaction` nor `getTransactionReceipt` can locate the transaction. After `retryCount` such blocks (default 6, matching the per-block retry semantics), reject with a new `TransactionDroppedError` instead of polling forever.

```
TransactionDroppedError: Transaction with hash "0x..." could not be found
on the network. The transaction may have been dropped from the mempool.

This can happen when:
- The transaction fee was too low and it was evicted from the mempool during a gas spike
- The transaction was replaced by another transaction with the same nonce

If the transaction was replaced, use `waitForTransactionReceipt` with the
new transaction hash instead.
```

### Design decisions

- **Scoped to `checkReplacement: true`** (the default): Drop detection only activates when replacement checking is enabled, because that's the code path that calls `getTransaction`. When `checkReplacement: false`, the function still falls through to the timeout — this preserves backward compatibility for callers that explicitly opted out of replacement checking.

- **Reuses `retryCount` as the block threshold**: Rather than introducing a new parameter, the existing `retryCount` (default 6) doubles as the number of consecutive blocks to tolerate before considering the tx dropped. This keeps the API surface unchanged while providing reasonable defaults (~72s on mainnet at 12s blocks).

- **No false positives for slow RPCs**: If a slow RPC eventually returns the transaction (e.g., after 2-3 blocks of lag), `transaction` gets populated and `notFoundCount` becomes irrelevant — the normal replacement detection path takes over.

## Changes

| File | Change |
|------|--------|
| `src/errors/transaction.ts` | New `TransactionDroppedError` with actionable meta messages |
| `src/actions/public/waitForTransactionReceipt.ts` | `notFoundCount` tracker + drop detection in the `!transaction` catch branch |
| `src/index.ts` | Export `TransactionDroppedError` and `TransactionDroppedErrorType` |
| `src/actions/public/waitForTransactionReceipt.test.ts` | 3 new tests: drop detection, slow-RPC tolerance, `checkReplacement: false` bypass |

4 files changed, ~150 lines added.

## Tests

**`dropped transactions > rejects with TransactionDroppedError when transaction is evicted from mempool`**
Mocks both `getTransaction` and `getTransactionReceipt` to consistently throw, simulating a fully evicted tx. Asserts the promise rejects with `TransactionDroppedError` within the retry threshold.

**`dropped transactions > does not reject prematurely when transaction appears after initial failures`**
Mocks `getTransaction` to fail twice then fall through to the real implementation. Sends and mines a real transaction. Asserts the receipt resolves successfully despite the initial not-found responses (slow RPC simulation).

**`dropped transactions > checkReplacement: false does not trigger drop detection`**
With replacement checking disabled, the function should fall through to the timeout rather than the drop detection path. Asserts `WaitForTransactionReceiptTimeoutError` is thrown, not `TransactionDroppedError`.

## Future work

The remaining gap from #3875 is **nonce-based replacement detection for evicted transactions**: when the original tx was evicted before `getTransaction` could read it, we don't know the sender address or nonce, so we can't search recent blocks for a same-nonce replacement. This could be addressed by adding optional `from` / `nonce` parameters to `WaitForTransactionReceiptParameters` in a follow-up PR.

### DIFF
--- a/.changeset/detect-dropped-transactions.md
+++ b/.changeset/detect-dropped-transactions.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `waitForTransactionReceipt` to detect dropped/replaced transactions instead of polling indefinitely.
--- a/src/actions/public/waitForTransactionReceipt.test.ts
+++ b/src/actions/public/waitForTransactionReceipt.test.ts
@@ -3,7 +3,10 @@ import { anvilMainnet } from '~test/anvil.js'
 import { accounts } from '~test/constants.js'
 import { privateKeyToAccount } from '../../accounts/privateKeyToAccount.js'
 import { prepareTransactionRequest } from '../../actions/index.js'
-import { WaitForTransactionReceiptTimeoutError } from '../../errors/transaction.js'
+import {
+  TransactionDroppedError,
+  WaitForTransactionReceiptTimeoutError,
+} from '../../errors/transaction.js'
 import { hexToNumber } from '../../utils/encoding/fromHex.js'
 import { keccak256 } from '../../utils/index.js'
 import { parseEther } from '../../utils/unit/parseEther.js'
@@ -18,6 +21,7 @@ import { mine } from '../test/mine.js'
 import { sendTransaction } from '../wallet/sendTransaction.js'
 import * as getBlock from './getBlock.js'
 import * as getTransactionModule from './getTransaction.js'
+import * as getTransactionReceiptModule from './getTransactionReceipt.js'
 import { waitForTransactionReceipt } from './waitForTransactionReceipt.js'
 
 const client = anvilMainnet.getClient()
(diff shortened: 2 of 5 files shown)


Kropiunig:
Hi @jxom — pinging for visibility. This PR adds dropped-transaction detection to `waitForTransactionReceipt` so it doesn't poll indefinitely when a transaction is replaced or dropped. CI is passing. Would appreciate a review when you have a moment!

Kropiunig:
Added the missing changeset. The Test Local failures are pre-existing (same evm_setIntervalMining errors on main and in recently merged PRs like #4472). Ready for review whenever you have a moment @jxom.

jxom:
Closing this out for now. The dropped-transaction heuristic can falsely reject a transaction that is still pending if the public RPC does not expose it via eth_getTransactionByHash before the receipt appears, especially with wallet/public-client provider splits. The branch also needs to catch up with current main/export snapshots. Feel free to reopen if you want to rework the heuristic.

Kropiunig:
Thanks for the clarification here and on #3875. I saw #4636 and your `sendTransactionSync` direction — that's a cleaner long-term fix than a polling heuristic. Won't pursue a rework here. Happy to help on `sendTransactionSync` ergonomics or docs if that's useful.

PR #4369, opened 2026-02-28
fix: cache `supportsFillTransaction` for InvalidRequestRpcError (-32600)

kangarang:
Alchemy returns -32600 (InvalidRequest) instead of -32601 (MethodNotFound) or -32004 (MethodNotSupported) when `eth_fillTransaction` is unsupported.

Without recognizing this error code, viem never caches the failure, causing every write call to re-attempt `eth_fillTransaction` and return unfilled requests.

### DIFF
--- a/src/actions/wallet/prepareTransactionRequest.test.ts
+++ b/src/actions/wallet/prepareTransactionRequest.test.ts
@@ -11,9 +11,11 @@ import {
   BaseError,
   createClient,
   http,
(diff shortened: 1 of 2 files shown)


jxom:
Thanks for this PR – however `InvalidRequest` may also mean other things (e.g. invalid properties, etc) and is not a deterministic property to detect if the method is supported or not. I think this is more an Alchemy RPC issue.

PR #4386, opened 2026-03-09
improve: Add missing MegaETH OP stack defs

pxrl:
MegaETH is an almost-OP-stack. This commit adds the generic definitions that are currently lacking. Native support for finalizing withdrawals is subject to some discussion on how to handle the changes unique to MegaETH.

While here, fix the Blockcout definitions for MegaETH as well.


pxrl:
@jxom We are running with a local patch to viem 2.46.1 that allows us to finalise MegaETH withdrawals as if they were native OP stack. The main differences are:
- OP stack defs are added (as in this PR).
- Custom handling for mega_getWithdrawalProof in place of eth_getProof.
- Custom handling for extra data returned by getGames().

We'd be interested in discussing whether some or all of this could be upstreamed, or whether there's scope for making the OP stack implementations a bit more modular, such that it's easy to supplement downstream. This would potentially be applicable for backporting to Blast, which we also have support for.

Patch: https://github.com/across-protocol/relayer/blob/master/patches/viem%2B2.46.1.patch

PR #4388, opened 2026-03-09
Preserve error cause in `getContractError`

alcuadrado:
Fixes https://github.com/wevm/viem/issues/4387

I didn't manage to run the entire test suite in my local machine, sorry in advance if something fails on CI. I'll fix it.

I'll leave some comments for the reviewer in the PR as well.

### DIFF
--- a/.changeset/six-pumpkins-wait.md
+++ b/.changeset/six-pumpkins-wait.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added error preservation in `getContractError`.


src/errors/contract.ts
```diff
@@ -179,19 +179,21 @@ export class ContractFunctionRevertedError extends BaseError {
     data,
     functionName,
     message,
+    cause: error,
```
alcuadrado: I didn't want to rename `cause`, which is used below, to keep the PR minimal. Let me know if you prefer something else.

src/utils/abi/decodeErrorResult.ts
alcuadrado: If `decodeAbiParameters` throws here, we loose the `cause`. I left it like this to keep the PR minimal, and added a test that acts as a witness of this limitation. Open to take another approach.

PR #4396, opened 2026-03-12
fix: garbage collect listeners in observe

### LINKED ISSUE #4390: observe() leaks empty arrays in listenersCache after all listeners unsubscribe
### Check existing issues

### Viem Version

2.31.7

### Current Behavior

When all listeners for a given `observerId` call `unwatch()`, the `observe()` function in `src/utils/observe.ts` removes individual callbacks via `filter()` (line 13), but **never deletes the Map entry** from `listenersCache` or `cleanupCache`. This leaves behind stale entries with empty arrays `[]` that accumulate indefinitely.

In a long-running Node.js server that creates many viem clients over time (e.g., `createPublicClient`, `createClient`), `listenersCache` grows without bound, causing a memory leak.

We confirmed this by inspecting the cache at runtime: thousands of entries with empty `[]` values, never garbage-collected because the `Map` holds strong references.

### Expected Behavior

When the last listener for an `observerId` unsubscribes, the entry should be fully removed from both `listenersCache` and `cleanupCache`, not left as an empty array.

### Steps To Reproduce

```
import { createPublicClient, http } from 'viem';

// Access internal caches
const viemObserve = require(
  require('path').join(
    require('path').dirname(require.resolve('viem/package.json')),
    '_cjs/utils/observe.js'
  )
);

// Simulate a server creating clients on each request
for (let i = 0; i < 1000; i++) {
    chain: mainnet,
    transport: http('https://rpc.ankr.com/eth'),
  });
  // Client goes out of scope but listenersCache entry remains
}

console.log('listenersCache size:', viemObserve.listenersCache.size);
// Expected: 0 (or a small number for active observers)
// Actual: grows with each iteration, entries are empty arrays
```

_No response_

### Anything else?

The fix would be to delete the Map entries in `unsubscribe()` when the filtered listeners array is empty. Something like:

```
const unsubscribe = () => {
  const listeners = getListeners();
  const remaining = listeners.filter((cb) => cb.id !== callbackId);
  if (remaining.length === 0) {
    listenersCache.delete(observerId);
    cleanupCache.delete(observerId);
  } else {
    listenersCache.set(observerId, remaining);
  }
};
```

Our current workaround is an hourly `setInterval` that sweeps `listenersC`ache for empty arrays and deletes them, plus caching our viem clients to reduce the rate of accumulation.

Sigmabrogz:
Fixes #4390. Properly garbage collects `listenersCache` and `cleanupCache` entries when all listeners are unsubscribed.


Sigmabrogz:
I've added the changeset as requested.

Sigmabrogz:
Hi, I've added the changeset as requested. The PR is ready for review. Thanks!

PR #4402, opened 2026-03-16
fix(errors): preserve original provider cause in getContractError

### LINKED ISSUE #4387: `getContractError` can loose the error `cause` when a custom provider is used
### Check existing issues

### Viem Version

2.47.1

### Current Behavior

When interacting with a contract, the function `getContractError` can loose the error cause when using a custom provider. The error returned by it is correct in general, but it can loose track of the original error, making debugging harder.

### Expected Behavior

The error returned by `getContractError` should have the right cause.

### Steps To Reproduce

Here's a self-contained minimal repro with viem@2.47.1:

```ts
import { createWalletClient, getContract, custom, parseAbi } from "viem";
import { mainnet } from "viem/chains";

// A mock custom provider that throws this error for everything except eth_chainId
const executionErrorInCustomProviderWithCode3: any = new Error(
  `Execution error in custom provider`,
);
executionErrorInCustomProviderWithCode3.name = "CustomProviderExecutionError";
// ABI encoded "Error(string)" of "reverted"
executionErrorInCustomProviderWithCode3.data =
  "0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000087265766572746564000000000000000000000000000000000000000000000000";
executionErrorInCustomProviderWithCode3.code = 3;

const ethereumProvider = {
  async request({ method }: { method: string }) {
    console.log("Provider called with method:", method);
    if (method === "eth_chainId") {
      return "0x1";
    }

    throw executionErrorInCustomProviderWithCode3;
  },
};

const abi = parseAbi([
  "function transfer(address to, uint256 amount) returns (bool)",
]);

const sender = "0x1234567890123456789012345678901234567890";

const client = createWalletClient({
  chain: mainnet,
  transport: custom(ethereumProvider),
  account: sender,
});

const contract = getContract({
  address: "0x1111111111111111111111111111111111111111",
  abi,
  client,
});

try {
  await contract.write.transfer([
    "0x2222222222222222222222222222222222222222",
    0n,
  ]);
} catch (error: any) {
  console.log("error.name", error.name);
  console.log("error.cause?.name", error.cause?.name);

  // This should be CustomProviderExecutionError, but is `undefined`
  console.log("error.cause?.cause?.name", error.cause?.cause?.code);
  console.log();
  console.log();
  console.log(error);
}
```

_No response_

### Anything else?

Sending a PR to fix it in a few minutes

jasonandjay:
Fixes #4387

## Summary
When a custom provider throws an execution error (e.g. `{ code: 3, data, name: 'CustomProviderExecutionError' }`), `getContractError` currently wraps it into `ContractFunctionRevertedError` but can lose the original nested cause.

This PR preserves that cause chain:
- `getContractError` now passes the discovered low-level error as `cause` into `ContractFunctionRevertedError`.
- `ContractFunctionRevertedError` now accepts an optional `cause` parameter and keeps it unless replaced by ABI decode failures.
- Added a regression test asserting `error.cause.cause?.name === 'CustomProviderExecutionError'`.

## Notes
I couldn't fully run local repo checks in this environment because `viem` currently requires Node >= 24.5 and this runner is Node 22. CI should validate on the project's standard toolchain.

### DIFF
--- a/src/errors/contract.ts
+++ b/src/errors/contract.ts
@@ -176,16 +176,18 @@ export class ContractFunctionRevertedError extends BaseError {
 
   constructor({
     abi,
+    cause,
     data,
     functionName,
     message,
   }: {
     abi: Abi
+    cause?: Error | undefined
     data?: Hex | undefined
     functionName: string
     message?: string | undefined
   }) {
-    let cause: Error | undefined
+    let cause_ = cause
     let decodedData: DecodeErrorResultReturnType | undefined
     let metaMessages: string[] | undefined
     let reason: string | undefined
@@ -222,13 +224,13 @@ export class ContractFunctionRevertedError extends BaseError {
           ]
         }
       } catch (err) {
-        cause = err as Error
+        cause_ = err as Error
       }
     } else if (message) reason = message
 
     let signature: Hex | undefined
-    if (cause instanceof AbiErrorSignatureNotFoundError) {
-      signature = cause.signature
+    if (cause_ instanceof AbiErrorSignatureNotFoundError) {
+      signature = cause_.signature
       metaMessages = [
         `Unable to decode signature "${signature}" as it was not found on the provided ABI.`,
         'Make sure you are using the correct ABI and that the error exists on it.',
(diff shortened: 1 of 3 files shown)


jxom:
already a pr for this https://github.com/wevm/viem/pull/4388

Issue #3549, opened 2025-04-08
Excessive type inference prevents code reuse, and hurts ecosystem


KholdStare:
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/viem/issues) for the bug I encountered.

### Viem Version

2.26.0

### Current Behavior

To start off, the overall API "shape" is great and is an improvement over ethers.

However, using viem/wagmi beyond just directly calling it is basically impossible because of the excessive use of type inference. _I have lost hours and hours of productivity when trying to make any generic wrappers of wagmi/viem functionality_. Most of the Ethereum community has embraced viem and wagmi over ethers, but I think the ecosystem is now actively being stifled by the inaccessibility of reuse of any types provided by viem. Let me give a few examples.

Because the type safety needs to be maintained throughout _with type inference_, that means the crazy types have to be added _at every single layer_. This means users who want to use the viem/wagmi API beyond "dumb/direct" calls are forced to wrestle with pages and pages of arcane typescript errors in order to add it to _their layer_. This is like the most extreme version of a [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/), where every single lower-level detail is _always_ bubbled up to the top in the types.

Type inference of function names and arguments from the ABI is great. However, anything beyond that, such as Chain, Account, RpcSchema inference is just overkill and I guarantee you the vast majority of users do not care and will not benefit from this. Having looser types for this will drastically simplify your implementation, make the source way more readable (I try to read the source to understand concrete type definitions to no avail), and allow others to wrap viem/wagmi functionality in their own code bases. If there is something special that is "keyed off" of the custom Chain type, then isolate it to that special function - just have a runtime test for a flag that the chain is "special". Then you don't have to polute 99% of the API for the marginal benefit of 1% of the API.

Even for argument type inference you should be able to isolate it to a single function, and not have it infect everything. Have a helper function that does the inference, and return a strong type that declares the input has been validated. That validated type can then be used in the rest of the API.

```ts
// A more simple/straightfoward API type that is publically used/visible
export type SimulateContractParameters = {
    account: Account;
    address: Address;
    abi: Abi;
    functionName: string; // yes, string!
    args: unknown[] // or more string union of possible primitives
};

// Some "newtype" or strong type that distinguishes unverified from validated params
// Type should be hard to create outside of viem library
export type ValidSimulateContractParameters = Tag<SimulateContractParameters, "Valid">;

[...]

PR #4455, opened 2026-04-03
fix: accept readonly arrays in LogTopic type

nikicat:
## Summary

`LogTopic` is changed from `Hex | Hex[] | null` to `Hex | readonly Hex[] | null`. Since `readonly T[]` is a supertype of `T[]`, this accepts both mutable and readonly arrays.

`LogTopicType` conditional is updated correspondingly: it now matches `readonly Hex[]` (which includes `Hex[]`) and produces `readonly primitiveType[]`.

## Motivation

TypeScript's `readonly T[]` is not assignable to `T[]`. Users who define topic arrays with `as const` or receive them from functions returning readonly arrays get type errors when passing them to `getLogs`, `createEventFilter`, `watchEvent`, etc.

Log topic filters are read-only by nature — they're passed to JSON-RPC calls and never mutated — so accepting readonly arrays is both correct and expected.

This is a non-breaking change: all existing mutable-array code continues to work since `T[]` is assignable to `readonly T[]`.

## Related discussions

- #2671 — readonly values friction with viem APIs
- #1480 — readonly tuple assignability errors with hex values
- #163 — log `topics` type narrowing
- #3549 — excessive type inference hurting ecosystem

## Test plan

- [x] All 51 type test files pass (343 tests, no type errors)

### DIFF
--- a/.changeset/readonly-log-topic.md
+++ b/.changeset/readonly-log-topic.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `LogTopic` type to accept readonly arrays.
--- a/src/types/contract.test-d.ts
+++ b/src/types/contract.test-d.ts
@@ -215,10 +215,10 @@ test('GetEventArgs', () => {
     'Transfer'
   >
   expectTypeOf<Result['from']>().toEqualTypeOf<
-    `0x${string}` | `0x${string}`[] | null | undefined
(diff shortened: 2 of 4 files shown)


nikicat:
Friendly ping — this has been open for about two weeks without a review. No rush, just want to make sure it hasn't slipped through. Happy to rebase or adjust anything if needed. Thanks!

PR #4469, opened 2026-04-07
perf: avoid O(n^2) memory allocation in bytesToHex

haydenshively:
The for loop in `bytesToHex` meant O(n^2) allocations for string concatenation, which can get heavy when:
- you're executing large multicalls
- you've enabled `dedupe: true`, which stringifies the request and converts it to hex for the dedupe key


jxom:
Thanks for raising! We benchmarked both approaches and string concatenation is actually ~2-3x faster than `Array.join` in V8. Are you seeing otherwise?

```
size=32    | concat: 5.5ms   | join: 14.8ms   | concat 167% faster
size=256   | concat: 37.8ms  | join: 113.8ms  | concat 201% faster
size=1024  | concat: 152.6ms | join: 443.8ms  | concat 191% faster
size=8192  | concat: 1.2s    | join: 3.5s     | concat 185% faster
size=65536 | concat: 860.9ms | join: 3.0s     | concat 248% faster
```

PR #4471, opened 2026-04-07
fix: proper offset to increase entropy in uid generation

haydenshively:
Simple fix to increase entropy (avoid sharing characters) when generating UIDs -- I think this is what was originally intended.


jxom:
Thanks for raising! The UIDs are still unique (each offset by 1 char), so there's no functional issue for our use case. Going to close this one.

PR #4486, opened 2026-04-10
feat: EIP-8141 - Frames Transactions

ch4r10t33r:
Draft PR: adds Frame Transaction support ([EIP-8141](https://eips.ethereum.org/EIPS/eip-8141)).

Aligned with the EIP-8141 draft as of ethereum/EIPs@b75cbe61 (2026-09-01).

### What's included

- **Types**: `Frame` (`mode`, `flags`, `target`, `limits: { execution, state }`, `value`, `data`), `FrameSignature` (`scheme`, `signer`, `msg`, `signature`), `FrameReceipt` (`status` incl. `'skipped'`, `gasUsed`, `stateGasUsed`, `logs`), `TransactionEIP8141` / `TransactionRequestEIP8141` / `TransactionSerializableEIP8141`, RPC counterparts.
- **Serialization / parsing** of the type `0x06` payload `[chain_id, nonce, sender, frames, signatures, fees, blob_versioned_hashes]`.
- **Signing**: `signTransaction` signs the canonical signature hash (empty-`msg` signature bytes elided) and stores `v || r || s` in the sender's `SECP256K1` slot of `signatures`; `recoverTransactionAddress` recovers from that slot.
- **Static validation** (`assertTransactionEIP8141`) mirroring the spec's Constraints section, including the EIP-7825 transaction gas cap.
- **Receipts**: `payer` and per-frame receipts with the three-valued frame status (`0x2` = skipped) and `stateGasUsed`.
- **Formatting** of `frames` / `signatures` in `formatTransaction`.
- Examples for the spec's simple, atomic-batch and sponsored flows.

### Notes

- Frame transactions are excluded from `sendTransaction` / `prepareTransactionRequest` (they need `eth_sendRawTransaction`).
- The JSON-RPC shape for `frames` / `signatures` is not yet standardised in execution-apis; this PR follows the spec's field names.


ayushsingh82:
Noticed while building against a live EIP-8141 chain: the frame-receipt formatter in `src/utils/formatters/transactionReceipt.ts` collapses a three-valued status into two.

```ts
status: fr.status === '0x1' ? 'success' : 'reverted',
```

ethrex's frame status is `0` failure, `1` success, `2` **skipped** (`crates/networking/rpc/types/receipt.rs:36` — *"0 = failure, 1 = success, 2 = skipped"*). A skipped frame is one rolled back by a failed atomic batch — it never executed and its gas was refunded — so mapping `0x2` to `'reverted'` reports a refund as a revert, and the mapping also can't distinguish an outright failure (`0x0`) from a skip.

The same block also drops `stateGasUsed`: the RPC frame receipt carries `stateGasUsed` alongside `gasUsed` (`receipt.rs:47`, EIP-8037 state gas), but the formatter only reads `fr.gasUsed`.

Verified against the live node (`ethrex/v23.0.0-hegota-testnet-31b5322665`). Not urgent given the PR's draft status — just flagging while it's fresh.

ch4r10t33r:
Thanks @ayushsingh82, fixed in 1fc89e2: frame status now maps `0x2` to `skipped` (and `0x0` to `reverted`), and `stateGasUsed` is read alongside `gasUsed`. The branch is also brought up to the current EIP-8141 draft (nested `limits`/`fees`, outer `signatures` list) and merged with upstream main.

PR #4545, opened 2026-04-23
fix(encodeEventTopics): handle anonymous events correctly

Jah-yee:
Fixes #4461

Anonymous events should not have the signature prepended as topic0. This fix adds an early return for anonymous events, returning only the encoded topic values.

## Changes

In `src/utils/abi/encodeEventTopics.ts`, added a check for `abiItem.anonymous` before returning. For anonymous events, the function now returns only the encoded indexed arguments (empty array if no args), rather than prepending the event signature as topic0.

## Test cases (per issue #4461)

`anonymous event: no args` → `[]`
`anonymous event: topic0` → `[0x...abcd, null, null, null]`
`anonymous event: all topics` → `[topic0, topic1, topic2, topic3]`

---

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly, RoomWithOutRoof

### DIFF
--- a/.changeset/rare-bats-wait.md
+++ b/.changeset/rare-bats-wait.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+fix(encodeEventTopics): handle anonymous events correctly
--- a/src/utils/abi/encodeEventTopics.ts
+++ b/src/utils/abi/encodeEventTopics.ts
@@ -121,6 +121,11 @@ export function encodeEventTopics<
         }) ?? []
     }
   }
+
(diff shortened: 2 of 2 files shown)


Jah-yee:
Hi, thanks for the review! I've added a changeset as requested. The PR addresses the case where `encodeEventTopics()` receives anonymous events — it now returns `topics` directly without prepending the signature, which is the correct behavior per the ABI event spec.

Please let me know if anything else needs adjustment. 🙂

Warmly,
RoomWithOutRoof

PR #4551, opened 2026-04-25
Add CCIP-Read Tunneling

adraffy:
- added `utils/ccipTunnel.ts` and tests
     * exposes `ccipReadTunnel()`
 
---

Usage:
```ts
// <meta http-equiv="Content-Security-Policy" content="connect-src 'self' <batchGateway>;">
createClient({
   ccipRead: ccipReadTunnel({ batchGateways: [<batchGateway>] })
});
```
&rarr; routes <ins>ALL</ins> CCIP-Read `OffchainLookup` requests through `<batchGateway>`.

---

Note: CCIP-Read is not ENS-specific.

### DIFF
--- a/src/utils/ccipTunnel.ts
+++ b/src/utils/ccipTunnel.ts
@@ -1,4 +1,5 @@
 import { batchGatewayAbi } from '../constants/abis.js'
+import { solidityError } from '../constants/solidity.js'
 import { HttpRequestError } from '../errors/request.js'
 import { decodeErrorResult } from './abi/decodeErrorResult.js'
 import { decodeFunctionResult } from './abi/decodeFunctionResult.js'
@@ -50,7 +51,18 @@ export function createCcipReadTunnel({
               status,
               url: urls.join(' | '),
             })
-          } catch {}
+          } catch {
+            let message: string | undefined
+            try {
+              ;({
+                args: [message],
+              } = decodeErrorResult({
+                abi: [solidityError],
+                data: responses[0],
+              }))
+            } catch {}
+            throw new Error(message ?? 'An unknown error occurred.')
+          }
         }
         return responses[0]
       }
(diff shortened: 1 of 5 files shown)


src/utils/ccipTunnel.ts
```diff
@@ -0,0 +1,67 @@
...
+import { encodeFunctionData } from './abi/encodeFunctionData.js'
+import { ccipRequest as ccipRequest_ } from './ccip.js'
+import { localBatchGatewayUrl } from './ens/localBatchGatewayRequest.js'
+
+export type CcipReadTunnelParameters = {
+  batchGateways: string[]
+  ccipRequest?: typeof ccipRequest_
+}
+
+export function createCcipReadTunnel({
```
jxom:
```suggestion
export function ccipReadTunnel({
```

jxom (commented):
one small nit

adraffy:
Oh cool, does the idea make sense?  It could technically be an external library since the code is independent but I think it's actually useful for frontend applications (and will get use as there's an increased desire to lock frontends down) and the new code is only a small addition.

I have a [matching solution](https://github.com/ensdomains/ethers-patch/pull/3) for ethers for solving the same problem (CSP).

---

This will be standardized as part of [ENSIP-21](https://docs.ens.domains/ensip/21/) but I'm thinking about writing an ERC, since it's actually a security improvement on [ERC-3668](https://eips.ethereum.org/EIPS/eip-3668) (although ENSIP-21 itself was too, and I didn't write one.)

src/utils/ccipTunnel.ts
```diff
@@ -0,0 +1,67 @@
...
+          try {
+            const res = decodeErrorResult({
+              abi: [...batchGatewayAbi, solidityError],
+              data: responses[0],
+            })
+            if (res.errorName === 'HttpError') {
+              error = new HttpRequestError({
+                body: { message: res.args[1] },
+                status: res.args[0],
+                url: urls.join(' | '),
```
adraffy, line 52: I was unsure what to put here.

Basically, the batch gateway failed to fetch all of the URLs, so it's not clear which one gets the blame.

adraffy:
What can be done here?  I'm getting spammed with Github emails

Edit: thanks!

PR #4572, opened 2026-05-02
fix: handle anonymous events in encodeEventTopics

aayushbaluni:
## Root cause

`encodeEventTopics` always prefixed the topic list with `toEventSelector(...)`. That matches non-anonymous events, where topic0 is the keccak256 of the signature string. For **anonymous** events, the execution-layer rule is different: logs do **not** carry that signature topic—indexed arguments occupy topics starting at topic0.

Because of this mismatch, filters built via `encodeEventTopics` (and callers such as `watchContractEvent` / `createContractEventFilter`) sent an incorrect topic layout for anonymous events.

## Fix

If the ABI event item has `anonymous: true`, return only the encoded indexed-parameter topics. Otherwise, keep the existing behavior `[signature, ...indexedTopics]`.

The public return type is widened accordingly (`signature-led tuple | indexed-only topics array`).

## Testing

- Added regression coverage for https://github.com/wevm/viem/issues/4461: anonymous vs non-anonymous `Deposit(address,uint256)` with the same indexed argument shows anonymous topics without the signature hash and non-anonymous topics with signature first.
- Ran: `SKIP_GLOBAL_SETUP=1 pnpm exec vitest run src/utils/abi/encodeEventTopics.test.ts -c ./test/vitest.config.ts`

Fixes https://github.com/wevm/viem/issues/4461

### DIFF
--- a/.changeset/fix-anonymous-event-topics.md
+++ b/.changeset/fix-anonymous-event-topics.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+fix: handle anonymous events in encodeEventTopics
--- a/src/utils/abi/encodeEventTopics.test.ts
+++ b/src/utils/abi/encodeEventTopics.test.ts
@@ -532,3 +532,46 @@ test('https://github.com/wevm/viem/issues/3278', () => {
     ]
   `)
 })
+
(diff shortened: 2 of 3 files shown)


aayushbaluni:
Hi @jxom, I noticed this was closed without a comment. Is there an issue with the approach, or was this superseded by another fix? Happy to adjust if needed.

PR #4596, opened 2026-05-11
fix(tempo): verify allowAccessKey for secp256k1 keychain

cmende:
## What is this PR solving?
                                                                                                                                                                                                 
  This PR fixes an incomplete `allowAccessKey` verification path in `tempo` `verifyHash`.                                                                                                        
                                                                                                                                                                                                 
  In the keychain branch, the current logic derives the access key address only from `envelope.inner.publicKey`.                                                                                 
  For `secp256k1` access keys, the inner envelope may not include `publicKey`, which can cause thrown errors or false-negative verification for otherwise valid signatures.                      
                                                                                                                                                                                                 
  This change adds a fallback path:                                                                                                                                                              
  - If `envelope.inner.publicKey` exists, keep current behavior.                                                                                                                                 
  - If it is missing and `envelope.inner.type === 'secp256k1'`, recover the access key address from the inner signature + payload.                                                               
  - If address derivation fails, return `false` instead of throwing.                                                                                                                             
                                                                                                                                                                                                 
  This preserves the existing metadata checks (authorized, not revoked, not expired) and only broadens support for valid `secp256k1` keychain envelopes.                                         
                                                                                                                                                                                                 
  ## Alternatives explored                                                                                                                                                                       
                                                                                                                                                                                                 
  - Require `publicKey` for all key types in keychain envelopes.                                                                                                                                 
    - Rejected: not compatible with existing `secp256k1` envelope shape.                                                                                                                         
  - Skip local envelope verification and always fall back to generic `verifyHash`.                                                                                                               
    - Rejected: would bypass the intended access-key metadata validation path in `allowAccessKey` mode.                                                                                          
  - Current approach: minimal fallback only for missing-`publicKey` `secp256k1` inner signatures.                                                                                                
    - Chosen for minimal surface area and behavior preservation.
                                                                                                                                                                                                 
  ## Reviewer attention                                                                                                                                                                          
                                                                                                                                                                                                 
  Please focus on:                                                                                                                                                                               
  - The new fallback in `src/tempo/chainConfig.ts` for keychain + `allowAccessKey`.                                                                                                              
  - `innerPayload` usage consistency for v1/v2 keychain envelopes.                                                                                                                               
  - Error handling behavior (`false` return instead of throw) in malformed/missing-key scenarios.                                                                                                
                                                                                                                                                                                                 
  ## Tests
                                                                                                                                                                                                 
  Added:                                                                                                                                                                                         
  - `accessKey: secp256k1 valid signature` in `src/tempo/chainConfig.test.ts`.                                                                                                                   

  The new test reproduces the previously unhandled `secp256k1` keychain path and validates the fix.                                                                                              
                                                                                                                                                                                                 
  ## Docs                                                                                                                                                                                        
                                                                                                                                                                                                 
  No public API changes. No docs update required.

### DIFF
--- a/.changeset/tempo-verify-secp256k1-keychain.md
+++ b/.changeset/tempo-verify-secp256k1-keychain.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
-Fixed `verifyHash` for Tempo `allowAccessKey` mode to support `secp256k1` keychain access keys (whose inner envelope does not carry a `publicKey`).
+Fixed `verifyHash` for Tempo `allowAccessKey` mode with `secp256k1` keychain access keys, whose inner envelope did not carry a `publicKey`.
--- a/.changeset/tempo-verify-secp256k1-keychain.md
+++ b/.changeset/tempo-verify-secp256k1-keychain.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `verifyHash` for Tempo `allowAccessKey` mode to support `secp256k1` keychain access keys (whose inner envelope does not carry a `publicKey`).
--- a/src/tempo/chainConfig.test.ts
+++ b/src/tempo/chainConfig.test.ts
@@ -388,6 +388,30 @@ describe('verifyHash', () => {
     ).toBe(true)
   })
 
+  test('accessKey: secp256k1 valid signature', async () => {
+    const rootAccount = accounts.at(0)!
+    const accessKey = Account.fromSecp256k1(generatePrivateKey(), {
+      access: rootAccount,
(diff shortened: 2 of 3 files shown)


imaksp:
Hi @jxom , FYI : this commit  has removed following namespace imports 
```ts
import * as from ox/PublicKey
import * as Address from 'ox/Address'
``` 
it helped in better tree shaking with ESbuild and Rollup, so now  P-256 and secp256k1  are dropped form our initial bundle, it saved ~400 kB for us, it was getting included even if we don't import tempo chain. so prefer named import where possible,  Claude helped  in finding the root cause.

PR #4597, opened 2026-05-11
feat(op-stack): Add support for super root games

wwared:
This PR makes a set of changes to support the upcoming interop optimism upgrade, which will migrate chains to use [super roots](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/superroot.md) instead of plain output roots. Note that even chains that don't make use of interop will migrate to the super root on-chain dispute games that use the new structure supported by this PR.

Currently, the `l2SequenceNumber` corresponds one-to-one with the old `l2BlockNumber`, but after the super root migration, the `l2SequenceNumber` will refer instead to the block's timestamp instead of its number.

This PR modifies the `op-stack` library with minimal changes to support the new game types. It adds:

1) Helper function `getL2BlockNumberAtTimestamp` to map a timestamp back to a L2 block number (needed to fetch the corresponding block's info via RPC)
2) Helper function `isSuperGameType` (uses simple hard coded constants)
3) Optional `l2Timestamp` parameter to `GetTimeToProveParameters` (enables the new code path without breaking existing callers)
4) An example script illustrating a withdrawal using op-stack
5) Simple tests for the new code

Then, based on the game type being used in the proof, the function will transparently use either the old L2 block number as-is or the new timestamp instead. To minimize breaking changes, I didn't rename any of the existing `l2BlockNumber` variables. In particular, `GetGameReturnType` now may return a `l2BlockNumber` that is a timestamp instead of a block number, indicated by the `usesSuperRoots` flag.

Note that the proofs are still done as-is and still reference the chain's output root (instead of the super root itself). This is intentional, and so withdrawals work mostly as-is, the only change needed is to replace the `l2BlockNumber` (`l2SequenceNumber`) value by a `l2Timestamp` instead.

It's possible the new example or tests don't follow existing repo convention; let me know if you'd need me to make any changes to them.

I've tested these changes manually against a local devnet, and it works fine with both the new super games and the old legacy game types as well.

### DIFF
--- a/.changeset/super-root-games.md
+++ b/.changeset/super-root-games.md
@@ -1,0 +1,5 @@
+---
+"viem": minor
+---
+
+`viem/op-stack`: Added OP Stack super-root dispute game support for withdrawal prove flows.
--- a/examples/op-stack_super-withdrawal/README.md
+++ b/examples/op-stack_super-withdrawal/README.md
@@ -1,0 +1,67 @@
+# OP Stack Super Withdrawal Example
+
+Manual end-to-end withdrawal verification for OP Stack devnets. The same script
+can run against:
+
+- A pre-super-roots devnet where the respected Portal game type uses legacy L2
+  block-number anchors.
+- A post-super-roots devnet where the respected Portal game type uses L2
+  timestamp anchors.
+
+This example is intentionally not wired into CI. It requires external devnets,
+funded accounts, and devnet finalization timings short enough for manual runs.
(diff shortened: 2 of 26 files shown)


pauldowman:
@jxom this change will be necessary for withdrawals to continue working on OP Stack chains with the upcoming interop upgrade. Could we get a review, or just a first-pass at one in case there's anything you'd like to see done differently?

PR #4607, opened 2026-05-13
feat(tempo): add sender scoped transaction hashes

0xrusowsky:
## Motivation

[Tempo's TIP-1034](https://github.com/tempoxyz/tempo/blob/main/tips/tip-1034.md) channel opens bind channel identity to the tx via an `expiringNonceHash`:

```txt
keccak256(encodeForSigning || sender)
```

this gives the precompile its replay/same-transaction semantics, but it means viem clients need a way to compute the exact sender-scoped hash before signing the initial voucher.

this depends on https://github.com/wevm/ox/pull/246, which exposes the Tempo transaction signing preimage from ox.

## Solution

add Tempo helpers for deriving the sender-scoped transaction hash:

```ts
Transaction.encodeForSigning(transaction): Hex
Transaction.getSenderScopedHash(transaction, { sender }): Hex
```

`getSenderScopedHash` hashes the Tempo signing preimage together with the resolved sender address. `getExpiringNonceHash` and `getChannelOpenContextHash` are exported as aliases for the same TIP-1034-derived value.

this also refactors Tempo serialization to share the same internal envelope normalization used by `encodeForSigning`, preserving the existing fee-payer behavior where `feePayer: true` leaves fee token selection to the fee payer.

## Tests

- `keccak256(encodeForSigning(tx))` matches the ox Tempo sender signing payload.
- `encodeForSigning` is stable across sender/fee-payer signatures and fee-payer-selected fee token changes when the sender does not commit to the fee token.
- `getSenderScopedHash` / `getExpiringNonceHash` / `getChannelOpenContextHash` match `keccak256(encodeForSigning(tx) || sender)`.

## Notes

opened as draft because this branch uses a local `ox@0.14.20` patch containing https://github.com/wevm/ox/pull/246, so CI is expected to fail until the ox change lands and viem can depend on the published version.

### DIFF
--- a/.changeset/tempo-sender-scoped-hash.md
+++ b/.changeset/tempo-sender-scoped-hash.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Added `Transaction.encodeForSigning` and `Transaction.getSenderScopedHash` helpers for Tempo TIP-1034 sender-scoped transaction hashes.
--- a/package.json
+++ b/package.json
@@ -311,6 +311,9 @@
       "bun",
       "protobufjs",
       "simple-git-hooks"
-    ]
+    ],
+    "patchedDependencies": {
+      "ox@0.14.20": "patches/ox@0.14.20.patch"
+    }
   }
 }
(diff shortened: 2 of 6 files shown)


jxom:
This is scoped into v3.

PR #4660, opened 2026-05-24
fix: handle prefilled Tempo fee payer signatures

struong:
## Summary

Fixes single-roundtrip Tempo sponsorship when `eth_fillTransaction` returns a prefilled fee payer signature. Keeps fill-response call expansion opt-in via `_capabilities.autoSwap` instead of automatically trusting `transaction.calls` returned by the relay.

## Motivation

`serializeTempo` previously emitted a partial sponsorship handoff even when both sender and fee payer signatures were already present, so `withRelay` made an unnecessary second `eth_signRawTransaction` round trip. The sponsorship sign payload also needed tighter `feeToken` handling so the sender does not commit to the fee token before final broadcast. Relay-provided `transaction.calls` should not be adopted automatically because a malicious relay could alter the call set; callers can explicitly include `_capabilities.autoSwap.calls` when they want that behavior.

## Changes

- Preserved caller-provided `calls` in `prepareTransactionRequest` and exposed `autoSwap.calls` through `_capabilities` for explicit caller opt-in.
- Emitted a full `0x76` broadcast envelope when both signatures are present, allowing `withRelay` to broadcast without a second fee payer signing request.
- Normalized prefilled fee payer signature handling and stripped `feeToken` from sender sign payloads and partial sponsorship handoffs while keeping it on the final broadcast envelope.
- Added a patch changeset for the Tempo sponsorship serialization fix.

## Testing

- Added regression coverage that `prepareTransactionRequest` ignores relay transaction `calls` while exposing `autoSwap.calls`.
- Added Tempo serialization coverage for full-envelope broadcast and `feeToken` stripping cases.
- `SKIP_GLOBAL_SETUP=true pnpm test --project core src/actions/wallet/prepareTransactionRequest.test.ts -t "ignores relay transaction"`
- `pnpm test --project tempo src/tempo/Transaction.test.ts src/tempo/Transport.test.ts`
- `pnpm biome check src/actions/wallet/prepareTransactionRequest.ts src/actions/wallet/prepareTransactionRequest.test.ts src/tempo/Transaction.ts src/tempo/Transaction.test.ts`

### DIFF
--- a/src/actions/wallet/prepareTransactionRequest.ts
+++ b/src/actions/wallet/prepareTransactionRequest.ts
@@ -428,18 +428,6 @@ export async function prepareTransactionRequest<
             rest.feePayerSignature !== null
               ? { feePayerSignature: rest.feePayerSignature }
               : {}),
-            // Adopt expanded calls from the fill response so the
-            // broadcast envelope matches what the fee payer signed over.
-            ...('calls' in rest &&
-            Array.isArray(rest.calls) &&
-            rest.calls.length > 0
-              ? {
-                  calls: rest.calls,
-                  to: undefined,
-                  data: undefined,
-                  value: undefined,
-                }
-              : {}),
             ...('feeToken' in rest &&
             typeof rest.feeToken !== 'undefined' &&
             rest.feeToken !== null &&
--- a/src/actions/wallet/prepareTransactionRequest.ts
+++ b/src/actions/wallet/prepareTransactionRequest.ts
@@ -428,6 +428,18 @@ export async function prepareTransactionRequest<
             rest.feePayerSignature !== null
               ? { feePayerSignature: rest.feePayerSignature }
               : {}),
+            // Adopt the relay's expanded calls (e.g. autoSwap) so the
+            // broadcast envelope matches what the fee payer signed over.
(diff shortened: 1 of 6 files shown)


src/actions/wallet/prepareTransactionRequest.ts
```diff
@@ -425,6 +428,18 @@ export async function prepareTransactionRequest<
...
+            ...('calls' in rest &&
+            Array.isArray(rest.calls) &&
+            rest.calls.length > 0
+              ? {
+                  calls: rest.calls,
+                  to: undefined,
+                  data: undefined,
+                  value: undefined,
+                }
+              : {}),
```
jxom: I don't think we should do this, as an `eth_fillTransaction` to a malicious relay may return "different" `calls`.
The `prepareTransactionRequest` action returns an `autoSwap` capability, so the consumer should use that to fill the calls for sign tx instead.

```typescript
const request = client.prepareTransactionRequest(...)
const autoSwap = request._capabilities.autoSwap
const signature = await client.signTransaction(client, {
  ...request,
  calls: [...autoSwap.calls, ...request.calls]
})
```

PR #4666, opened 2026-05-26
test: fix false-positive in eip712 signing test

Olexandr88:
The signed with customSignature test was using baseEip712 which already contained a hardcoded customSignature, causing the serializer to ignore the provided privateKey entirely and produce output identical to the default test. Fixed by setting customSignature: undefined and aligning from with accounts[0].address to match the signing key, so the test now exercises the actual signing path.


jxom:
Thanks, but this doesn't do what the description says. For `type: 'eip712'` transactions, `serializeTransaction` routes to `serializeTransactionEIP712`, which ignores the ECDSA signature argument and serializes `customSignature ?? '0x'`. With `customSignature: undefined` the private key is still unused. Verified locally: the new snapshot still passes with `accounts[1].privateKey`.

The new snapshot is also an unsigned transaction (empty `customSignature` slot), so the test no longer matches its name, and it drops the behavior the original pinned: signing an EIP-712 transaction that already carries a `customSignature` must preserve it rather than clobber it. The actual signing path is already covered by `zksync/actions/signTransaction.test.ts`.

PR #4714, opened 2026-06-04
chore: fix ladyChain formatting

jxom:
Fixes the formatting of `src/chains/definitions/ladyChain.ts`, which was merged over-indented and breaks `pnpm biome check` (the CI Check job runs biome across the whole repo, so this blocks all PRs branched off `main`).

Autofixed via `biome check --write`.


jxom:
Redundant — the formatting fix was already included in #4713, which is now merged to main.

PR #4723, opened 2026-06-10
Update access key docs

jenpaff:
Updates access key hardfork docs.


tmm (approved):
Docs-only access key hardfork update. I verified the touched MDX has no em dashes and guide headings remain Title Case. Site build is blocked by existing Tempo/Ox type mismatches before docs build runs.

PR #4756, opened 2026-06-23
Fix: ENS `getEnsName()` should enforce normalization

adraffy:
[ENSIP-23](https://docs.ens.domains/ensip/23) states:
> If the primary name is [unnormalized](https://docs.ens.domains/ensip/15), eg. `normalize("Nick.eth") != "Nick.eth"`, then name and resolver are invalid.

---

* changed `getEnsName()` to return `null` if not normalized
* added test with unnormalized name
* added test with offchain reverse resolver

---

- [ ] Replace ENS names with official test cases

### DIFF
--- a/src/actions/ens/getEnsName.ts
+++ b/src/actions/ens/getEnsName.ts
@@ -19,6 +19,7 @@ import {
   type ReadContractParameters,
   readContract,
 } from '../public/readContract.js'
+import { normalize } from '../../ens/index.js'
 
 export type GetEnsNameParameters = Prettify<
   Pick<ReadContractParameters, 'blockNumber' | 'blockTag'> & {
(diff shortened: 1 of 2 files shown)


src/actions/ens/getEnsName.ts
```diff
@@ -19,6 +19,7 @@ import {
   type ReadContractParameters,
   readContract,
 } from '../public/readContract.js'
+import { normalize } from '../../ens/index.js'
```
jxom, line 22: what would it take for `ens-normalize` to be much lighter in terms of bundle size? :D

adraffy: Since this is an async context, switch to dynamic import and guard it with a regex that covers common cases?

Now that LLMs are pretty good, I could revisit the normalization compression logic.

adraffy: Or, can you tell me what minimum JS engine you're supporting?  There is an approach that leverages regex `\p`.

The actual library logic is extremely small — nearly all of the size is the compressed Unicode data.

---

I've been contemplating making a `ens-normalize.ts` variant since the current library is basically the research version and all of the other language ports are clean and typed.

adraffy: Could I get some input on this? @jxom

adraffy: @jxom

PR #4768, opened 2026-06-29
fix(account-abstraction): prepare fees when estimating user operation gas

### LINKED ISSUE #4726: bug(account-abstraction): estimateUserOperationGas strips gas prices, causing validation failure on strict bundlers (EntryPoint 0.7)
### Check existing issues

### Viem Version

2.48.4

### Current Behavior

When using smartAccountClient with EntryPoint 0.7, the internal estimateUserOperationGas action explicitly ignores/strips maxFeePerGas and maxPriorityFeePerGas before calling the JSON-RPC method eth_estimateUserOperationGas. Strict bundlers (like Pimlico v2 endpoints) reject this payload with validation errors because they expect these string/hex properties to be defined in the user operation object for EP 0.7.

Root Cause Analysis
Looking at viem/account-abstraction/actions/bundler/estimateUserOperationGas.ts:
`const request = account
  ? await getAction(client, prepareUserOperation, 'prepareUserOperation')({
      ...parameters,
      parameters: [
        'authorization',
        'factory',
        'nonce',
        'paymaster',
        'signature',
      ],
    })
  : parameters;`

The parameters array explicitly passes only specific keys to prepareUserOperation. Because gas prices (maxFeePerGas, maxPriorityFeePerGas) are omitted here, the subsequent formatUserOperationRequest(request) outputs an operation object completely stripped of gas price fields.

While some bundlers fall back gracefully, strict RPC nodes enforce the presence of these fields in EntryPoint 0.7 schema validation, causing transaction flows to break entirely.

### Expected Behavior

`estimateUserOperationGas` should retain `maxFeePerGas` and `maxPriorityFeePerGas` if they are already resolved or provided by the `gasPrice` middleware/arguments, instead of filtering them out before hitting `client.request`.

### Steps To Reproduce

1. Set up a smartAccountClient using custom middleware hooks.
2. In the gasPrice middleware, return calculated maxFeePerGas and maxPriorityFeePerGas.
3. Trigger an action that invokes estimateUserOperationGas.
4. The bundler returns: Invalid input: expected string, received undefined at "params[0].userOp.maxFeePerGas".

_No response_

### Anything else?

Environment:

Viem Version: 2.48.4 (and newer v2 releases)
Bundler: Pimlico (v2 RPC endpoint)
EntryPoint: 0.7

nikhilkumar1612:
## Description
Fixes #4726.

`estimateUserOperationGas` was passing `maxFeePerGas` and `maxPriorityFeePerGas` in the `prepareUserOperation` `parameters` list, but those are User Operation field names, not valid preparation keys.

This updates the call to use the existing `fees` preparation key instead, so `prepareUserOperation` runs the `client.userOperation.estimateFeesPerGas` hook before sending `eth_estimateUserOperationGas`.

This makes the outgoing gas estimation request include `maxFeePerGas` and `maxPriorityFeePerGas` for strict bundlers.

Included a patch changeset.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] I have followed the steps in the CONTRIBUTING.md document.
- [x] My code follows the code style of this project.

### DIFF
--- a/.changeset/fix-estimate-user-operation-gas-fees.md
+++ b/.changeset/fix-estimate-user-operation-gas-fees.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `estimateUserOperationGas` omitting `fees` during User Operation preparation, which caused strict bundlers to reject gas estimation requests.
--- a/src/account-abstraction/actions/bundler/estimateUserOperationGas.ts
+++ b/src/account-abstraction/actions/bundler/estimateUserOperationGas.ts
@@ -177,11 +177,10 @@ export async function estimateUserOperationGas<
         parameters: [
           'authorization',
           'factory',
+          'fees',
           'nonce',
           'paymaster',
           'signature',
-          'maxFeePerGas',
-          'maxPriorityFeePerGas',
         ],
       } as unknown as PrepareUserOperationParameters)
     : parameters


jxom (approved):
Verified locally against a forked-anvil harness: without fees, the outgoing `eth_estimateUserOperationGas` payload now includes `maxFeePerGas`/`maxPriorityFeePerGas` (previously absent since the invalid preparation keys made the fees step a no-op); `userOperation.estimateFeesPerGas` hook values reach the wire; explicit fee overrides pass through unchanged; no double fee-fetch in the `sendUserOperation` path (inner prepare short-circuits on resolved fees).

Issue #921, opened 2023-07-21
bug: encodeFunctionData fails on empty `bytes[]` arg


timbrinded:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

1.3.1

### Current Behavior

I am trying to construct a contract write interaction using `encodeFunctionData()`, however it it failing due to passing an empty array to a `bytes[]` typed arg.

In my specific case I am providing the args:
```
[0, []]
```
to this function parameter `feeAsset`:
```
Multilocation memory feeAsset,
```
whose struct is defined as:
```
    struct Multilocation {
        uint8 parents;
        bytes[] interior;
    }
```

giving error:
```
TypeError: hex_.replace is not a function
 ❯ padHex node_modules/.pnpm/viem@1.3.12_typescript@5.1.6/node_modules/viem/src/utils/data/pad.ts:23:20
 ❯ encodeBytes node_modules/.pnpm/viem@1.3.12_typescript@5.1.6/node_modules/viem/src/utils/abi/encodeAbiParameters.ts:203:9
 ❯ prepareParam node_modules/.pnpm/viem@1.3.12_typescript@5.1.6/node_modules/viem/src/utils/abi/encodeAbiParameters.ts:101:12
 ❯ prepareParams node_modules/.pnpm/viem@1.3.12_typescript@5.1.6/node_modules/viem/src/utils/abi/encodeAbiParameters.ts:68:25
 ❯ encodeAbiParameters node_modules/.pnpm/viem@1.3.12_typescript@5.1.6/node_modules/viem/src/utils/abi/encodeAbiParameters.ts:43:26
 ❯ Module.encodeFunctionData node_modules/.pnpm/viem@1.3.12_typescript@5.1.6/node_modules/viem/src/utils/abi/encodeFunctionData.ts:58:9
```

### Expected Behavior

Expected to be able to handle this, as have had to rely on ethers.js to do the function encoding via an interface instead. Should be able to handle empty array, or array containing "0x"

### Steps To Reproduce

It can be easiest reproduced with this:
```
console.log(encodePacked(["uint8", "bytes[]"], [0, []]))
```

jxom:
Fixed for `encodePacked` on https://github.com/wagmi-dev/viem/commit/e3ff8509879d175f89634f4e2fcadd65ec4b7216.

[...]

PR #4777, opened 2026-06-30
fix(abi): encode zero-width static aggregates as empty bytes

spokodev:
## Summary

An empty fixed array (`T[0]`) or empty tuple (`()`) is a **zero-width static** type. In `encodeAbiParameters.ts`, the static branches of `encodeArray` and `encodeTuple` build their encoding with `concat(preparedParams.map(p => p.encoded))` over an **empty** array.

`concat([])` (in `utils/data/concat.ts`) checks `typeof values[0] === 'string'`. For an empty input `values[0]` is `undefined`, so it routes to `concatBytes([])` and returns an empty **`Uint8Array`** rather than the hex string `'0x'`. That `Uint8Array` then leaks back into `encodeParams`, poisoning the string-vs-bytes detection and producing either **wrong bytes** or an uncaught **`TypeError`**.

The function's documented return type is `Hex`, never `Uint8Array`.

## Reproduction

```ts
encodeAbiParameters([{ type: 'uint256[0]' }, { type: 'uint256' }], [[], 3n])
// before: Uint8Array(66)            (33 leading zero bytes + the uint256)
// after:  the 32-byte uint256 0x..0003

encodeAbiParameters([{ type: 'uint256' }, { type: 'tuple', components: [] }], [5n, []])
// before: throws TypeError: x.replace is not a function   (concatHex on a Uint8Array)
// after:  the 32-byte uint256 0x..0005
```

## Oracle

Per the [ABI spec](https://docs.soliditylang.org/en/latest/abi-spec), `enc(T[0])` and `enc(())` are zero-length: an empty fixed array or empty tuple contributes no bytes. ethers agrees:

```ts
AbiCoder.defaultAbiCoder().encode(['uint256[0]', 'uint256'], [[], 3n])
// '0x0000000000000000000000000000000000000000000000000000000000000003'
AbiCoder.defaultAbiCoder().encode(['uint256', 'tuple()'], [5n, []])
// '0x0000000000000000000000000000000000000000000000000000000000000005'
```

## Fix

Emit `'0x'` directly when `preparedParams` is empty in the static branch of `encodeArray` and `encodeTuple`, instead of relying on `concat([])`. The change is local to the two aggregate encoders; `concat`'s behaviour is untouched (it has other callers).

Added tests cover both repros against the ethers/ABI-spec oracle.

This is a separate root cause from #921 (empty **dynamic** `bytes[]`) and from the RLP trailing-bytes fix in a sibling PR.

### DIFF
--- a/src/utils/abi/encodeAbiParameters.test.ts
+++ b/src/utils/abi/encodeAbiParameters.test.ts
@@ -1921,3 +1921,33 @@ test('https://github.com/wevm/viem/issues/1960', () => {
     Version: viem@x.y.z]
   `)
 })
+
+describe('zero-width static aggregate', () => {
+  // Per the ABI spec, `enc(T[0])` and `enc(())` are zero-length: an empty fixed
+  // array or empty tuple contributes no bytes. The neighbouring uint256 must be
+  // the only output. ethers `AbiCoder.encode(['uint256[0]','uint256'], [[], 3n])`
+  // returns the 32-byte uint256 below.
+  test('empty fixed array', () => {
+    expect(
+      encodeAbiParameters(
+        [{ type: 'uint256[0]' }, { type: 'uint256' }],
+        [[], 3n],
+      ),
+    ).toEqual(
+      '0x0000000000000000000000000000000000000000000000000000000000000003',
+    )
+  })
+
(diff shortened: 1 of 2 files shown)


spokodev:
Sibling PR for a separate root cause: #4778 (fromRlp trailing bytes).

Issue #4770, opened 2026-06-29
[Security] RLP decoder uses Infinity recursiveReadLimit → stack overflow with deeply nested payload


ROCKLEE-1998:
## Summary
`fromRlp()` explicitly sets `recursiveReadLimit: Number.POSITIVE_INFINITY` when creating the byte cursor (line 44-46 of `src/utils/encoding/fromRlp.ts`). A deeply nested RLP list (~10000 levels) causes unbounded recursion → `RangeError: Maximum call stack size exceeded`.

## Affected Code
`src/utils/encoding/fromRlp.ts:44-46`
```typescript
const cursor = createCursor(bytes, {
    recursiveReadLimit: Number.POSITIVE_INFINITY,  // ← disables protection
})
```

The default value (used by ABI decoding) is `8_192`. RLP decoding explicitly overrides this to infinity.

## Impact
Any application that uses Viem to parse untrusted RLP data (transactions, block data) can be crashed. Attack vectors:
1. Malicious node returns deeply nested RLP in transaction data
2. Blockchain indexer parsing blocks hits crafted transaction → crash
3. Denial of service for backend services

## Fix
Use the default `recursiveReadLimit` value instead of overriding to Infinity:
```typescript
const cursor = createCursor(bytes)  // uses default 8_192
```

## Reporter
zhantang bai (zhantang233@gmail.com)

ROCKLEE-1998:
**Requesting CVE assignment**
Severity: Medium (CVSS 5.5) | CWE-674 (Uncontrolled Recursion)
Researcher: zhantang bai (zhantang233@gmail.com)
Once confirmed, please convert to GitHub Security Advisory for auto CVE assignment, or acknowledge so I can submit via MITRE.

ROCKLEE-1998:
Requesting CVE assignment
Severity: Medium (CVSS 5.5) | CWE-674 (Uncontrolled Recursion)
Researcher: zhantang bai ([zhantang233@gmail.com](mailto:zhantang233@gmail.com))
Once confirmed, please convert to GitHub Security Advisory for auto CVE assignment, or acknowledge so I can submit via MITRE.

PR #4778, opened 2026-06-30
fix(rlp): reject trailing bytes in fromRlp

spokodev:
## Summary

`fromRlp` decodes a single item via `fromRlpCursor(cursor, to)` and returns **without** asserting that the cursor consumed all input. Any bytes after the first RLP item are silently dropped.

RLP is a bijection (Ethereum Yellow Paper, Appendix B): a payload encodes exactly one item, with no leftover bytes. Both geth and ethers reject over-long input; viem instead returns a partial value, which is unsafe in the transaction-decode path (`parseTransaction`).

## Reproduction

```ts
fromRlp('0x80deadbeef', 'hex') // before: '0x'             (4 trailing bytes dropped)
fromRlp('0xc2010203', 'hex')   // before: ['0x01','0x02']  (trailing 0x03 dropped)
fromRlp('0xc000', 'hex')       // before: []               (trailing 0x00 dropped)
```

## Oracle

- geth: `rlp: input contains more than one value`
- ethers `decodeRlp`: `unexpected junk after rlp payload`

Both reject all three inputs above.

## Fix

In the public `fromRlp` wrapper (not in the `fromRlpCursor` recursion, so nested lists are unaffected), after decoding assert that the cursor reached the end of the input, otherwise throw a `BaseError` naming the count of leftover bytes.

```ts
fromRlp('0x80deadbeef', 'hex')
// BaseError: Unexpected trailing bytes: RLP payload encodes a single item but 4 byte(s) remain after it.
```

Valid single-item inputs (`0x80`, the empty list `0xc0`, nested lists `0xc1c0`, byte strings) keep decoding unchanged, and the full `parseTransaction` suite still passes (well-formed transactions consume all bytes).

Added tests cover all three trailing-byte cases plus a control that valid single items still decode.

This is a separate root cause from #4770 (Infinity `recursiveReadLimit` / deep-nesting stack overflow) and from the ABI zero-width encode fix in a sibling PR.

### DIFF
--- a/src/utils/encoding/fromRlp.test.ts
+++ b/src/utils/encoding/fromRlp.test.ts
@@ -35,6 +35,51 @@ describe('list', () => {
   })
 })
 
+describe('trailing bytes', () => {
+  // RLP is a bijection (Ethereum Yellow Paper, Appendix B): a payload encodes
+  // exactly one item with no leftover bytes. geth rejects with `rlp: input
+  // contains more than one value` and ethers with `unexpected junk after rlp
+  // payload`. viem must not silently drop the trailing bytes.
+  test('after string item', () => {
+    // `0x80` encodes `''`; the trailing `deadbeef` is junk.
+    expect(() =>
+      fromRlp('0x80deadbeef', 'hex'),
+    ).toThrowErrorMatchingInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


spokodev:
Sibling PR for a separate root cause: #4777 (abi zero-width static aggregates).

PR #4857, opened 2026-07-17
fix: parseUnits rounding for long fractional values

### LINKED ISSUE #4855: [Bug]: parseUnits mis-rounds long fractional tails via Math.round(Number(...))
### Describe the bug

`parseUnits` rounds long fractional tails using `Math.round(Number(...))`. When the fractional part beyond `decimals` is long, IEEE-754 `Number` is not exact and half-up rounding can go the **wrong way**.

```ts
const rounded = Math.round(Number(`${unit}.${right}`))
```

### Debug information

- viem `2.55.3` / checkout @ `a7cb7b60b6e95e1c766de7f456f0bf283f1ae11d`
- File: `src/utils/unit/parseUnits.ts`

### Repro steps

Logic equivalent to current rounding branch:

```js
// parseUnits('1.4499999999999999999', 1)
// viem-style Number rounding → 15n  (as if 1.5)
// correct half-up to 1 decimal → 14n  (1.4…)

// parseUnits('1.14999999999999999', 1)
// viem-style → 12n
// correct → 11n
```

### Expected behavior

Rounding should be exact in decimal/bigint space (inspect the digit after `decimals` and carry without `Number`).

### Actual behavior

Some long fractional strings round **1 ULP too high** at the kept precision (affects `parseUnits` / `parseEther` / `parseGwei` consumers that pass extra precision).

### Suggested direction

Replace `Number`/`Math.round` with string or bigint-only rounding.

Happy to contribute a failing test + fix if maintainers agree on the rounding contract.

nikhilkumar1612:
Fixes #4855

This PR fixes `parseUnits` rounding for long fractional values near half boundaries.

Previously, `parseUnits` used `Number(...)` and `Math.round(...)` when the fractional part exceeded the requested decimal precision. For long decimal strings, JavaScript floating-point conversion can round values like `.49999999999999999999` to `0.5`, causing `parseUnits` to round up incorrectly.

This changes rounding to stay in decimal string/`BigInt` space by checking the first discarded fractional digit and carrying manually.

### DIFF
--- a/.changeset/fix-parse-units-rounding.md
+++ b/.changeset/fix-parse-units-rounding.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `parseUnits` rounding for long fractional values near half boundaries.
--- a/src/utils/unit/parseUnits.test.ts
+++ b/src/utils/unit/parseUnits.test.ts
@@ -53,6 +53,7 @@ test('decimals === 0', () => {
   expect(parseUnits('99999999999999999999999.5', 0)).toMatchInlineSnapshot(
     '100000000000000000000000n',
   )
+  expect(parseUnits('1.49999999999999999999', 0)).toMatchInlineSnapshot('1n')
 })
 
 test('decimals < fraction length', () => {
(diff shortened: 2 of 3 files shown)


jxom:
Fixed.

PR #4858, opened 2026-07-17
fix(tempo): randomize expiring nonce hashes

decofe:
Tempo expiring transactions currently default to the same `nonceKey`, `nonce`, `validAfter`, and `validBefore`. Identical calldata prepared within the same second therefore produces the same expiring nonce hash, which can leave subsequent sponsored requests blocked behind an in-flight transaction.

This sets a random, already-valid `validAfter` when Viem selects an expiring nonce, while preserving an explicitly supplied value. The regression test uses deterministic randomness to verify otherwise-identical sponsored requests receive distinct values.

Validation:
- `biome check src/tempo/chainConfig.ts src/tempo/chainConfig.test.ts .changeset/random-lions-smile.md`
- Targeted Vitest execution was unavailable from the GitHub source archive because its `ox` submodule build is not included; CI will exercise the regression test in the complete checkout.

Prompted by: @jxom

### DIFF
--- a/.changeset/random-lions-smile.md
+++ b/.changeset/random-lions-smile.md
@@ -1,0 +1,5 @@
+---
+'viem': patch
+---
+
+Randomize `validAfter` for Tempo expiring-nonce transactions to prevent otherwise-identical transactions from sharing a nonce hash.
--- a/.changeset/random-lions-smile.md
+++ b/.changeset/random-lions-smile.md
@@ -2,4 +2,4 @@
 'viem': patch
 ---
 
-Randomize `validAfter` for Tempo expiring-nonce transactions to prevent otherwise-identical transactions from sharing a nonce hash.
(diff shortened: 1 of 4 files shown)


.changeset/random-lions-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'viem': patch
+---
+
+Randomize `validAfter` for Tempo expiring-nonce transactions to prevent otherwise-identical transactions from sharing a nonce hash.
```
brendanjryan: is this not "randomize validAfter if not set"?

brendanjryan: I odn't think we are providing a global jitter

jxom (changes requested):
.

jxom (approved):
The sponsored integration regression now exercises distinct hashes and successful receipts without mocking randomness.

PR #4885, opened 2026-07-22
fix: decode bytesToNumber/bytesToBigInt independent of size

nikolauspschuetz:
`bytesToBigInt` / `bytesToNumber` (and `fromBytes` with `to: "bigint" | "number"`) passed the full `opts` — including `size` — into `bytesToHex`. `bytesToHex` with a `size` **right-pads** the hex, appending low-order zero bytes, which multiplies the decoded integer by `256^(size - len)` whenever the input is shorter than `size`.

`assertSize` only enforces an upper bound (a value larger than `size` throws), so shorter inputs are deliberately allowed through — and then silently corrupted.

```ts
fromBytes(Uint8Array.from([1, 164]), { to: 'bigint', size: 32 })
// actual:   420n * 2n ** 240n
// expected: 420n
```

This also disagrees with the equivalent hex path, which only asserts the size and never pads the value:

```ts
fromHex('0x01a4', { to: 'bigint', size: 32 }) // 420n  ✅
```

**Fix:** convert the bytes to hex without `size`, so the decoded value is independent of the padding width. The `size` assertion (`assertSize`) is unchanged, and full-length inputs are unaffected (the right-pad was a no-op there, which is why existing tests passed).

**Test:** added a shorter-than-`size` case to the existing `args: size` blocks for both `bytesToNumber` and `bytesToBigInt`. They fail on `main` and pass with this change.

### DIFF
--- a/.changeset/frombytes-size-scaling.md
+++ b/.changeset/frombytes-size-scaling.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `bytesToBigInt` and `bytesToNumber` (and `fromBytes` with `to: "bigint" | "number"`) scaling the decoded value when a `size` larger than the input byte length was provided.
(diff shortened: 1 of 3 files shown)


jxom:
Thanks for putting this together! We’ve carried the change forward in #4887, so I’m closing this PR to keep the work in one place. Appreciate the contribution.

Issue #1986, opened 2024-03-20
bug: fromBlobs stops parsing too early


RogerPodacter:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

2.8.14

### Current Behavior

`fromBlobs` will stop parsing on the first instance of `0x80`, not the last instance, which leads to incorrect parsing. Here's a proof of concept:

```javascript
const input = new Uint8Array([1, 2, 128])
const blobs = toBlobs({ data: input })

const original = fromBlobs({ blobs: blobs })

const blobTwo = toBlobs({ data: original })

console.log(JSON.stringify(blobTwo) == JSON.stringify(blobs))
```

This logs `false`.

### Expected Behavior

The expected behavior is that `toBlobs` and `fromBlobs` always correctly "round trip." The `console.log` above should return true in all cases, just as it does here:

```javascript
const input = new Uint8Array([1, 2, 3])
const blobs = toBlobs({ data: input })

const original = fromBlobs({ blobs: blobs })

const blobTwo = toBlobs({ data: original })

console.log(JSON.stringify(blobTwo) == JSON.stringify(blobs))
```

### Steps To Reproduce

Use `fromBlobs` on any blob where `0x80` exists "naturally" (i.e., not as the blob terminator). Observe incorrect decoding behavior.

jxom:
Nice catch!

PR #4886, opened 2026-07-22
fix: don't misread a 0x80 data byte as the fromBlobs terminator

nikolauspschuetz:
`toBlobs` marks the end of the encoded data with a `0x80` terminator byte. That terminator is only ever written to the **last** blob: every preceding blob is completely filled with data (a terminator is only emitted when a data segment shorter than 31 bytes is reached, which ends encoding).

`fromBlobs`, however, scanned for the terminator in *every* blob, disambiguating it from a genuine `0x80` data byte by checking that no other `0x80` followed it within the same blob. When a completely-full blob's final data byte happens to be `0x80`, there are no bytes left in that blob, so the check passes and the byte is misclassified as the terminator: it is dropped and every subsequent blob is skipped.

This breaks the round-trip `fromBlobs(toBlobs(data)) === data`:

```ts
// one blob's data capacity: fieldElementsPerBlob * (bytesPerFieldElement - 1) = 4096 * 31
const data = new Uint8Array(4096 * 31).fill(1)
data[data.length - 1] = 0x80

fromBlobs({ blobs: toBlobs({ data }) })
// ❌ returns data with the trailing 0x80 dropped (and, with more data, the
//    entire following blob missing)
```

**Fix:** only look for the terminator in the last blob, since that is the only blob `toBlobs` ever writes it to.

**Tests:** added two round-trip regressions (full blob ending in `0x80`; and the same followed by another blob). Both fail on `main` and pass with this change. Existing `fromBlobs`/`toBlobs` tests (including the #1986 regression) still pass.

### DIFF
--- a/.changeset/fromblobs-full-blob-terminator.md
+++ b/.changeset/fromblobs-full-blob-terminator.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `fromBlobs` dropping data (and skipping subsequent blobs) when a completely-full blob's final data byte was `0x80`, which was misread as the blob terminator.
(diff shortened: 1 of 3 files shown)


jxom:
Thanks for putting this together and for catching this edge case! We’ve carried the fix forward in #4888, so I’m closing this PR to keep the work in one place. Appreciate the contribution.

PR #4938, opened 2026-07-31
Add nested Tempo Earn vault actions

Slokh:
## Summary

Adds the Viem integration surface for nested Tempo Earn vaults introduced in [tempoxyz/earn#158](https://github.com/tempoxyz/earn/pull/158).

- updates generated Earn ABI provenance to Earn commit `56bc902`;
- adds `NestedEarnVaults` and `NestedAllocation` with consistent `inner` and `outer` roles;
- adds unified position, allocation, deposit, redemption, and unwrapping actions;
- provides asynchronous and confirmed mutation variants plus composable call builders;
- keeps deposits and redemptions atomic with independent per-vault output bounds;
- verifies that the outer engine wraps the supplied inner vault and both vaults use one asset; and
- adds action references, a complete nested Earn guide, a changeset, type coverage, call-builder tests, and a real nested-contract lifecycle test.

## Public Surface

- `getNestedPosition` values inner and outer shares as one aggregate position.
- `getNestedAllocation` reads the live outer allocation and derives the inner remainder.
- `getOuterShareQuote` converts the inner share quote for an outer deposit into expected outer shares.
- `depositNested` executes the exact outer-first then inner split atomically.
- `getNestedRedeemQuote` and `redeemNested` quote and redeem both vaults.
- `getUnwrapQuote` and `unwrapNested` convert outer shares directly into inner shares without moving the underlying asset.

If outer capacity changes after quoting, the complete deposit reverts and the application re-quotes. When outer admission ends, new deposits allocate entirely to the inner vault. Existing outer shares remain backed and may be redeemed or unwrapped by their holders.

## Merge Order

The Earn PR should land first. Regenerate the vendored ABIs and artifacts from the final merged Earn commit before merging this PR so the source marker remains durable.

## Validation

- `npm run check`
- `./node_modules/.bin/tsc -b`
- `npm run build:types`
- `vocs build`

The focused Tempo test suite currently reaches the local server but fails during setup because `eth_getBlockByNumber(latest)` returns HTTP 400 before the first block is readable. It reproduced twice and is recorded in the included friction log. The prior branch revision passed all 74 Earn action tests before this environment regression.


Slokh:
WIP, AI opened a PR for no reason 😭

PR #4949, opened 2026-08-03
Fix withRetry hanging when retry callbacks throw

grandpig:
`withRetry` could remain pending indefinitely when `shouldRetry` threw or rejected, or when the retry delay callback threw.

`attemptRetry()` is asynchronous, but its initial invocation did not handle its rejection. The recursive retry invocation was also not returned, which broke the rejection chain for errors occurring after the first retry.

This change:

- returns the recursive `attemptRetry` invocation;
- forwards the initial `attemptRetry` rejection to the outer promise;
- adds regression coverage for:
  - a synchronous `shouldRetry` exception after a retry;
  - an asynchronous `shouldRetry` rejection;
  - a delay callback exception after a retry.

This preserves the existing retry and abort behavior while ensuring callback failures reject the promise returned by `withRetry`.

### DIFF
--- a/.changeset/fuzzy-pandas-retry.md
+++ b/.changeset/fuzzy-pandas-retry.md
@@ -1,0 +1,5 @@
+---
+'viem': patch
+---
+
+Fixed `withRetry` hanging when `shouldRetry` or the retry delay callback throws.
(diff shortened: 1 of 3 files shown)


grandpig:
@jxom Hi, Could you please review this PR at your convenience? Thank you very much.

PR #4951, opened 2026-08-03
fix(tempo): respect token metadata batching config

struong:
## Summary

Updates Tempo token metadata reads to honor explicit client multicall configuration while preserving deployless Multicall3 as the action's default.

## Motivation

Zone clients cannot execute deployless contract-creation `eth_call` requests. Explicitly disabling client multicall should allow `token.getMetadata` to use ordinary contract reads without changing the default behavior for existing Tempo clients.

## Changes

- Preserved deployless metadata multicall when the client does not specify a multicall setting.
- Used concurrent `readContract` calls when the client explicitly configures multicall behavior.
- Preserved pathUSD fields, declared-token overrides, logo URI fallback, and deterministic required-field errors.
- Added focused Tempo and Zone coverage plus a patch changeset.

## Testing

- `pnpm test src/tempo/actions/token.test.ts --run` — 76 passed, 1 existing skip.
- `pnpm check:types`
- `pnpm build:types`
- Attempted `VITE_TEMPO_ZONES=true pnpm test src/tempo/actions/zone.test.ts --run`; the local Zone container exited before tests because the configured dev key did not own `ZoneFactory`.

### DIFF
--- a/.changeset/tidy-pans-fetch.md
+++ b/.changeset/tidy-pans-fetch.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Respect explicit client multicall configuration when fetching Tempo token metadata.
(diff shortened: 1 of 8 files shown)


jxom:
@codex review

struong:
Superseded by #4956, which uses a branch in `wevm/viem` so CI can access repository secrets.

PR #4961, opened 2026-08-04
fix: default uint/int size in validateTypedData

RezaRahemtola:
`validateTypedData` crashes with `RangeError` on uint/int types (no explicit bit size, the Solidity aliases for uint256/int256).

It looks like the size capture group in integerRegex is optional, so a bare uint matches with an `undefined` size causing a crash later on.
Every other integer handler defaults the missing size to '256' (`encodeAbiParameters`, `encodePacked`, `decodeAbiParameters`), but it was probably forgotten here.

Fixed it and added regression tests for uint and int.

### DIFF
--- a/.changeset/typed-data-uint-default-size.md
+++ b/.changeset/typed-data-uint-default-size.md
@@ -2,4 +2,4 @@
 "viem": patch
 ---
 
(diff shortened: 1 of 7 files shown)


jxom:
EIP-712 only defines explicitly sized integer types and states that the `uint` and `int` aliases are not supported. Accepting those aliases would encode a noncanonical type string into the struct hash, producing signatures that canonical EIP-712 implementations can reject.

PR #4984, opened 2026-08-07
Create recanetTestnet.ts

jinu-boop:


jxom:
Closing because chain ID `785` is already assigned to AUTHEO Testnet, and the proposed RPC and explorer hosts do not resolve.

PR #4990, opened 2026-08-08
fix(siwe): reject unparseable expirationTime and notBefore

SashaMIT:
## Summary
`parseSiweMessage` turns fields like `Expiration Time: never` into a truthy `Invalid Date`. In `validateSiweMessage`, the lifetime checks were:

```ts
if (message.expirationTime && time >= message.expirationTime) return false
if (message.notBefore && time < message.notBefore) return false
```

Comparisons against `Invalid Date` are always `false`, so both gates no-op and a signed SIWE message with garbage expiration / not-before still validates. This rejects unparseable dates when those fields are present (same class as [thirdweb-dev/js#8875](https://github.com/thirdweb-dev/js/pull/8875) and [supabase/auth#2688](https://github.com/supabase/auth/pull/2688)).

## Test plan
- [x] `pnpm exec vitest run -c ./test/vitest.config.ts src/utils/siwe/validateSiweMessage.test.ts` → 12/12
- [x] Cases for `Expiration Time: never` / garbage `Not Before`

Made with [Cursor](https://cursor.com)


jxom:
Two remaining cases:

- `getTime()` checks JavaScript date validity, not RFC 3339. Validate the raw SIWE timestamp before converting it to `Date`.
- An invalid `time` makes both lifetime comparisons false. Reject it before checking `expirationTime` or `notBefore`.

See the [ERC-4361 validation requirements](https://eips.ethereum.org/EIPS/eip-4361).

SashaMIT:
Thanks @jxom. Follow-up covers both cases:

1. Raw SIWE timestamps are checked against the EIP-4361 / RFC 3339 profile before `Date` coercion in `parseSiweMessage`.
2. Invalid caller `time` is rejected before `expirationTime` / `notBefore` comparisons.

Added a changeset. SIWE unit tests green locally (25/25).

SashaMIT:
Thanks for the merge @jxom. Glad the RFC 3339 and invalid-time follow-up landed with the original fail-closed checks.

PR #4997, opened 2026-08-11
Fix asset discovery in simulateCalls

Ghadi8:
### What this solves

With `traceAssetChanges: true`, candidate tokens were discovered by running `eth_createAccessList` per call, independently, against unmodified state at `latest`. Consequences:

1. **Threw instead of reporting a revert.** `createAccessList` rejects reverting calls, so one call that fails against current state (a transfer the account cannot yet afford) failed the whole action instead of landing in `results`.
2. **Ignored `stateOverrides`.** The probes passed none (support for overrides on `eth_createAccessList` is not portable across nodes), so discovery could not see balances or allowances the caller overrode, even though the result simulation could.
3. **Ignored `blockNumber`.** No block was forwarded; discovery ran at `latest` while results were simulated at the requested block.
4. **Could not see intra-batch state.** Each call was probed against pre-state, so a call that only succeeds because an earlier one ran (`deposit()` then `transfer()`) contributed no assets.
5. **The two passes could disagree.** With no block requested, discovery and measurement resolved `latest` independently: separate requests, possibly different replicas, head free to advance in between. A token the measurement pass moved but discovery never saw was silently absent from `assetChanges`.

Discovery now simulates the batch once via `eth_simulateV1` and takes token addresses from `Transfer` logs where the account is `from` or `to`, plus each call's `to`. That inherits sequential execution, state overrides, and the requested block. The access-list pass remains as supplementary hints (assets that change without a `Transfer` the account participates in), now block-pinned and failure-tolerant. For (5), the base block is resolved once up front and both passes plus the hints are pinned to it; only `latest` is resolved, an explicit block or tag passes through untouched.

`balanceOf` probes now require non-empty return data, not just `status === 'success'`: each call's `to` is a candidate, and a `to` without code returns empty data that would decode as a bogus `0n` balance.

### Tests

New coverage: a call that reverts in isolation; a call that depends on an earlier call; a token reachable only through logs (ETH to USDC via the Uniswap V2 router, where the only `to` is the router); and two RPC-spy tests asserting that both simulations and the access-list pass use the requested block, and that they stay pinned to the base when the head advances between them.

All `traceAssetChanges` tests are skipped on the pinned Anvil, including the pre-existing `behavior: with mutation calls + asset changes`. Tracing reads the account's ETH balance through a deployless call with no `to`, and Anvil v1.7.1 (the version `verify.yml` pins) rejects those in `eth_simulateV1` with `eip1559 transaction can't be built due to missing keys: ["to"]`. Anvil 1.5.1 handles these correctly, and the capability was restored on master by foundry-rs/foundry#15784 (merged 2026-07-16, unreleased at the time of writing). The tests pass against Anvil 1.5.1 and against builds carrying that fix.

### Alternatives explored

- **try/catch around the existing probes.** Fixes the throw and nothing else; overrides, block, and intra-batch state stay wrong.
- **`parseEventLogs` with `erc20Abi`.** ERC-721 `Transfer` has three indexed params and fails strict decoding; matching `topics[0..2]` directly covers both under one predicate.
- **Deriving the base block from the discovery response (`number - 1n`) instead of resolving it.** No extra request, but the numbering is not portable: the spec and geth put the first simulated block at base + 1, while Anvil (through v1.7.1) puts it at base. The derivation would pin measurement one block behind discovery on Anvil, reintroducing the divergence deterministically. Resolving `eth_blockNumber` once (`cacheTime: 0`, so a cached head cannot pin to a stale block) is version-agnostic at the cost of a round trip.

### Worth reviewer attention

- `traceAssetChanges: true` now costs one extra `eth_simulateV1`, plus one `eth_blockNumber` when no block is requested. The hints run concurrently with the discovery simulation. Discovery cannot fold into the existing multi-block request: the token set must be known before the balance blocks are built.
- Number-pinning, not hash-pinning. `eth_simulateV1` takes no block hash, so a reorg between the two requests remains possible; that is the ceiling the API allows.
- `accessListHints` swallows all errors by design (some nodes lack the method, and it rejects reverting calls); log-based discovery is the primary path. Happy to drop the supplementary pass entirely if you prefer the simpler version.
- ERC-1155 remains untraced: its `balanceOf` takes a token id, so it does not fit the single-probe shape.

Docs updated with how discovery works and the balances that are not traced.


jxom:
@codex review

PR #5015, opened 2026-08-17
fix(rpc): trigger auto-reconnect on unexpected WebSocket close events

taiwofrancis:
Fixed WebSocket transport connection termination by triggering an explicit auto-reconnect fallback error routing loop when unexpected socket close frames occur.

### DIFF
--- a/src/utils/rpc/webSocket.ts
+++ b/src/utils/rpc/webSocket.ts
@@ -1,4 +1,4 @@
-import type { MessageEvent } from 'isows'
+import type { MessageEvent, CloseEvent } from 'isows'
 
 import {
   SocketClosedError,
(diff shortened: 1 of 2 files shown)


src/utils/rpc/webSocket.ts
```diff
@@ -28,13 +28,26 @@ export async function getWebSocketRpcClient(
...
+      // Updated to receive the native close event
+      function onClose_(event?: CloseEvent) {
+        socket.removeEventListener('close', onCloseWrapper)
         socket.removeEventListener('message', onMessage)
         socket.removeEventListener('error', onError)
         socket.removeEventListener('open', onOpen)
-        onClose()
+        
+        // If closed abnormally (not 1000 Normal or 1001 Going Away) and reconnect is active
+        if (event && event.code !== 1000 && event.code !== 1001 && reconnect) {
```
official-burak, line 39: `reconnect` here is the raw option from `getWebSocketRpcClient`. Callers that omit it pass `undefined`, so this branch never runs even though `getSocketRpcClient` defaults `reconnect` to `true`.

`onClose` and `onError` both already call `attemptReconnect()` in `socket.ts`. An unexpected close already reconnects via `onClose()`. Treat omitted as enabled (`reconnect !== false`), or drop the extra `&& reconnect` check.

official-burak (commented):
The new reconnect branch is skipped for the default client, and unexpected close already reconnects via `onClose`.

PR #5024, opened 2026-08-18
fix: fail closed on non-HTTP(S) and link-local CCIP Read URLs

SashaMIT:
`ccipRequest` fetched resolver-supplied gateway URLs with no scheme or host check, and `fetch` followed redirects.

A contract `OffchainLookup` can return `file:///...` or `http://169.254.169.254/...`. On Node that becomes a client-side request to a local file or link-local metadata address. A public URL that 302s to the same place had the same effect.

This is the URL-validation class web3.py already applies in `validate_ccip_url_scheme` / `validate_ccip_url_host`. Loopback HTTP stays allowed so local gateways and the existing test servers keep working. Callers who want a stricter allowlist can still wrap `ccipRead.request`.

## Change

- Reject non-`http(s)` schemes before `fetch`.
- Reject IPv4 `0.0.0.0/8` and `169.254.0.0/16`, plus IPv6 unspecified and `fe80:` link-local literals.
- Set `redirect: 'manual'` so a public gateway cannot bounce the client onto a blocked host.
- If one URL in the list is blocked, try the next.

## Test

`SKIP_GLOBAL_SETUP=true pnpm exec vitest run -c ./test/vitest.config.ts --project core src/utils/ccipRequest.url.test.ts` (4/4). Revert-tested: the three reject assertions fail when the pre-fetch guards are removed.

### DIFF
--- a/src/utils/ccip.ts
+++ b/src/utils/ccip.ts
@@ -263,19 +263,42 @@ function assertCcipRequestUrl(url: string) {
   }
 }
 
-function isBlockedCcipHostname(hostname: string) {
-  const ipv4 = /^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$/.exec(hostname)
-  if (ipv4) {
-    const octets = ipv4.slice(1, 5).map(Number)
-    if (octets.some((octet) => octet > 255)) return false
-    const [a, b] = octets
-    // 0.0.0.0/8 unspecified, 169.254.0.0/16 link-local / cloud metadata
-    return a === 0 || (a === 169 && b === 254)
-  }
+function isBlockedIpv4([a, b]: readonly number[]) {
+  // 0.0.0.0/8 unspecified, 169.254.0.0/16 link-local / cloud metadata
+  return a === 0 || (a === 169 && b === 254)
+}
 
-  const normalized = hostname.toLowerCase()
-  if (normalized === '::' || normalized === '0:0:0:0:0:0:0:0') return true
+function parseIpv4Literal(host: string) {
+  const ipv4 = /^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$/.exec(host)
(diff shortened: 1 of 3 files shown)


src/utils/ccip.ts
```diff
@@ -217,3 +225,57 @@ export async function ccipRequest({
...
+    throw new HttpRequestError({
+      details:
+        `CCIP Read request to '${parsed.hostname}' is not allowed: ` +
+        'host is in a blocked link-local or unspecified range.',
+      url,
+    })
+  }
+}
+
+function isBlockedCcipHostname(hostname: string) {
```
official-burak, line 291: Node's `URL.hostname` keeps brackets on IPv6 literals:

- `new URL('http://[::]/').hostname` is `'[::]'`
- `new URL('http://[fe80::1]/').hostname` is `'[fe80::1]'`
- `new URL('http://[::ffff:169.254.169.254]/').hostname` is `'[::ffff:a9fe:a9fe]'`

None of those equal `'::'`, start with `'fe80:'`, or match the dotted-quad regex, so the v6 and IPv4-mapped link-local cases this PR is meant to reject still go through to `fetch`.

Strip brackets before the checks. Run `:ffff:` mapped addresses through the same 169.254 / `0.0.0.0/8` rules.

official-burak (commented):
IPv6 literals never hit the new block list. Node keeps brackets on `URL.hostname`, and v4-mapped link-local addresses are a second miss.

SashaMIT:
Good catch. Node keeps the brackets and rewrites the mapped form to hex, so those never hit the first checks.

The guard now strips brackets and runs v4-mapped addresses through the same 169.254 / 0.0.0.0/8 rules. Added tests for `[::]`, `[fe80::1]`, and `[::ffff:169.254.169.254]`.

official-burak:
Looks right. Stripping brackets and running v4-mapped addresses through the same 169.254 / 0.0.0.0/8 checks covers the two misses.

PR #5029, opened 2026-08-20
feat(tempo): multisig

jxom:
Adds persistent multisig operation storage and explicit approval actions to Tempo clients. This keeps partial quorum coordination out of standard transaction sending while preserving direct sends for complete multisig envelopes.

Supersedes https://github.com/wevm/viem/pull/5026.


gakonst:
How do you handle transaction hashes not landed on chain yet? Ie what does eth get tx receipt return of anything?

gakonst:
Also should nail the DX / UX of going through the 1 of 1 flow - since that will be very common by default as well as going from 1 of 1 to 1 of 2 when upgrading an agent account to a passkey backed or otherwise account (maybe using prf)

PR #5050, opened 2026-08-29
Fix isHex to reject odd-length hex strings in strict mode

### LINKED ISSUE #5048: isHex shouldn't accept odd number of nibbles
### Check existing issues

### Viem Version

2.48.1

### Current Behavior

'isHex' accepts odd numbered nibbles, even in the strict mode, `isHex("0x1") === true`.

### Expected Behavior

'isHex' rejects odd numbered nibbles in the scrict mode, `isHex(0x1) === false`.

### Steps To Reproduce

_No response_

_No response_

### Anything else?

It really doesn't make sense to accept odd-length hex values.

Everything around EVM works with full bytes and never with half-bytes. I'm not aware of any API or tool that would accept odd-length hex values, they all reject such hex strings as malformed with the risk of weirder ones silently corrupting data by "fixing" the parity. There's no use case for handling odd-length hex values, but there are risks in accepting them as valid.

psamuelvijay:
Fixes #5048

## Changes
- Updated the regex in `isHex()` from `/^0x[0-9a-fA-F]*$/` to `/^0x([0-9a-fA-F][0-9a-fA-F])*$/` to require pairs of hex digits (full bytes) in strict mode
- Updated existing test that depended on the old behavior (`isHex('0x0')` now returns false instead of true)
- Added new test cases:
  - `isHex('0x01')` returns true (even length)
  - `isHex('0x123')` returns false (odd length)
  - Non-strict mode still allows odd-length hex strings

## Notes
- This is a breaking change: `isHex('0x0')` flips from true to false in strict mode
- Non-strict mode behavior is unchanged
- Empty hex string `'0x'` (zero bytes) remains valid
- Changeset added as patch version bump

### DIFF
--- a/.changeset/fix-ishex-odd-nibbles.md
+++ b/.changeset/fix-ishex-odd-nibbles.md
@@ -1,0 +1,5 @@
+---
+"viem": patch
+---
+
+Fixed `isHex` incorrectly returning `true` for odd-length hex strings in strict mode
--- a/src/utils/data/isHex.test.ts
+++ b/src/utils/data/isHex.test.ts
@@ -4,11 +4,15 @@ import { isHex } from './isHex.js'
 
 test('is hex', () => {
   expect(isHex('0x')).toBeTruthy()
-  expect(isHex('0x0')).toBeTruthy()
+  expect(isHex('0x0')).toBeFalsy()
+  expect(isHex('0x01')).toBeTruthy()
+  expect(isHex('0x123')).toBeFalsy()
   expect(isHex('0x0123456789abcdef')).toBeTruthy()
   expect(isHex('0x0123456789abcdefABCDEF')).toBeTruthy()
   expect(isHex('0x0123456789abcdefg')).toBeFalsy()
   expect(isHex('0x0123456789abcdefg', { strict: false })).toBeTruthy()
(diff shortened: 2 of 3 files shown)


psamuelvijay:
## Local Testing Complete

Fixed submodule SSH authentication issue and ran tests locally:

✅ **Core test passes:**
\\\
npx pnpm vitest run src/utils/data/isHex.test.ts
 Test Files  1 passed (1)
      Tests  1 passed (1)
\\\

✅ **Related functionality verified:**
- \src/utils/hash/isHash.test.ts\ (uses \isHex\ internally) - **PASSED**

All tests specific to the \isHex\ function pass successfully. The fix correctly:
- Rejects odd-length hex strings in strict mode (\'0x0'\, \'0x123'\)  
- Accepts even-length hex strings (\'0x01'\, \'0x0123'\)
- Preserves non-strict mode behavior (allows odd-length)
- Keeps empty hex \'0x'\ valid

Ready for review.

psamuelvijay:
Rebased on latest \main\ (includes #5052, #5053). No conflicts.

psamuelvijay:
Thanks for taking a look! Could you share some context on why this didn't land? Happy to rework the approach if there's a direction you'd prefer.

PR #5063, opened 2026-09-03
feat: add HyveChain

nicknixcape:
Adds HyveChain (chain ID 7847).

### Requirements

- [x] **Merged into [ethereum-lists/chains](https://github.com/ethereum-lists/chains)** — [`_data/chains/eip155-7847.json`](https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-7847.json)
- [x] **Unique chain ID** — 7847 (`0x1ea7`), also its `networkId`
- [x] **Human-readable name** — HyveChain
- [x] **Native currency** — HYVE, 18 decimals
- [x] **Public RPC** — `https://rpc.hyvechain.com`
- [x] **Block explorer** — `https://explorer.hyvechain.com` (EIP-3091)

### Verified before opening

```
eth_chainId              0x1ea7
net_version              7847
JSON-RPC batch requests  supported
EIP-1559                 baseFeePerGas present
wss://ws.hyvechain.com   HTTP 101 Switching Protocols, eth_chainId -> 0x1ea7
```

The WebSocket URL is included because it was tested end to end, not just
resolved — a raw upgrade handshake followed by a live `eth_chainId` call.

### No multicall3

`0xcA11bde05977b3631167028862bE2a173976CA11` holds no code on this chain, so
`contracts.multicall3` is deliberately omitted rather than pointed at an address
that would revert. Happy to follow up with a second PR once it is deployed and
its bytecode matches the canonical contract.

### Note on the RPC

The public endpoint sits behind Cloudflare and is served by a node that is
currently modestly provisioned. It answers reliably for the calls above, but if
CI hits it hard and sees timeouts, that is capacity rather than misconfiguration
— tell me and I will point the definition at a better-provisioned endpoint.


nickhyvechain:
Multicall3 is now deployed, so I have added `contracts.multicall3` to the definition.

```
address      0xcA11bde05977b3631167028862bE2a173976CA11
blockCreated 4090846
tx           0x07471adfe8f4ec553c1199f495be97fc8be8e0626ae307281c22534460184ed1
```

Deployed with the canonical pre-signed transaction, so it landed at the standard
address. Verified rather than assumed: the deployed bytecode is byte-identical to
the contract on Ethereum mainnet (7618 chars, sha256 `0fb6a9dbcc4e17098c35d4b821bfc8cc`),
and `getBlockNumber()` answers correctly through it.

The Vercel check failure is the fork-secrets one that also affects #5062 and #5058, not this change.

PR #5113, opened 2026-09-20
fix: throw when TestClient.revert gets a false evm_revert result

### LINKED ISSUE #5064: TestClient.revert doesn't check for errors
### Check existing issues

### Viem Version

2.48.1

### Current Behavior

Calling `TestClient.revert` with an invalid, malformed or nonexinstent snapshot ID is a failure that is not observable when calling the function.

### Expected Behavior

An issue like this should throw an error. In most cases it will crash the test in which it's used, which is the correct behavior because the snapshot handling logic must be invalid and the chain is in an unexpected state. The users may explicitly choose to catch and ignore the error to return to the current behavior.

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

rome-xi:
## Summary

- `TestClient.revert` now checks the boolean returned by `evm_revert` and throws `SnapshotRevertError` when Hardhat/Anvil report that the snapshot ID is missing, malformed, or already consumed (previously the failure was silent).
- Adds regression tests for an invalid id and a double-revert of the same snapshot.
- Patch changeset + docs note.

Fixes #5064

## Context

Prior community PRs (#5065, #5066, #5071) closed without merging; the bug is still present on current `main` (`revert` ignores the RPC result).

## Test plan

- [x] Code review against current `main`
- [ ] `pnpm test src/actions/test/revert.test.ts` (Anvil/Foundry required)
- [ ] Confirm invalid snapshot id throws `SnapshotRevertError`
- [ ] Confirm consumed snapshot id throws on second revert


rome-xi:
Understood — thanks for taking a look. Closing out from our side; happy to reopen with a tighter repro if useful later.

